Issue #9, opened 2019-03-11
How to deal with partial resources


skoging:
As far as I can see, there isn't currently a robust way to handle requests that return partial resources.

In particular, I want to do a list request where I only want to request a couple of the fields of a resource. Then I will request the complete resource when I need it. This works as expected, where the detailed resource replaces the existing partial resource. The problem is that if I reverse the order of these requests. 
Say I'm at a detail page, and navigate back to the list page, if the list request is stale or not in the cache, the partial resources that are fetched will replace any complete resource that is already cached.

Are there any suggestions for how to deal with this? Or plans for improvements that will make this use case possible?

skoging:
I did dome more testing, and it appears lodash merge deals with class instances differently to plain objects. 

https://runkit.com/skoging/5c866ccaa0922f0014c0835e

ntucker:
Are you looking for more feedback, or just busy?

skoging:
I've just been busy lately. I'll take another pass on this in the next
couple days.

ntucker:
Closed in https://github.com/coinbase/rest-hooks/pull/10

ntucker:
https://github.com/coinbase/rest-hooks/pull/10

PR #10, opened 2019-03-11
[enhance] Merge changes to resource instances

skoging:
Adresses #9 

lodash merge only recursively merges arrays and plain objects, not class instances.

Use lodash mergeWith and a customizer that can handle instances of Resource.

The `resourceCustomizer` function will merge Resources recursively.

### DIFF
--- a/src/state/reducer.ts
+++ b/src/state/reducer.ts
@@ -1,5 +1,5 @@
 import { normalize } from '../resource';
-import { mergeWith, MergeWithCustomizer } from 'lodash';
+import { mergeWith } from 'lodash';
 import { Resource } from '../resource';
 import { ActionTypes, State } from '../types';
 
@@ -13,15 +13,28 @@ type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
 }
 
-export const resourceCustomizer: MergeWithCustomizer = (a, b) => {
-  if (a instanceof Resource && b instanceof Resource) {
-    const Static = b.constructor as typeof Resource;
+interface MergeableStatic<T> {
+  new(): T;
+  merge(a: T, b: T): T;
+}
+
+function isMergeable<T>(
+  constructor: any
+): constructor is MergeableStatic<T> {
+  return (
+    constructor &&
+    typeof constructor.merge === 'function'
+  );
+}
+
+export const resourceCustomizer = (a: any, b: any): any => {
+  const Static = b && b.constructor;
+  if (a && Static && isMergeable(Static)) {
     return Static.merge(a, b);
   }
 
   // use default merging in lodash.merge()
-  return undefined;
-}
+};
 
 export default function reducer(state: State<Resource>, action: ActionTypes) {
   switch (action.type) {
--- a/src/state/reducer.ts
+++ b/src/state/reducer.ts
@@ -1,5 +1,5 @@
 import { normalize } from '../resource';
-import { merge } from 'lodash';
+import { mergeWith } from 'lodash';
 import { Resource } from '../resource';
 import { ActionTypes, State } from '../types';
 
@@ -11,6 +11,14 @@ export const initialState: State<Resource> = {
 
 type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
+}
+
+function resourceCustomizer(a, b) {
+	if (b instanceof Resource) {
+			const merged = mergeWith({ ...a }, b, resourceCustomizer);
+
+			return Object.assign(b, merged);
+	}
 }
 
 export default function reducer(state: State<Resource>, action: ActionTypes) {
@@ -31,7 +39,7 @@ export default function reducer(state: State<Resource>, action: ActionTypes) {
       }
       const normalized = normalize(action.payload, action.meta.schema);
       return {
-        entities: merge({ ...state.entities }, normalized.entities),
+        entities: mergeWith({ ...state.entities }, normalized.entities, resourceCustomizer),
         results: {
           ...state.results,
           [action.meta.url]: normalized.result,
@@ -49,7 +57,7 @@ export default function reducer(state: State<Resource>, action: ActionTypes) {
       let { entities } = normalize(action.payload, action.meta.schema);
       return {
         ...state,
-        entities: merge({ ...state.entities }, entities),
+        entities: mergeWith({ ...state.entities }, entities, resourceCustomizer),
       };
     case 'purge':
       if (action.error) return state;
--- a/src/state/reducer.ts
+++ b/src/state/reducer.ts
@@ -14,9 +14,9 @@ type Writable<T> = {
 }
 
 function resourceCustomizer(a: any, b: any): any {
-  if (b instanceof Resource) {
-    const merged = mergeWith({ ...a }, b, resourceCustomizer);
-    return Object.assign(b, merged);
+  if (a instanceof Resource && b instanceof Resource) {
+    const Static = b.constructor as typeof Resource;
+    return Static.merge(a, b);
   }
 }
 
--- a/src/state/reducer.ts
+++ b/src/state/reducer.ts
@@ -1,5 +1,5 @@
 import { normalize } from '../resource';
-import { mergeWith } from 'lodash';
+import { mergeWith, MergeWithCustomizer } from 'lodash';
 import { Resource } from '../resource';
 import { ActionTypes, State } from '../types';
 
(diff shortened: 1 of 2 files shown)


src/state/reducer.ts
```diff
@@ -13,6 +13,14 @@ type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
 }
 
+function resourceCustomizer(a, b) {
+	if (b instanceof Resource) {
+			const merged = mergeWith({ ...a }, b, resourceCustomizer);
+
+			return Object.assign(b, merged);
```
skoging: The source Resource instance is mutated, which in this case will be from `normalize()`, and should be safe to mutate. There is probably a better way to do this.

src/state/reducer.ts
```diff
@@ -13,6 +13,14 @@ type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
 }
 
+function resourceCustomizer(a, b) {
+	if (b instanceof Resource) {
+			const merged = mergeWith({ ...a }, b, resourceCustomizer);
```
skoging: Resource instances are merged recursively, not sure if this is necessary.

ntucker:
Can you add tests for your use case? Thanks for the pr!

skoging:
Already on it 👍

src/state/reducer.ts
ntucker: any mutation of existing state in a reducer will have bad side effects. You'll need to copy on merge. Perhaps adding a new bulk update method to Resource could be of assistance.

skoging: I agree. This mutation felt very dirty, and even though it might be safe right now, it is likely to cause bugs that are difficult to debug if that changes.

I figured the better option was to add a new method on Resource, but didn't want to spend to much time on this until I got some feedback.

ntucker: Probably something resembling https://immutable-js.github.io/immutable-js/docs/#/Map/merge since I'll want to eventually have this library optionally use immutable-js

ntucker:
```typescript
  static fromJS<T extends typeof Resource>(
    this: T,
    props: Partial<AbstractInstanceType<T>>,
    overrides?: Partial<AbstractInstanceType<T>>,
  ) {
    if (this === Resource)
      throw new Error('cannot construct on abstract types');
    // we type guarded abstract case above, so ok to force typescript to allow constructor call
    const instance = new (this as any)(props) as AbstractInstanceType<T>;
    Object.assign(instance, props, overrides);
    // to trick normalizr into thinking we're Immutable.js does it doesn't copy
    Object.defineProperty(instance, '__ownerID', {
      value: 1337,
      writable: false,
    })
    return instance;
  }
  merge<T extends typeof Resource>(this: AbstractInstanceType<T>, other: Partial<AbstractInstanceType<T>>) {
    return (this.constructor as T).fromJS(this, other);
  }
```

ntucker:
actually may be bad idea to add any instance methods as they could potentially conflict with a member. If you just change fromJS tho you could use that to merge yourself

ntucker:
I'm excited for this change; is there anything else I can help with unblocking?

skoging:
I went with a new static method on Resource that handles merging. To make this easy to customize on Resources, it accepts an predicate as an optional third argument:
```typescript
shouldOverrideField: (fieldname: string, newValue: any, oldValue: any) => boolean
```

The default implementation will override if `newValue` is not undefined and not strictly equal to `oldValue`.

I would appreciate some feedback on this. If the approach is acceptable I'll add some documentation as well.

ntucker:
Can you provide an example use case for needing such merge logic customization? In my use cases a simple one-overrides-the-other-if-it-exists has been sufficient. I try to keep things simple if possible, but of course sometimes things need to be a bit more complicated.

skoging:
I think the biggest issue here is that the simple one-overrides-the-other-if-it-exists (in the response) logic simple isn't possible to implement as I see it. Once the response has been normalized and entities has been constructed, the information about what fields were present in the response is lost. What we're left with is the default values specified for the resource fields, or undefined for fields that have no default value. 

By default only fields that are undefined after the normalized entity has been constructed will be overriden with the value from the existing entity. This means merging only works for fields that have no default value. 

I'm not sure how best to handle this, so I figure at least allowing some form of customization is needed.

Additionally there is a strict equals check between the old and new value, so that we can keep referential equality in the cases where no field has changed in the new entity.

skoging:
Yeah, that would solve all reasonable use cases that I can think of.

ntucker:
Landed https://github.com/coinbase/rest-hooks/commit/a2c4793dc8c575f6ad676e00d6d4901a843bd222 which exposes a new `merge()` method that only copies defined members. Tell me if there's anything else I can do :)

skoging:
I'll get on it. With the merging logic added to Resource, this should be a relatively small change.

src/state/reducer.ts
```diff
@@ -13,6 +13,13 @@ type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
 }
 
+function resourceCustomizer(a: any, b: any): any {
+  if (a instanceof Resource && b instanceof Resource) {
```
ntucker: maybe just check that constructor.merge is a function? That way people can provide their own classes that still have this functionality.

src/state/reducer.ts
```diff
@@ -13,6 +13,13 @@ type Writable<T> = {
   [P in keyof T]: NonNullable<T[P]>;
 }
 
+function resourceCustomizer(a: any, b: any): any {
```
ntucker: can you make this generic so return type is same as input params?

src/state/reducer.ts
```diff
@@ -13,6 +13,13 @@ type Writable<T> = {
...
+  if (a instanceof Resource && b instanceof Resource) {
+    const Static = b.constructor as typeof Resource;
+    return Static.merge(a, b);
+  }
+}
```
ntucker: can you just return the second one as fallback?

ntucker: also this needs a test case

skoging: https://lodash.com/docs/#mergeWith

By returning undefined, `lodash.merge()` will be used. This is necessary to make sure that the `entities` object itself is merged properly.

But I can make it explicitly return undefined, and add a comment to explain it.

src/state/reducer.ts
skoging: Making this typed in any way will be misleading, as the inputs really can be anything. I'll change it to be explicitly typed as a `MergeWithCustomizer` from lodash.

src/state/reducer.ts
skoging: I'm not sure this is a good idea. I suspect there are (probably rare) edge cases that might cause issues that will be difficult to debug.

The way `Resource.merge()` is implemented currently for instance, 
it will only work if both objects are instances of Resource.

Seeing as the schema used for normalization can be customized, I'm not sure if entities are guaranteed to have a consistent type.

skoging: Maybe I'm just overthinking this, and any edge cases are rare enough to not matter.

ntucker: The data structure is {entities: [Resource type key]: { [instance pk]: Instance }}}

So the invariant is that all instances at a given point in the state tree will be the same class. If that invariant is broken there is something else wrong with the system.

src/state/reducer.ts
ntucker: that seems to be breaking the tests as it's not found in import?

src/state/reducer.ts
ntucker: Ah didn't realize that. I think a comment is sufficient no need for the explicit return.

src/state/reducer.ts
ntucker: Also if you use a user defined typeguard and have a type that just looks for merge as a function then you won't need the 'as typeof Resource'. (As generally considered a hack and should be avoided if possible)

src/state/reducer.ts
skoging: Apperantly thats caused by https://github.com/lodash/babel-plugin-lodash/issues/180

ntucker: I looked at @types/lodash and MergeWithCustomizer isn't exported. Typescript is structurally not nominally typed tho, so you don't need to use the type they provide - it simply has to match the structure.

```typescript
    // mergeWith

    type MergeWithCustomizer = { bivariantHack(value: any, srcValue: any, key: string, object: any, source: any): any; }["bivariantHack"];
```

PR #11, opened 2019-03-12
[feature] Support for custom dataExpiryLength and errorExpiryLength on resources

skoging:
This feature allows defining custom data and error lifetimes for resources and request shapes.

NetworkManager will pass on dataExpiryLength and errorExpiryLength from FetchAction. Defaults to NetworkManager values if undefined.

Static dataExpiryLength and errorExpiryLength fields on Resource, undefined by default. Passed to all request shapes.

### DIFF
--- a/docs/guides/resource-lifetime.md
+++ b/docs/guides/resource-lifetime.md
@@ -1,0 +1,30 @@
+# Custom Resource cache lifetime
+
+By default the NetworkManager specifies the liftime of data and errors in the cache.
+If some resources are longer living, or shorter living than other, the can specify their own expiry length values,
+which will be passed on to all request shape creator functions on Resource.
+
+## Examples
+
+### Long cache lifetime
+
+`LongLivingResource.ts`
+```typescript
+
+// We can now extend LongLivingResource to get a resource that will be cached for one hour
+abstract class LongLivingResource extends Resource {
+  static readonly dataExpiryLength = 60 * 60 * 1000; // one hour
+}
+```
+
+
+### Never retry on error
+
+`NoRetryResource.ts`
+```typescript
+
+// We can now extend NoRetryResource to get a resource that will never retry on network error
+abstract class NoRetryResource extends Resource {
+  static readonly errorExpiryLength = Infinity;
+}
+```
--- a/docs/guides/resource-lifetime.md
+++ b/docs/guides/resource-lifetime.md
@@ -1,6 +1,6 @@
 # Custom Resource cache lifetime
 
-By default the NetworkManager specifies the liftime of data and errors in the cache.
+By default the NetworkManager specifies the lifetime of data and errors in the cache.
 If some resources are longer living, or shorter living than other, the can specify their own expiry length values,
 which will be passed on to all request shape creator functions on Resource.
 
@@ -9,22 +9,31 @@ which will be passed on to all request shape creator functions on Resource.
 ### Long cache lifetime
 
 `LongLivingResource.ts`
-```typescript
 
+```typescript
 // We can now extend LongLivingResource to get a resource that will be cached for one hour
 abstract class LongLivingResource extends Resource {
-  static readonly dataExpiryLength = 60 * 60 * 1000; // one hour
+  static getRequestOptions() {
+    return {
+      ...super.getRequestOptions(),
+      dataExpiryLength: 60 * 60 * 1000, // one hour
+    };
+  }
 }
 ```
-
 
 ### Never retry on error
 
 `NoRetryResource.ts`
-```typescript
 
+```typescript
 // We can now extend NoRetryResource to get a resource that will never retry on network error
 abstract class NoRetryResource extends Resource {
-  static readonly errorExpiryLength = Infinity;
+  static getRequestOptions() {
+    return {
+      ...super.getRequestOptions(),
+      errorExpiryLength: Infinity,
+    };
+  }
 }
 ```
--- a/src/react-integration/__tests__/hooks.tsx
+++ b/src/react-integration/__tests__/hooks.tsx
@@ -8,6 +8,7 @@ import {
   CoolerArticleResource,
   UserResource,
   PaginatedArticleResource,
+  StaticArticleResource,
 } from '../../__tests__/common';
 import {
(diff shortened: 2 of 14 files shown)


skoging:
I couldn't find any test cases for `NetworkManager.handleFetch()`, but I can add some if that's wanted.

ntucker:
I'm gonna have to think about this one. I really want to keep requestshape as small as possible, but I definitely see how this could be useful. Something that might help me is outlining your specific use case so I get a good idea of the needs here.

Really comprehensive PR tho, thanks for the efforts!

ntucker:
How do you feel about putting these inside an optional 'options' member? 

```typescript
export interface RequestShape<
S extends Schema,
Body extends Readonly<object> | void,
> {
  readonly type: 'read' | 'mutate' | 'delete';
  fetch(url: string, body: Body): Promise<any>;
  getUrl(params: Params): string;
  readonly schema: S;
  readonly options?: {
    readonly dataExpiryLength?: number;
    readonly errorExpiryLength?: number;
  };
}
```

ntucker:
I'd be happy to accept this with the options being moved into an options section. Tell me what you think :)

skoging:
I've been thinking a bit on how caching and cache invalidation could be improved, and had some doubt about whether the changes in this PR is the right direction. That's the main reason I haven't updated this PR in a while. 

In the long term I think I'd like to see an approach more similar to how Apollo Client handles different usage scenarios with a [fetch policy](https://www.apollographql.com/docs/react/api/react-apollo#graphql-config-options-fetchPolicy).

That is something that I think deserves it's own discussion. For now, I think these changes are sufficient for our use, and will likely be useful even if bigger changes happen down the road.

docs/guides/resource-lifetime.md
```diff
@@ -0,0 +1,30 @@
+# Custom Resource cache lifetime
+
+By default the NetworkManager specifies the liftime of data and errors in the cache.
```
ntucker: typo liftime -> lifetime

src/resource/Resource.ts
```diff
@@ -190,6 +194,14 @@ export default abstract class Resource {
     return getEntitySchema(this);
   }
 
+  /** Get the entity schema defining  */
```
ntucker: This should be added to Resource.md docs

src/resource/Resource.ts
```diff
@@ -190,6 +194,14 @@ export default abstract class Resource {
...
+  static getRequestOptions<T extends typeof Resource>(this: T) {
+    return {
```
ntucker: seems like this method itself is sufficient without needing properties as well. (thus reducing api surface). Then we could just return undefined here by default which would maybe be preferable.

src/react-integration/__tests__/hooks.tsx
```diff
@@ -286,6 +290,27 @@ describe('useRetrieve', () => {
     rerender();
     expect(dispatch).toBeCalled();
   });
+  it('should dispatch with resource defined dataExpiryLength', async () => {
+    function FetchTester() {
+      useRetrieve(StaticArticleResource.singleRequest(), payload);
```
ntucker: this only tests the dispatch side; I would really appreciate some tests that touch the network manager side - hopefully that actually test the caching length.

At the very least, the `integration.tsx` test file should get a full path done so the code coverage doesn't decrease.

ntucker:
Interested to hear the use case for the apollo fetch policy. As far as I can tell these use cases are covered here:

- cache-first - set expiry time to infinity
- cache-and-network - set expiry time to 0
-network-only - useFetch()
-cache-only - useCache()
-no-cache - call Resource.fetch() yourself

And given that expiry time can be something besides 0 and infinity, there are cases not covered by apollo's cache policy (I would argue the common case).

Perhaps we should discuss this in a ticket instead tho?

src/resource/Resource.ts
skoging: What do you think: 
```typescript
// static method:
static getRequestOptions<T extends typeof Resource>(this: T): RequestOptions | undefined;

// or static field:
static requestOptions?: RequestOptions;
```

skoging: Also, perhaps this typed options object should be passed on as is in the meta of the [FetchAction](https://github.com/coinbase/rest-hooks/pull/11/files#diff-39f8fa22b8033e26276f0093acfaa3d4R43) for consistency

ntucker: it should be a function so it can be resolved polymorphically - in case it's called on its own and not by one of the other requestshape generators

ntucker: since it's called in singleRequest() i'm not sure how you'd have the meta to pass it; or you mean what singleRequest ends up returning?

skoging: I was thinking of the fetch action dispatched in useFetcher. dataExpiryLength and errorExpiryLength are destructured from the options object and passed as plain properties in the action meta.

Since the RequestShape was changed to have an options property, perhaps it would be more consistent to pass that object on directly instead of destructuring.

ntucker: Oh sure; so like `meta: { options?: { dataExpiryLength?: ....` that sounds smart

src/react-integration/__tests__/hooks.tsx
skoging: Added test cases for the middleware from NetworkManager. Covers fetch actions success/error as well as the expiry time with and without overridden dataExpiryLength or errorExpiryLength.

src/types.ts
```diff
@@ -18,6 +18,13 @@ export type State<T> = Readonly<{
   }>;
 }>;
 
+export interface RequestOptions {
```
ntucker: one final nit: can you export this from teh package itself (root index.ts)

skoging:
Yeah, think it's better to move that discussion to separate issue. I can write one up, but I think I'll take some time and see which use cases are actually need and not covered already.

src/types.ts
skoging: Done

skoging:
Thanks for the detailed reviews on these PRs. It gives me a lot of confidence in this project!

PR #12, opened 2019-03-12
Enable cross platform development

skoging:
Use cross-env and cpy-cli to make scripts run on windows.


ntucker:
I just use WSL myself, but I suppose this shouldn't be required.

PR #16, opened 2019-03-25
[feature] Add merge() to Resource

ntucker:
Keeps track of which values are actually set so partial merges can work properly.

Using static methods as instance methods will not be able to be overridden.

Enables https://github.com/coinbase/rest-hooks/pull/10


ntucker:
@skoging How do you feel about the API naming?

PR #19, opened 2019-03-26
[enhance] More refined typings for NetworkErrorBoundary

christophehurpeau:
I didn't check that typescript is correct as I edited this on github

### DIFF
--- a/src/react-integration/NetworkErrorBoundary.tsx
+++ b/src/react-integration/NetworkErrorBoundary.tsx
@@ -1,27 +1,31 @@
 import React from 'react';
 
+export interface ErrorWithStatus extends Error {
+  status: number;
+}
+
 interface Props {
   children: React.ReactNode;
-  fallbackComponent: React.ComponentType<{ error: Error }>;
+  fallbackComponent: React.ComponentType<{ error: ErrorWithStatus }>;
 }
 interface State {
-  error?: Error;
+  error?: ErrorWithStatus;
 }
 export default class NetworkErrorBoundary extends React.Component<
   Props,
   State
 > {
   static defaultProps = {
-    fallbackComponent: ({ error }: { error: any }) => (
+    fallbackComponent: ({ error }: { error: ErrorWithStatus }) => (
       <div>
         {error.status} {error.response && error.response.statusText}
       </div>
     ),
   };
 
-  static getDerivedStateFromError(error: any) {
+  static getDerivedStateFromError(error: ErrorWithStatus | any) {
     if (error.status) {
-      return { error };
+      return { error: error as ErrorWithStatus };
     }
   }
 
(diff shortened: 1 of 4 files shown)


christophehurpeau:
Thanks to the build, I didn't typed response ! Do you know what type it is ?

ntucker:
It should be ResponseError from superagent. However, we shouldn't assume people we not change what agent is used, so to support the usage of it I would just add optional response part.

```
export interface ErrorWithStatus extends Error {
  status: number;
  response?: { statusText?: string };
}
```

src/react-integration/NetworkErrorBoundary.tsx
```diff
@@ -1,27 +1,31 @@
...
-    fallbackComponent: ({ error }: { error: any }) => (
+    fallbackComponent: ({ error }: { error: ErrorWithStatus }) => (
       <div>
         {error.status} {error.response && error.response.statusText}
       </div>
     ),
   };
 
-  static getDerivedStateFromError(error: any) {
+  static getDerivedStateFromError(error: ErrorWithStatus | any) {
```
ntucker: `| any` might as well just be any. Why did you keep it around?

christophehurpeau: because we know it can be ErrorWithStatus and that's the type that interests us. if you prefer I can change it back to any ?

christophehurpeau:
Thank you for the response type, I also added `body` because it contains the response in json that we use to specify an error message

src/react-integration/NetworkErrorBoundary.tsx
ntucker: ah right so you know what you're setting.

Might I suggest a user-defined type-guard then:

```typescript
function isNetworkError(error: NetworkError | any): error is NetworkError {
    return !!(error as NetworkError).status;
}
```

then

```typescript
if (isNetworkError(error)) {
  return { error };  // after if block error will be typed as NetworkError
}
```

(read more here: https://www.typescriptlang.org/docs/handbook/advanced-types.html)

src/react-integration/NetworkErrorBoundary.tsx
```diff
@@ -1,27 +1,32 @@
...
 interface Props {
   children: React.ReactNode;
-  fallbackComponent: React.ComponentType<{ error: Error }>;
+  fallbackComponent: React.ComponentType<{ error: ErrorWithStatus }>;
 }
 interface State {
-  error?: Error;
+  error?: ErrorWithStatus;
 }
 export default class NetworkErrorBoundary extends React.Component<
```
ntucker: To make this even more reusable you might even use a generic component (https://www.mojotech.com/blog/typescript-generic-react-components/)

```typescript
interface Props<E> {
  children: React.ReactNode;
  fallbackComponent: React.ComponentType<{ error: E }>;
}
interface State<E> {
  error?: E;
}
export default class NetworkErrorBoundary<E extends ErrorWithStatus = ErrorWithStatus>
```

ntucker:
Thanks so much for the follow up on this PR, looks great :D

PR #24, opened 2019-03-27
[fix] Network error can have an undefined status

christophehurpeau:
```tsx
console.log(error, error.status, Object.prototype.hasOwnProperty.call(error, 'status'));
```

Screenshot 2019-03-27 at 13 38 01


ntucker:
I guess there's no tests for this already and this is an important fix so I won't block its inclusion

PR #29, opened 2019-03-31
[enhance] Weaken requirements for RestProvider 'manager' prop

ntucker:
The NetworkManager interface currently exposes its constructor signature, which is not a requirement for consumption in RestProvider. Therefore, providing an interface with much weaker definition could help in allowing fully custom NetworkManager plugins.


ntucker:
Realized the instance type doesn't have constructor information so this does nothing.

PR #36, opened 2019-04-07
[feature] Add polling based subscriptions

ntucker:
Integration into the data flow occurs through middleware; Managers have become a standard controlling class to provide such middlewares. They also keep a cleanup() method so they won't leave hanging async things (promises, setIntervals, etc).

This adds a new managers for subscriptions known as `SubscriptionManager`. This will take one argument in its constructor which implements the `Subscription` interface. This enables users to provide their own mechanisms for subscriptions - for instance using a websocket connection.

Provided as default is a `PollingSubscription` which uses setInterval set to the lowest frequency requested for a given resource.

Theoretically users could use multiple SubscriptionManagers to tie together multiple subscription types at once. This flexibility will come in a breaking [change later](https://github.com/coinbase/rest-hooks/issues/15#issuecomment-480911714) that makes the managers supplied completely open-ended.

`useSubscription()` hook is added to be the barebones way to subscribe. By providing it separately users are able to compose hooks as they desire.


README.md
```diff
@@ -63,7 +63,7 @@ For the small price of 7kb gziped. &nbsp;&nbsp; [🏁Get started now](docs/getti
 - [x] 🧘 Flexible to fit any API design (one size fits all)
 - [x] 🌳 Tree-shakable (only use what you need)
 - [ ] 🚯 Pluggable garbage collection policy
-- [ ] 🔁 Subscriptions
+- [x] 🔁 Subscriptions
```
mprobber: nit: maybe move this above `Pluggable garbage collection policy`, so all of the features that have been completed are grouped together

src/state/PollingSubscription.ts
```diff
@@ -0,0 +1,129 @@
...
+        this.frequencyHistogram.delete(frequency);
+
+        // this was the min, so find the next size
+        if (frequency <= this.frequency) {
+          if (this.frequencyHistogram.size === 0) return true;
+          this.frequency = min(this.frequencyHistogram.keys());
+          this.run();
+        }
+      }
+    } else if (process.env.NODE_ENV !== 'production') {
```
mprobber: What happens here if a user of `rest-hooks` doesn't have a DefinePlugin with `process.env`?

ntucker: It depends on which build they are consuming. The unpkg one gets this compiled out. The es6 module version will keep it in assuming they are using a standard webpack or rollup setup.

If these assumptions turn out to be bad I'm sure there will be issues created to think about future direction; but until those people pop up I can't really assume anything about their use case.

README.md
ntucker: 👍

PR #48, opened 2019-04-18
[feature] Add optional redux integration

ntucker:
This is tree-shakable so it won't increase bundle size at all if not uses. Could even be used only server-side to do SSR.

Added: `<ExternalCacheProvider />` as alternative to `<RestProvider />`. This enables external store as prop instead of store based on `useReducer()`.

### DIFF
--- a/src/react-integration/__tests__/integration.tsx
+++ b/src/react-integration/__tests__/integration.tsx
@@ -1,202 +1,164 @@
-import React, { Suspense } from 'react';
-import { render } from 'react-testing-library';
-import { cleanup, renderHook, act } from 'react-hooks-testing-library';
+import React from 'react';
+import { cleanup } from 'react-hooks-testing-library';
 import nock from 'nock';
 
 import { CoolerArticleResource, UserResource } from '../../__tests__/common';
 import { useResource, useFetcher } from '../hooks';
-import RestProvider from '../provider';
-import NetworkManager from '../../state/NetworkManager';
-import { FetchAction, ReceiveAction } from '../../types';
-
-class MockNetworkManager extends NetworkManager {
-  handleFetch(action: FetchAction, dispatch: React.Dispatch<any>) {
-    const mockDispatch: typeof dispatch = (v: any) => {
-      act(() => {
-        dispatch(v);
-      });
-    };
-    return super.handleFetch(action, mockDispatch);
-  }
-  handleReceive(action: ReceiveAction) {
-    act(() => {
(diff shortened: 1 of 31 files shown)


src/react-integration/provider/RestProvider.tsx
```diff
@@ -14,27 +14,20 @@ interface ProviderProps {
   manager?: NetworkManager;
   subscriptionManager?: SubscriptionManager<any>;
   initialState?: State<unknown>;
-  store?: {
```
ntucker: this turned out to be worthless

src/react-integration/__tests__/integration.tsx
```diff
@@ -1,202 +1,164 @@
...
-    {
-      id: 7342,
-      username: 'lindsey',
-      email: 'lindsey@bob.com',
-      isAdmin: true,
-    },
-  ];
-  const nested = [
-    {
+for (const makeProvider of [makeRestProvider, makeExternalCacheProvider]) {
```
ntucker: this tabbed over most of the file making it look at way more changed. truthfully only the use of createRenderRestHook was changed

src/react-integration/__tests__/subscriptions.tsx
```diff
@@ -3,110 +3,72 @@ import { cleanup, renderHook, act } from 'react-hooks-testing-library';
...
 afterEach(cleanup);
 
-describe('<RestProvider /> with subscriptions', () => {
-  const articlePayload = {
-    id: 5,
-    title: 'hi ho',
-    content: 'whatever',
-    tags: ['a', 'best', 'react'],
-  };
+for (const makeProvider of [makeRestProvider, makeExternalCacheProvider]) {
```
ntucker: this tabbed over most of the file making it look at way more changed. truthfully only the use of createRenderRestHook was changed

src/react-integration/provider/ExternalCacheProvider.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { ReactNode, useEffect, useState } from 'react';
```
ntucker: this file is the only *real* changed in code. this is mostly tests and docs as it was already built to be easy redux integration.

src/test/helper.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { Suspense } from 'react';
```
ntucker: this new 'test' folder will eventually be exported for consumers of the library to use

PR #56, opened 2019-04-29
Added invalidIfStale flag to options

zacharyfmarion:
Added invalidIfStale flag to RequestShape options. This will make `useResource` suspend if the resource is stale. For example:

```tsx
class InvalidIfStaleArticleResource extends ArticleResource {
  static getRequestOptions(): RequestOptions {
    return {
      ...super.getRequestOptions(),
      dataExpiryLength: 0,
      errorExpiryLength: 0,
      invalidIfStale: true,
    };
  }
}
```

### DIFF
--- a/docs/api/RequestShape.md
+++ b/docs/api/RequestShape.md
@@ -90,4 +90,5 @@ an effect.
 
 #### invalidIfStale: boolean
 
-Forces useResource to suspend if the data in cache is stale. Requires `dataExpiryLength` to have effect.
+Indicates stale data should be considered unusable and thus not be returned from the cache. This means
+that useResource() will suspend when data is stale even if it already exists in cache.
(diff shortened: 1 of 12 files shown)


docs/api/RequestShape.md
```diff
@@ -86,3 +87,7 @@ Custom data error lifetime for the fetched resource. Will override the value set
 
 Frequency in millisecond to poll at. Requires using [useSubscription()](./useSubscription.md) to have
 an effect.
+
+#### invalidIfStale: boolean
+
+Forces useResource to suspend if the data in cache is stale. Requires `dataExpiryLength` to have effect.
```
ntucker: how about something like, "Indicates stale data should be considered unusable and thus not be returned from the cache. This means that useResource() will suspend when data is stale even if it already exists in cache."

don't want to exclude other possible uses of this

src/types.ts
```diff
@@ -25,6 +25,8 @@ export interface RequestOptions {
   readonly errorExpiryLength?: number;
   /** Poll with at least this frequency in miliseconds */
   readonly pollFrequency?: number;
+  /** If the result is stale then we suspend instead of using the cached value */
```
ntucker: Please don't mention suspend since that's implementation specific to useResource(). This marks the cache data as considered unusable not just out of date.

PR #57, opened 2019-05-01
[feature] Add useInvalidator(); optionally pass initialState into renderRestHook

zacharyfmarion:
- Add `useInvalidator` hook that invalidates a particular item in the cache:

```ts
// Throw error on a restriction found that is caught by an error boundary
export default function useRestriction(type: RestrictionType) {
  const restrictions = useResource(
    RestrictionResource.listRequest(),
    { type },
  );
  const invalidate = useInvalidator(PolicyRestrictionResource.listRequest());

  if (restrictions.length) {
    invalidate({ type });
    throw new Error(restrictions);
  }
}
```

- Optionally pass initialState into renderRestHook:

```ts
const state = mockState([ ... ]);
const renderRestHook = createRenderRestHook(makeRestProvider);
renderRestHook(() => {
  return useResource(ArticleResource.singleRequest(), { id });
}, state)

```


src/state/reducer.ts
```diff
@@ -19,6 +19,28 @@ function isMergeable<T>(constructor: any): constructor is MergeableStatic<T> {
   return constructor && typeof constructor.merge === 'function';
 }
 
+function purgeEntity(
```
ntucker: why?

src/state/reducer.ts
```diff
@@ -33,6 +55,7 @@ export default function reducer(
   action: ActionTypes,
 ) {
   if (!state) state = initialState;
+  console.log(state.meta, action.type, action.meta);
```
ntucker: oops

src/test/providers.tsx
```diff
@@ -1,20 +1,30 @@
...
   SubscriptionManager,
   ExternalCacheProvider,
   RestProvider,
 } from '../index';
 
+// Extension of the DeepPartial type defined by Redux which handles unknown
+type DeepPartialWithUnkown<T> = {
+  [K in keyof T]?: T[K] extends unknown
+    ? any
+    : (T[K] extends object ? DeepPartial<T[K]> : T[K])
```
ntucker: this should be calling DeepPartialWithUnkown else it only works on one level

src/react-integration/hooks/useInvalidator.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { ReadShape, Schema } from '../../resource';
+import { DispatchContext } from '../context';
+
+/** Invalidate a certain item within the cache */
+export default function useInvalidator<
+Params extends Readonly<object>,
+Body extends Readonly<object> | void,
+S extends Schema
+>(selectShape: ReadShape<S, Params, Body>, params: Params | null) {
```
ntucker: can we move params to the function itself so this mirrors useFetcher?

docs/api/README.md
```diff
@@ -11,6 +11,7 @@ title: API Reference
   - [useResultCache](useResultCache.md)
   - [useSubscription](useSubscription.md)
   - [useRetrieve](useRetrieve.md)
+  - [useInvalidator](useInvalidator.md)
```
ntucker: also needs to be added to website/sidebars.json

src/state/reducer.ts
zacharyfmarion: why not? Makes reducer more readable and better function abstraction IMO. Why did I change it here? Because I was originally thinking I was going to re-use it. I can make this a separate PR if you want.

src/test/providers.tsx
zacharyfmarion: good point

src/react-integration/hooks/useInvalidator.ts
zacharyfmarion: Hmm so would it be `invalidate({ id })`? No need for body I guess

docs/api/useInvalidator.md
```diff
@@ -3,14 +3,12 @@ title: useInvalidator()
...
-  Body extends Readonly<object> | void,
-  S extends Schema
->(selectShape: ReadShape<S, Params, Body>): (params: Params) => void;
+function useInvalidator<Params extends Readonly<object>, S extends Schema>(
+  selectShape: ReadShape<S, Params, any>,
+): (params: Params) => void;
 ```
 
-Mostly useful for imperatively triggering cache invalidation, with a similar signiature as
+Mostly useful for imperatively invalidating the cache, with a similar signiature to
```
ntucker: still have signiature which is not a word.

docs/api/useInvalidator.md
```diff
@@ -0,0 +1,50 @@
...
+
+  /** Used as default options for every RequestShape */
+  static getRequestOptions(): RequestOptions {
+    return {
+      invalidIfStale: true,
+    };
+  }
+}
+```
+
```
ntucker: can you add a short sentence or two explaining why invalidIfStale is useful in conjunction? (also please link to that part of options every time you reference it)

docs/api/useInvalidator.md
```diff
@@ -0,0 +1,50 @@
...
+```tsx
+function ArticleName({ id }: { id: string }) {
+  const asset = useResource(ArticleResource.singleRequest(), { id });
+  const invalidate = useInvalidator(ArticleResource.singleRequest());
```
ntucker: nit: a better example might be creating a new hook that invalidates on unmount that wraps both useInvalidator and useEffect(). This shows people good hook patterns.

src/state/reducer.ts
ntucker: meh, I don't think it is much better either way, but I would use the type directly instead of init:

State<unknown>['entities']

ntucker:
Can you paste output from `yarn test:coverage` since coverage doesn't run on external branches?

zacharyfmarion:
Looks like my NetworkManager tests are failing, dispatch isn't being called...didn't think I touched this codepath. Any idea why that might be?
Screen Shot 2019-05-01 at 12 14 22 PM

zacharyfmarion:
Never mind, was a node versioning issue, here is the coverage:
Screen Shot 2019-05-01 at 12 20 01 PM

ntucker:
looks like you're not at 100% for useInvalidator. line 16 seems to be missed.

zacharyfmarion:
Screen Shot 2019-05-01 at 12 28 01 PM

Okay we're at 100% now

src/react-integration/hooks/useInvalidator.ts
```diff
@@ -0,0 +1,28 @@
...
+/** Invalidate a certain item within the cache */
+export default function useInvalidator<
+Params extends Readonly<object>,
+S extends Schema
+>(selectShape: ReadShape<S, Params, any>): (params: Params) => void {
+  const { getUrl } = selectShape;
+  const dispatch = useContext(DispatchContext);
+
+  const invalidateDispatcher = useCallback(
+    (params: Params) => {
```
ntucker: currently params is not typed to be null. Do you want this to be true? (see useCache for example of how nullable works)

zacharyfmarion: good catch 🤦‍♂

PR #58, opened 2019-05-01
[docs] Add testing utility docs and guides

ntucker:
Resolves https://github.com/coinbase/rest-hooks/issues/13

### DIFF
--- a/docs/api/MockProvider.md
+++ b/docs/api/MockProvider.md
@@ -1,0 +1,73 @@
+---
+title: <MockProvider />
+---
+
+```typescript
+function MockProvider({
+  children,
+  results,
+}: {
+  children: React.ReactChild;
+  results: Fixture[];
+}): JSX.Element;
+```
+
+\<MockProvider /> is a simple substitute provider to prefill the cache with fixtures so the 'happy path'
+can be tested. This is useful for [storybook](../guides/storybook.md) as well as component testing.
+
+## Arguments
+
+### results
+
+```typescript
+interface Fixture {
+  request: ReadShape<Schema, object, any>;
+  params: object;
+  result: any;
+}
+```
+
+This prop specifies the fixtures to use data from. Each item represents a fetch defined by the
+[RequestShape](./RequestShape.md) and params. `Result` contains the JSON response expected from said fetch.
+
+## Returns
+
+```typescript
+JSX.Element
+```
+
+Renders the children prop.
+
+## Example
+
+```typescript
+import { MockProvider } from 'rest-hooks/test';
+
+import ArticleResource from 'resources/ArticleResource';
+import MyComponentToTest from 'components/MyComponentToTest';
+
+const results = [
+  {
+    request: ArticleResource.listRequest(),
+    params: { prop: 10 },
+    result: [
+      {
+        id: 5,
+        content: 'have a merry christmas',
+        author: 2,
+        contributors: [],
+      },
+      {
+        id: 532,
+        content: 'never again',
+        author: 23,
+        contributors: [5],
+      },
+    ],
+  },
+];
+
+<MockProvider results={results}>
+  <MyComponentToTest />
+</MockProvider>
+```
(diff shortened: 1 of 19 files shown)


docs/api/MockProvider.md
```diff
@@ -0,0 +1,73 @@
...
+```typescript
+import { MockProvider } from 'rest-hooks/test';
+
+import ArticleResource from 'resources/ArticleResource';
+import MyComponentToTest from 'components/MyComponentToTest';
+
+const results = [
+  {
+    request: ArticleResource.listRequest(),
+    params: { prop: 10 },
```
zacharyfmarion: why is this called prop?

docs/guides/unit-testing-hooks.md
```diff
@@ -0,0 +1,50 @@
...
+  beforeEach(() => {
+    nock('http://test.com')
+      .get(`/article-cooler/0`)
+      .reply(403, {});
+    renderRestHook = makeRenderRestHook(makeRestProvider);
+  });
+  afterEach(() => {
+    renderRestHook.cleanup();
+  });
+  it('useResource() should throw errors on bad network', async () => {
```
zacharyfmarion: nit, since the describe is `useResource()` this can just say:

`it('should throw errors on bad network', async () => {`

docs/guides/unit-testing-components.md
```diff
@@ -0,0 +1,62 @@
...
+    },
+  ],
+  loading: [],
+};
+```
+
+#### `__tests__/ArticleList.tsx`
+
+```tsx
+import { shallow } from 'enzyme';
```
zacharyfmarion: does this work? I was under the impression that shallow still did not work with hooks

docs/api/MockProvider.md
ntucker: haha, i stole it from the issue. I'll make it something more meaningful

docs/guides/unit-testing-components.md
ntucker: oh hmm, i should change the test to use test-renderer then...you did point me toward a file with enzyme used :S

docs/guides/unit-testing-components.md
```diff
@@ -46,17 +46,19 @@ export default {
...
 import results from './fixtures';
 
 describe('<ArticleList />', () => {
   it('renders', () => {
-    const wrapper = shallow(
-      <MockProvider results={results}>
-        <ArticleList prop={10} />
-      </MockProvider>,
+    const tree = (
+      <MockProvider results={results.full}>
```
zacharyfmarion: MockProvider should be imported right?

docs/api/makeRenderRestHook.md
```diff
@@ -0,0 +1,103 @@
...
+
+- [makeRestProvider()](./makeRestProvider.md)
+- [makeExternalCacheProvider()](./makeExternalCacheProvider.md)
+
+## Returns
+
+```typescript
+type RenderRestHookFunction = {
+  <T>(callback: () => T, initialState?: State<unknown>): {
+    readonly result: {
```
zacharyfmarion: I feel like adding typing for the API of react-hooks-testing-library makes these docs more brittle

zacharyfmarion:
why the docusaurus downgrade?

docs/api/makeRenderRestHook.md
ntucker: But the alternative is forcing them to go to an external site which can change and is extra work to find the info. Whereas I just have to be conscious everytime I bump that dependency that i update this.

Certainly not ideal but best I can come up with.

ntucker:
Upgrade actually. I was on next before because 1.9 wasn't out yet. 1.9 is more recent than what next had before. I just needed to take advantage of my fixes ASAP

docs/api/makeRenderRestHook.md
zacharyfmarion: fair enough

PR #64, opened 2019-05-07
[enhance] Use React's scheduler to queue resolution of promises

ntucker:
RequestIdleCallback was always a hack to try to let React update context before the promises were considered resolved so it won't try to re-render until the state is available.

Since React is using scheduler to queue the context update itself, this will place the promise resolution in queue after so things will more consistently happen in the correct order.


ntucker:
Can probably can this in favor of new middleware design that lets you run things after commit.

PR #65, opened 2019-05-13
[breaking][test] renderRestHook() takes results fixtures in options instead of initialState

zacharyfmarion:
- Change name of `makeRenderRestHook` file from helper to match the name of the default export.
- Change `makeRenderRestHook` to take in an array of fixtures instead of initialState so that hooks using useCache can be properly tested.

NOTE: running `yarn eslint --fix --ext .ts,.tsx src` and `yarn format` change `makeRenderRestHook` back and forth, indenting and un-indenting part of the code.

Closes https://github.com/coinbase/rest-hooks/issues/63

### DIFF
--- a/src/test/makeRenderRestHook.tsx
+++ b/src/test/makeRenderRestHook.tsx
@@ -1,8 +1,9 @@
-import React, { Suspense } from 'react';
+import React from 'react';
 import { RenderOptions } from 'react-testing-library';
 import { renderHook } from 'react-hooks-testing-library';
 
 import { MockNetworkManager } from './managers';
+import mockInitialState, { Fixture } from './mockState';
 import {
   State,
   NetworkManager,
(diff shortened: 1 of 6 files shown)


src/test/makeRenderRestHook.tsx
```diff
@@ -23,13 +24,15 @@ export default function makeRenderRestHook(
     callback: (props: P) => R,
     options?: {
       initialProps?: P;
-      initialState?: State<unknown>;
+      results?: Fixture[];
```
ntucker: there's some docs that need updating to reflect this.

src/test/makeRenderRestHook.tsx
```diff
@@ -14,7 +15,7 @@ export default function makeRenderRestHook(
   makeProvider: (
     manager: NetworkManager,
     subManager: SubscriptionManager<any>,
-    initialState?: State<unknown>,
+    results?: State<unknown>,
```
ntucker: is this the right type?

ntucker: actually you should have just kept the same name. updated.

PR #67, opened 2019-05-27
[breaking] Don't import polyfills but assume they are already loaded

ntucker:
This makes the library more flexible to different usages.

- Different polyfill libraries can be used (helps with https://github.com/coinbase/rest-hooks/issues/25)
- Old browsers can be unsupported without adding to bundle size
- Different bundles targeting different browsers can be built.
- Potential feature-detection of polyfills

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,6 @@
   "dependencies": {
     "@babel/runtime": "^7.4.5",
     "@types/superagent": "^4.1.2",
-    "core-js": "^3.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -77,7 +77,7 @@
     }
   },
   "devDependencies": {
-    "@anansi/babel-preset": "^0.15.0",
+    "@anansi/babel-preset": "^0.16.0",
```
ntucker: this changes the default useBuiltIns to 'entry'

PR #70, opened 2019-05-28
[fix] Send most updated state to manager middleware

ntucker:
Before getState() would only ever return the initial value. This was not caught as it is never actually used by the existing managers.

We use a ref here so we can retrieve inside a mutable object.

### DIFF
--- a/src/react-integration/provider/__tests__/middleware.tsx
+++ b/src/react-integration/provider/__tests__/middleware.tsx
@@ -33,6 +33,19 @@ describe('createEnhancedReducerHook', () => {
       return next(action);
     };
   };
+  const makeStatefulMiddleware = ({
+    callBefore,
+    callAfter,
+  }: {
+  callBefore: Function;
+  callAfter: Function;
+  }) => ({ getState }: MiddlewareAPI) => {
+    return (next: any) => (action: any) => {
+      callBefore(getState());
+      next(action);
+      callAfter(getState());
+    };
+  };
 
   test('runs through a single middleware', () => {
     const faker = jest.fn();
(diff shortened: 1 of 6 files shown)


src/react-integration/provider/__tests__/middleware.tsx
```diff
@@ -139,6 +152,39 @@ describe('createEnhancedReducerHook', () => {
...
+    });
+    let [state, dispatch] = result.current;
+    expect(callBefore.mock.calls.length).toBe(0);
+    let action: any = { type: 'hi' };
+    act(() => {
+      dispatch(action);
+    });
+    [state, dispatch] = result.current;
+    expect(callBefore.mock.calls.length).toBe(1);
+    //expect(callAfter.mock.calls.length).toBe(1);
```
ntucker: i think this doesn't work because for store to be updated it not only has to finish the reducer, but actually run render in the react function.

line 16 of middleware.tsx:
```tsx
store.current.state = state;
```

I wonder if there's a mechanism to block the rest of dispatch until that commit takes place?

sylvaingi: Yes that's where `useReducer` behaves not exactly like Redux. One way to solve this is to manually call the reducer and store the state in a ref: https://github.com/sylvaingi/rest-hooks/commit/c1ad7e3fd810e97790bb2c942c95b15970288127

ntucker: Hmm, actually then if you just use scheduler (like in https://github.com/coinbase/rest-hooks/pull/64) you should be able to get the next value after it commits.

ntucker: And persisting should honestly be done async on a scheduler anyway - with a lower priority even. So probably works out fine.

PR #94, opened 2019-07-06
[breaking] getUrl() -> getFetchKey(); shape.fetch() takes params

ntucker:
Fixes https://github.com/coinbase/rest-hooks/issues/87

Make clearer for non-url based request identifier protocols like graphql and grpc


ntucker:
This does have the disadvantage of making some overrides a bit more verbose. like: https://github.com/coinbase/rest-hooks/pull/94/files#diff-46476cc1d2278f4dd419caa31bb8544bR81

docs/guides/url.md
```diff
@@ -44,7 +44,7 @@ ArticleResource.url({ id: 5 });
 ```
 
 Customizing the url patterns is typically as easy as overriding either of these
-methods. However, in more extreme scenarios, [getUrl()](../api/FetchShape#geturlparams-param-string)
+methods. However, in more extreme scenarios, [getFetchKey()](../api/FetchShape#fetchkeylparams-param-string)
```
zacharyfmarion: fetchkeylparams?

src/resource/Resource.ts
```diff
@@ -306,11 +306,11 @@ export default abstract class Resource {
...
       options,
-      getUrl(params: object) {
-        return self.url(params);
+      getFetchKey(params: object) {
+        return 'PUT ' + self.url(params);
       },
-      fetch(url: string, body: Partial<AbstractInstanceType<T>>) {
-        return self.fetch('put', url, body);
+      fetch(params: Readonly<object>, body: Partial<AbstractInstanceType<T>>) {
+        return self.fetch('put', self.url(params), body);
```
zacharyfmarion: nit: why is PUT uppercase in getFetchKey but lowercase in self.fetch?

src/state/__tests__/selectors.ts
```diff
@@ -104,7 +104,7 @@ describe('selectors', () => {
             [params.id]: article,
           },
         },
-        results: { [CoolerArticleResource.url(urlParams)]: params.id },
+        results: { [CoolerArticleResource.detailShape().getFetchKey(urlParams)]: params.id },
```
zacharyfmarion: hmmm this is annoyingly verbose, but I see why it is the way it is

docs/guides/url.md
ntucker: damn that extra l

src/resource/Resource.ts
ntucker: lowercase because it's used for method name in superagent....upper case is kinda arbitrary but i guess it seems to call it out more distinctly

PR #109, opened 2019-07-22
[enhance] useRetrieve() no longer subscribes to state

ntucker:
The cache in rest hooks keep track of its entry and expiry time - this enables fine grained controls over when it invalidates. useRetrieve() short-circuits requests if the cache is considered fresh (not stale), meaning the data expiry time is still in the future. Previously, useRetrieve() would use state context to determine this. However, this means that every component wanting to retrieve would need to subscribe to state context. While mostly this is used within useResource() so those components are already subscribed to state, this also seems against the principals of keeping orchestration logic separated from hooks.

Removing the state context in useRetrieve() means that NeworkManager will now inspect state (using getState() from middleware) to determine if the data is considered fresh. This can then short-circuit actual fetching.

A side effect of the new design means that useRetrieve() will now return a promise even when data is considered stale. This will resolve to null if a fetch doesn't actually occur (alternatives?). This necessitates useResource() becoming aware of data staleness to determine actually throwing the promise from useRetrieve() (remember if data is not stale and thus it short circuits useRetrieve() will still return a promise).

Additionally, network error handled had to be reworked a bit to be more consistent since useRetrieve()'s return value doesn't say as much. This seems like a fix really as it only worked because of the exact sequencing before.

Closes https://github.com/coinbase/rest-hooks/issues/75

Open questions:

1. useRetrieve() still returning a promise makes me a bit sad. It seems like a worse interface even if the implementation is cleaner. At the very least I would want to think about what it resolves to. People should probably be able to depend on that promise in their usage.
3. Pushing the data staleness check to useResource() seems like it's not really helping much. Now the logic is split between NetworkManager and useResource() instead of just being handled in useRetrieve(). Since useRetrieve() is simply triggering effects rather than returning a memoized value or callback the normal decoupling of state benefit to dispatch seems pointless. The only real benefit in this case seems to be ability to not subscribe to state.

TODO: rewrite https://resthooks.io/docs/guides/no-suspense to fit new design.


src/react-integration/hooks/useResource.ts
```diff
@@ -55,17 +58,33 @@ function useManyResources<A extends ResourceArgs<any, any, any>[]>(
...
+    .filter((p, i) => p && (!resources[i] || isExpiredList[i]));
+
+  // get the first error that is valid.
+  let error: null | Error = null;
+  for (let i = 0; i < resourceList.length; i++) {
+    const [fetchShape, params] = resourceList[i];
+    const resource = resources[i];
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const e = useError(fetchShape, params, resource);
+    if (e && !isExpiredList[i]) error = e;
```
nickcherry: :nit: Do we need the `error` variable here, or could we just throw if we encounter an error and `! isExpiredList[i]`?

src/react-integration/hooks/useFetcher.ts
```diff
@@ -17,7 +17,11 @@ export default function useFetcher<
   Params extends Readonly<object>,
   Body extends Readonly<object> | void,
   S extends Schema
->(fetchShape: FetchShape<S, Params, Body>, throttle = false) {
+>(
+  fetchShape: FetchShape<S, Params, Body>,
+  throttle = false,
+  onlyIfStale = false,
```
zacharyfmarion: not really in love with this variable name, doesn't really say _what_ is happening only if stale.

src/react-integration/hooks/useResource.ts
```diff
@@ -55,17 +58,33 @@ function useManyResources<A extends ResourceArgs<any, any, any>[]>(
...
       S extends Schema
-    >([select, params]: ResourceArgs<S, Params, Body>) =>
+    >([fetchShape, params]: ResourceArgs<S, Params, Body>) =>
       // eslint-disable-next-line react-hooks/rules-of-hooks
-      useCache(select, params),
+      useCache(fetchShape, params),
+  );
+  const isExpiredList = resourceList.map(([fetchShape, params]) =>
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useIsExpired(fetchShape, params),
```
zacharyfmarion: should probably add some user warnings if the length of resource list ever changes, given that the rules of hooks are being broken

src/react-integration/hooks/useResource.ts
```diff
@@ -11,17 +13,17 @@ type ResourceArgs<
 > = [ReadShape<S, Params, Body>, Params | null];
 
 /** If the invalidIfStale option is set we suspend if resource has expired */
-function hasUsableData<
-  S extends Schema,
+function useIsExpired<
```
zacharyfmarion: nit: seeing as this moved around, can we just put this in a separate file?

zacharyfmarion:
not a fan of the whole null resolving thing either but can't think of a way around it

ntucker:
Closing in favor of https://github.com/coinbase/rest-hooks/pull/149

* Since context is not used outside hook react likely short-circuits anyway
* A better solution would be to pass down the promise container if we choose to go down this route.

PR #111, opened 2019-07-24
[docs] Replace SchemaBase with SchemaDetail

fnky:
I found that the docs were out of date, using the latest release (2.0.3) when it came to using `SchemaBase`. The migration page says that the interface was renamed and so the docs were not correct.

This PR replaces references of the outdated interface `SchemaBase` to `SchemaDetail` in the docs.

### DIFF
--- a/docs/guides/endpoints.md
+++ b/docs/guides/endpoints.md
@@ -180,14 +180,14 @@ we'll be extended that schema definition.
 import {
   Resource,
   MutateShape,
-  SchemaBase,
+  SchemaDetail,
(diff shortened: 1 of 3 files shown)


fnky:
I don't know your release pipeline for docs, but I assume that it will build the docs and publish once the PR has been merged. Is that correct?

ntucker:
Sadly docusaurus versioning requires updating the 2.0 docs as well but I'll just copy these over after merge.

Thanks so much for this contribution!

PR #113, opened 2019-07-27
[feature] Enable customizing networking library reducing bundle size

### LINKED ISSUE #103: Bundle size incorrect
Hi there, this is an interesting looking package, but I think your bundle size in the readme may be a little misleading.

Given that you have a few dependencies that other people may not have (like superagent and reselect, for example), the bundle size is actually quite a bit higher than advertised: : https://bundlephobia.com/result?p=rest-hooks@2.0.1

ntucker:
Fixes #103 .

### Motivation
The core of rest-hooks clocks in at around 7kb gzip. However, there is one part - superagent - which is quite sizable. If someone decides to override their networking library it won't get tree-shaked out due to it being included in the Resource class. This results in what is essentially a doubling of this library's impact on their bundle size.

### Solution
Splitting out the fetch implementation to a derived class allows those implementing their own fetch solution to not have to import the superagent fetch solution. This means with tree-shaking superagent will never even be included in the bundle.

Will have a follow blog post to talk about bundle sizes to clear things up since bundlephobia is pretty misleading and we have no desire to make the rest-hooks library worse just to rig artificial metrics that have no bearing in real life.


src/resource/SimpleResource.ts
```diff
@@ -0,0 +1,366 @@
+import { memoize } from 'lodash';
```
ntucker: this is actually just copy/paste of Resource but with rename of class.

src/resource/SimpleResource.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  /** SimpleResource factory. Takes an object of properties to assign to SimpleResource. */
+  static fromJS<T extends typeof SimpleResource>(
+    this: T,
+    props: Partial<AbstractInstanceType<T>>,
+  ) {
+    // we type guarded abstract case above, so ok to force typescript to allow constructor call
+    const instance = new (this as any)(props) as AbstractInstanceType<T>;
+
+    if (instance.pk === undefined)
```
ntucker: except this part was changed. This is better detection because there can be multiple abstract classes in the hierarchy beyond this.

ntucker:
I'm very open to naming suggestions for the base class. (Known now as `SimpleResource`)

PR #121, opened 2019-08-09
[fix] Allow Resource to have urlRoots without a trailing slash

### LINKED ISSUE #81: Consider adding a check for trailing slash in  `this.urlRoot`
The API I use doesn't accept trailing slashes. Thus, the default string concatenation in `url()` doesn't work for me.

https://github.com/coinbase/rest-hooks/blob/17bc481aa2b55ebe087d7fdecd663f6a528f0cf6/src/resource/Resource.ts#L174

A simple if can check if the last character of `this.urlRoot` is a slash and act accordingly.

**Describe alternatives you've considered**
I'm currently overriding url as a workaround. I think my use case is common enough to be contributed upstream, as explained above.

I can provide a PR if there is interest.

WilhelmWillie:
Fixes #81 

### Motivation
See issue: https://github.com/coinbase/rest-hooks/issues/81

### Solution
Simple conditionals!

--- 

Always wanted to submit an open source PR, saw issue #81, seemed like a good start :) Appreciate any feedback

### DIFF
--- a/src/resource/__tests__/resource.ts
+++ b/src/resource/__tests__/resource.ts
@@ -37,7 +37,6 @@ describe('Resource', () => {
     );
   });
   it('should render url property', () => {
-    CoolerArticleResource.urlRoot = 'http://test.com/article-cooler';
     const article = CoolerArticleResource.fromJS({
       id: 5,
       title: 'happy',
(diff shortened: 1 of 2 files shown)


src/resource/__tests__/resource.ts
```diff
@@ -37,13 +37,26 @@ describe('Resource', () => {
     );
   });
   it('should render url property', () => {
+    CoolerArticleResource.urlRoot = 'http://test.com/article-cooler';
```
ntucker: is this supposed to be here?

WilhelmWillie: whoops nope

PR #122, opened 2019-08-13
[pkg] Upgrade to @testing-library packages

ntucker:
Optional peerDeps have been added to reflect the needed testing library versions. These are optional because you don't have to use the testing part of rest-hooks.

Future work: 'rest-hooks/test' should be its own package so breaking changes can be separated.


src/react-integration/provider/__tests__/middleware.tsx
```diff
@@ -1,7 +1,13 @@
...
-import { unstable_scheduleCallback } from 'scheduler';
-import { cleanup, renderHook, act } from 'react-hooks-testing-library';
+import {
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  unstable_advanceTime,
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  unstable_scheduleCallback,
+  // eslint-disable-next-line @typescript-eslint/camelcase
+  unstable_NormalPriority,
+} from 'scheduler/unstable_mock';
```
ntucker: all testing with scheduler is done via this mock module

ntucker:
build currently failing because coveralls is down

website/versioned_docs/version-2.1/api/makeRenderRestHook.md
```diff
@@ -0,0 +1,124 @@
+---
```
ntucker: these new files are just because they didn't change from 2.0 previously so it didn't create a copy

src/react-integration/provider/__tests__/middleware.tsx
ntucker: https://github.com/facebook/react/pull/14964 << context

PR #133, opened 2019-09-10
[enhance] Use own deepmerge instead of lodash.mergeWith

ntucker:
### Motivation
Lodash.mergeWith is pretty bloated. Custom solution will be much more efficient.

### Solution
Integrated deepmerge that detects .merge members

### DIFF
--- a/src/react-integration/NetworkErrorBoundary.tsx
+++ b/src/react-integration/NetworkErrorBoundary.tsx
@@ -5,7 +5,7 @@ export interface NetworkError extends Error {
   response?: { statusText?: string; body?: any };
 }
 
-function isNetworkError(error: NetworkError | any): error is NetworkError {
+function isNetworkError(error: NetworkError | unknown): error is NetworkError {
   return Object.prototype.hasOwnProperty.call(error, 'status');
 }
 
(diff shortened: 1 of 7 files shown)


src/react-integration/NetworkErrorBoundary.tsx
```diff
@@ -5,7 +5,7 @@ export interface NetworkError extends Error {
   response?: { statusText?: string; body?: any };
 }
 
-function isNetworkError(error: NetworkError | any): error is NetworkError {
+function isNetworkError(error: NetworkError | unknown): error is NetworkError {
```
ntucker: unrelated - just snuck this in

src/state/__tests__/reducer.ts
```diff
@@ -2,97 +2,14 @@ import {
...
 import {
   FetchAction,
   RPCAction,
   ReceiveAction,
   PurgeAction,
   InvalidateAction,
 } from '../../types';
-import { mergeWith } from 'lodash';
-
-describe('resourceCustomizer', () => {
```
ntucker: these tests are slightly changed and moved into the general merge code

src/react-integration/NetworkErrorBoundary.tsx
nickcherry: sneaky

PR #135, opened 2019-09-15
[enhance] Fully support Union and Values schemas in types and useDenormalized()

ntucker:
### Motivation
Should support full range of schema types. Types before were sloppy and non-specific.

### Solution
- Adding tests for both Values and Unions in useDenormalized()
- Fleshing out all cases for typescript types - including resolving Denormalized<S> and ResultType<S>

### DIFF
--- a/src/__tests__/common.ts
+++ b/src/__tests__/common.ts
@@ -162,6 +162,59 @@ export class PaginatedArticleResource extends OtherArticleResource {
   }
 }
 
+export class UnionResource extends Resource {
+  readonly id: string = '';
+  readonly body: string = '';
+  readonly type: string = '';
+
+  pk() {
+    return this.id;
+  }
+
+  static urlRoot = '/union/';
+
+  static detailShape<T extends typeof Resource>(
+    this: T,
+  ): ReadShape<SchemaDetail<AbstractInstanceType<T>>> {
+    const schema = new schemas.Union(
+      {
+        first: FirstUnionResource.getEntitySchema(),
+        second: SecondUnionResource.getEntitySchema(),
+      },
+      'type',
+    );
+    return {
+      ...super.detailShape(),
+      schema,
+    };
+  }
+  static listShape<T extends typeof Resource>(
+    this: T,
+  ): ReadShape<SchemaList<AbstractInstanceType<T>>> {
+    const schema = [
+      new schemas.Union(
+        {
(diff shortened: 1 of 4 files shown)


src/__tests__/common.ts
```diff
@@ -162,6 +162,59 @@ export class PaginatedArticleResource extends OtherArticleResource {
...
+  static listShape<T extends typeof Resource>(
+    this: T,
+  ): ReadShape<SchemaList<AbstractInstanceType<T>>> {
+    const schema = [
+      new schemas.Union(
+        {
+          first: FirstUnionResource.getEntitySchema(),
+          second: SecondUnionResource.getEntitySchema(),
+        },
+        (input: FirstUnionResource | SecondUnionResource) => input['type'],
```
ntucker: just making sure both forms work here. normally this would be the same as the union for detailShape()

src/state/selectors/buildInferredResults.ts
```diff
@@ -16,11 +16,24 @@ export default function buildInferredResults<
...
+    Array.isArray(schema) ||
+    schema instanceof schemas.Values
+  ) {
     // array schemas should not be inferred because they're likely to be missing many members
+    // Values cannot be inferred because they have aribtrary keys
     return null;
   }
-  // TODO: handle other schemas besides array and object
+  if (schema instanceof schemas.Union) {
+    const discriminatedSchema = schema.inferSchema(params, undefined, '');
```
ntucker: this requires sending the type discriminator which may not always be natural, but can be an easy addition to optimize specific scenarios

src/resource/normal.d.ts
```diff
@@ -25,25 +35,41 @@ declare namespace schemas {
...
+        | (Choices[keyof Choices] extends Entity<infer T> ? keyof T : never)
+        | SchemaFunction<keyof Choices>,
     );
     define(definition: Schema): void;
+    inferSchema: SchemaAttributeFunction<Choices[keyof Choices]>;
+    getSchemaAttribute: SchemaFunction<keyof Choices>;
   }
 
-  export class Values<T = any> {
+  export class Values<Choices extends EntityMap | Schema = any> {
```
ntucker: values have these two forms (https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#valuesdefinition-schemaattribute) so we end up with many ternary type operators when using it to distinguish between the two.

PR #136, opened 2019-09-15
[feature] Add useResourceNew() and useCacheNew()

ntucker:
### Motivation
See https://github.com/coinbase/rest-hooks/pull/134 for more details.

Provide an easy migration path to new behavior of selectors that no longer 'dive into entities'.

### Solution
- Provide new implementation with 'New' postfix in 2.2
- Add warnings at start of all affected APIs informing of migration path
- 3.0 Rename new hooks to the default ones, rename old hooks to 'Legacy' postfix
- 3.1 Get rid of all the extra names; also get rid of `useResultCache()`


src/react-integration/__tests__/fixtures.ts
```diff
@@ -0,0 +1,78 @@
...
+      id: 23,
+      username: 'charles',
+      email: 'bob@bob.com',
+    },
+  },
+];
+
+describe('fixtures', () => {
+  it('should pass', () => {})
+});
```
nickcherry: Is this intentional?

ntucker: ahaha. ya jest expects the files in __tests__ to have some test in there. didn't want to waste too much time figuring out another way around this, but would be open to alternatives. this was an easy fix for now

PR #139, opened 2019-09-19
[breaking] Make middleware async - enabling execution after commit

ntucker:
### Motivation
React useReducer() operates differently than redux; thus needing a different middleware API. The primary difference being the async nature. State is committed async, and thus using as synchronous API doesn't allow for doing things after the state is committed or even knowing what the final state is.

### Solution
Promises make this all very nice. The dispatch function now returns a promise that resolves to void. This is reflected in next(action) in the middleware API, as this eventually calls through to the promise. This also means all middleware functions must pass through that promise to work. Thankfully typescript will enforce this and help ensure this is properly followed.

This also means the NetworkManager's middleware has been cleaned up as it can now easily resolve its promises only after the commit stage.

```typescript
({ getState }: MiddlewareAPI) => {
    return (next: any) => async (action: any) => {
      const beforeState = getState();
      await next(action);
      const afterState = getState();
    };
  };
```

### DIFF
--- a/docs/api/Manager.md
+++ b/docs/api/Manager.md
@@ -15,22 +15,30 @@ or unresolved Promises; and finally getMiddleware() - providing the mechanism to
 the flux data flow.
 
 ```typescript
-type ReduxMiddleware = <R extends React.Reducer<any, A>, A extends Actions>({
+type Dispatch<R extends React.Reducer<any, any>> = (action: React.ReducerAction<R>) => Promise<void>;
+
+type Middleware = <R extends React.Reducer<any, A>, A extends Actions>({
   dispatch,
 }: MiddlewareAPI<R>) => (
-  next: React.Dispatch<React.ReducerAction<R>>,
-) => (action: Actions) => void;
+  next: Dispatch<R>,
+) => Dispatch<R>;
 
 interface Manager {
-  getMiddleware<T extends Manager>(this: T): ReduxMiddleware;
+  getMiddleware<T extends Manager>(this: T): Middleware;
   cleanup(): void;
 }
 ```
 
 ## getMiddleware()
 
-getMiddleware() returns a function that is 100% redux compatible. This enables it to be integrated into redux,
-or used by the internal useReducer() enhancer in <CacheProvider />.
+getMiddleware() returns a function that very similar to a [redux middleware](https://redux.js.org/advanced/middleware).
+The only differences is that the `next()` function returns a `Promise`. This promise resolves when the reducer update is
+[committed](https://indepth.dev/inside-fiber-in-depth-overview-of-the-new-reconciliation-algorithm-in-react/#general-algorithm)
+when using <CacheProvider />. This is necessary since the commit phase is asynchronously scheduled. This enables building
+managers that perform work after the DOM is updated and also with the newly computed state.
+
+Since redux is fully synchronous, an adapter must be placed in front of Rest Hooks style middleware to
+ensure they can consume a promise. Conversely, redux middleware must be changed to pass through promises.
 
 Middlewares will intercept actions that are dispatched and then potentially dispatch their own actions as well.
 To read more about middlewares, see the [redux documentation](https://redux.js.org/advanced/middleware).
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,7 @@ import {
   schemas,
 } from './resource';
 import NetworkManager from './state/NetworkManager';
+import RIC from './state/RIC';
 import PollingSubscription from './state/PollingSubscription';
 import SubscriptionManager from './state/SubscriptionManager';
 import reducer, { initialState } from './state/reducer';
@@ -46,6 +47,8 @@ import {
   ReceiveAction,
   RPCAction,
   PurgeAction,
+  Dispatch,
+  MiddlewareAPI,
   Middleware,
   Manager,
 } from './types';
@@ -55,6 +58,7 @@ const __INTERNAL__ = {
   initialState,
   StateContext,
   DispatchContext,
+  RIC,
 };
 
 export type DeleteShape<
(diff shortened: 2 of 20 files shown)


src/react-integration/provider/__tests__/middleware.tsx
```diff
@@ -47,16 +39,10 @@ describe('createEnhancedReducerHook', () => {
...
       callBefore(getState());
-      next(action);
-      // eslint-disable-next-line @typescript-eslint/camelcase
-      unstable_scheduleCallback(
-        // eslint-disable-next-line @typescript-eslint/camelcase
-        unstable_NormalPriority,
-        () => callAfter(getState()),
-        {},
-      );
+      await next(action);
```
zacharyfmarion: nice

src/react-integration/provider/middleware.ts
```diff
@@ -12,27 +14,53 @@ export default function createEnhancedReducerHook(
...
+      if (dispatchPromiseRef.current) {
+        dispatchPromiseRef.current.resolve();
+        dispatchPromiseRef.current = null;
+      }
+    }, [state]);
+
+    const dispatchWithPromise = useCallback(
+      (action: React.ReducerAction<R>) => {
+        if (!dispatchPromiseRef.current) {
+          const holder: any = {};
```
zacharyfmarion: not a very descriptive variable

src/index.ts
```diff
@@ -98,6 +100,10 @@ export type RPCAction<
 > = RPCAction<Payload>;
 export type PurgeAction = PurgeAction;
 
+export type Dispatch<R extends React.Reducer<any, any>> = Dispatch<R>;
+export type MiddlewareAPI<
```
ntucker: empowering people to build their own middlewares

src/react-integration/provider/__tests__/middleware.tsx
```diff
@@ -181,23 +167,21 @@ describe('createEnhancedReducerHook', () => {
...
     [state, dispatch] = result.current;
     expect(callBefore.mock.calls.length).toBe(1);
     expect(callAfter.mock.calls.length).toBe(1);
     expect(callBefore.mock.calls[0][0]).toEqual({ counter: 0 });
     expect(callAfter.mock.calls[0][0]).toEqual({ counter: 1 });
     expect(callAfter.mock.calls[0][0]).toEqual(state);
     action = { type: 'dispatch', payload: 5 };
-    act(() => {
-      dispatch(action);
+    await act(() => {
```
ntucker: even the tests look better with the new api

src/react-integration/provider/middleware.ts
```diff
@@ -12,28 +11,62 @@ export default function createEnhancedReducerHook(
...
+      dispatch = compose(chain)(dispatchWithPromise);
       return dispatch;
-    }, [realDispatch, store]);
+    }, [dispatchWithPromise]);
     return [state, outerDispatch];
   };
   return useEnhancedReducer;
 }
+
+function NewPromiseHolder(): PromiseHolder {
```
ntucker: maybe i should be using this in the NetworkManager as well... hmm

src/state/NetworkManager.ts
```diff
@@ -108,23 +114,18 @@ export default class NetworkManager implements Manager {
...
+      if (action.error) {
+        promiseHandler = this.rejectors[action.meta.url];
+      } else {
+        promiseHandler = this.resolvers[action.meta.url];
       }
-    };
-    // TODO: this should call after the reducer has been updated
-    // in all concurrent fibers
-    RIC(completePromise, { timeout: 1000 });
+      promiseHandler(action.payload);
```
ntucker: this code is not only prettier...it's now guaranteed bug-free in concurrent mode.

src/react-integration/provider/middleware.ts
```diff
@@ -12,28 +11,62 @@ export default function createEnhancedReducerHook(
...
-    const store = useRef(state);
-    store.current = state;
+
+    useEffect(() => {
+      stateRef.current = state;
+      if (dispatchPromiseRef.current) {
+        dispatchPromiseRef.current.resolve();
+        dispatchPromiseRef.current = null;
+      }
+    }, [state]);
```
mprobber: I'm not sure this will work.

Because this effect is ran when `state` is updated, and the # of times `state` is updated doesn't necessarily equal the number of times `dispatch` is called due to batched state changes, you'll end up with some `dispatch` promises that never get resolved.

To get around this, you can wrap the `dispatch` function so it adds a promise to the action sent to the reducer and returns that promise, then wrap the reducer so it resolves that action with the new state after it's run.

ntucker: Batching works fine because it only creates new promises when one doesn't exist. It will reuse the existing promise for all dispatches until the next commit.

mprobber: So here's a concern I have - imagine the following scenario

1. function `f` calls dispatch, attaches a callback to the promise.
2. In the same event loop iteration, function `k` calls dispatch, attaches a callback to the promise.
3. `f`'s state change is flushed,
4. The state is now updated for `f`, but `f` and `k` have the same promise, so `k`'s promise must be resolved, but `k`s update hasn't happened yet.

docs/api/Manager.md
```diff
@@ -15,22 +15,30 @@ or unresolved Promises; and finally getMiddleware() - providing the mechanism to
...
-getMiddleware() returns a function that is 100% redux compatible. This enables it to be integrated into redux,
-or used by the internal useReducer() enhancer in <CacheProvider />.
+getMiddleware() returns a function that very similar to a [redux middleware](https://redux.js.org/advanced/middleware).
+The only differences is that the `next()` function returns a `Promise`. This promise resolves when the reducer update is
+[committed](https://indepth.dev/inside-fiber-in-depth-overview-of-the-new-reconciliation-algorithm-in-react/#general-algorithm)
+when using <CacheProvider />. This is necessary since the commit phase is asynchronously scheduled. This enables building
+managers that perform work after the DOM is updated and also with the newly computed state.
+
+Since redux is fully synchronous, an adapter must be placed in front of Rest Hooks style middleware to
+ensure they can consume a promise. Conversely, redux middleware must be changed to pass through promises.
```
kmartinezmedia: Nice explanation!

PR #141, opened 2019-09-20
[feature] useCacheNew() to better support extra data like pagination cursors

### LINKED ISSUE #100: Improve pagination interface (useResultCache)
Specify defaults in schema:

```typescript
const schema = { results: [this.getEntitySchema()], nextPage: '', prevPage: '' }
```

So when using result cache:

```typescript
  const { nextPage, prevPage } = useResultCache(
    ArticleResource.listRequest(),
    {},
    { nextPage: '', prevPage: '' }
  );
```
can turn into

```typescript
const { nextPage, prevPage } = useResultCache(ArticleResource.listRequest(), {});
```

Blocked on: https://github.com/paularmstrong/normalizr/pull/397

ntucker:
Fixes #100 .

### Motivation

With the new hooks returning the original shape, we can easily provide defaults for primitives in the schema, thus eliminating the need for the extra (and awkward) useResultCache().

### Solution
Instead of returning null when any piece is missing, always return the same basic shape and just put holes for the parts missing.

- Use schema primitives as default values when inferring
- In *new* hooks, fill all missing pieces with undefined instead of null
- Helper hooks hasUsableData() and useError() now take a boolean instead of resource (since before this was just a null check.
- Use new version of @rest-hooks/normalizr to track whether it was able to find all entities.
  - This is used to determine whether it 'hasUsableData' and thus should suspend.
  - However, if results were actually found in cache (isn't inferring them), will always consider data usable (so deleted entities is fine)
  - This was not needed in legacy versions due to the 'dive in' (now known as useSchemaSelect()) stopping when it hits unfound items.
- Moving the array entities filter into @rest-hooks/normalizr (to support optimistic deletes)
- Typing for ResultType<> and Denormalized<> are now designed to handle the generic (union of many types) definition of base SimpleResource FetchShapes by greedily 'choosing' the simplest part of schema unions.
  - In the future the best route is to probably have default shapes be optional mixins.


src/react-integration/hooks/hasUsableData.ts
```diff
@@ -1,14 +1,11 @@
...
 /** If the invalidIfStale option is set we suspend if resource has expired */
 export default function hasUsableData<
   S extends Schema,
   Params extends Readonly<object>,
   Body extends Readonly<object | string> | void
->(
-  resource: RequestResource<FetchShape<S, Params, Body>> | null,
-  fetchShape: Pick<FetchShape<S, Params, Body>, 'options'>,
-) {
+>(resource: boolean, fetchShape: Pick<FetchShape<S, Params, Body>, 'options'>) {
```
zacharyfmarion: pls change name

src/state/selectors/useDenormalized.ts
```diff
@@ -26,15 +26,17 @@ export default function useDenormalized<
   }: Pick<ReadShape<S, Params, any>, 'schema' | 'getFetchKey'>,
   params: Params | null,
   state: State<any>,
-): Denormalized<typeof schema> | null {
+): [
+  DenormalizedNullable<typeof schema>,
+  typeof params extends null ? false : boolean,
+] {
```
zacharyfmarion: can you add comment describing second return value

src/react-integration/hooks/useError.ts
```diff
@@ -11,7 +11,7 @@ export default function useError<
 >(
   fetchShape: ReadShape<S, Params, Body>,
   params: Params | null,
-  resource: RequestResource<typeof fetchShape> | null,
+  resource: boolean,
```
zacharyfmarion: same here

src/react-integration/hooks/useResourceNew.ts
```diff
@@ -21,46 +22,46 @@ function useOneResource<
...
   const maybePromise = useRetrieve(fetchShape, params);
-  // resource is null when it is not in cache or params is null
-  const resource = useCacheNew(fetchShape, params);
-  const error = useError(fetchShape, params, resource);
+  const state = useContext(StateContext);
+  const [denormalized, ready] = useDenormalized(fetchShape, params, state);
+  const error = useError(fetchShape, params, ready);
 
-  if (!hasUsableData(resource, fetchShape) && maybePromise) throw maybePromise;
+  if (!hasUsableData(ready, fetchShape) && maybePromise) throw maybePromise;
```
zacharyfmarion: do you need to check the falsiness of maybePromise here

src/react-integration/hooks/useResourceNew.ts
```diff
@@ -21,46 +22,46 @@ function useOneResource<
...
+  return denormalized as any;
 }
 
 /** many form resource */
 function useManyResources<A extends ResourceArgs<any, any, any>[]>(
   ...resourceList: A
 ) {
-  const resources = resourceList.map(
+  const state = useContext(StateContext);
+  const denormals = resourceList.map(
```
zacharyfmarion: denormalizedValues?

src/state/selectors/buildInferredResults.ts
```diff
@@ -9,46 +9,46 @@ import { Schema, schemas } from '~/resource/normal';
...
     // Was unable to infer the entity's schema from params
-    if (discriminatedSchema === undefined) return null;
+    if (discriminatedSchema === undefined) return undefined as any;
     return {
       id: buildInferredResults(discriminatedSchema, params),
       schema: schema.getSchemaAttribute(params, parent, ''),
     } as any;
   }
+  if (schema instanceof schemas.Array || Array.isArray(schema)) {
+    return undefined as any;
```
zacharyfmarion: Note: We talked, needs to be updated

src/react-integration/hooks/hasUsableData.ts
ntucker: also changed in useError()

src/react-integration/hooks/useResourceNew.ts
ntucker: confirmed that I do

PR #146, opened 2019-09-25
[fix] useRetrieve(): Only refetch on param changes

ntucker:
### Motivation
Previously, useRetrive() (and thus also useResource()) would trigger fetches anytime the component it was used in re-rendered for any reason - so long as the data was considered stale. This is not the designated behavior.

* Breaks idempotence of render()
* Which resulted in unpredictable fetches

We were previously syncing to `dataStale` for our fetch trigger. This is bad because it is derivative data of Date.now() - which is not idempotent. By the transitive property, any component using this hook not idempotent. **General rule: only sync with values that are derived from React-controlled state (context, props, useState) and/or things that are constant.**

Issue discovered by @saivann

### Solution
We need to move the check for Date.now() inside our hook callback so we don't need to sync on it. We can still sync on expiresAt - which enables us to still refetch on any `useInvalidator()` calls.

### DIFF
--- a/src/react-integration/__tests__/fixtures.ts
+++ b/src/react-integration/__tests__/fixtures.ts
@@ -74,5 +74,5 @@ export const nested = [
 ];
 
 describe('fixtures', () => {
-  it('should pass', () => {})
+  it('should pass', () => {});
 });
--- a/src/react-integration/__tests__/hooks.tsx
+++ b/src/react-integration/__tests__/hooks.tsx
@@ -40,8 +40,7 @@ async function testDispatchFetch(
     </DispatchContext.Provider>
   );
   render(tree);
-  expect(dispatch).toHaveBeenCalled();
-  expect(dispatch.mock.calls.length).toBe(payloads.length);
+  expect(dispatch).toHaveBeenCalledTimes(payloads.length);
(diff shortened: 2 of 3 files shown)


src/react-integration/__tests__/hooks.tsx
```diff
@@ -398,6 +403,43 @@ describe('useRetrieve', () => {
     }
     await testDispatchFetch(FetchTester, [payload]);
   });
+
+  it('should not refetch after expiry and render', async () => {
+    let time = 1000;
+    global.Date.now = jest.fn(() => time);
```
ntucker: it's a faaaaake!

src/react-integration/__tests__/fixtures.ts
```diff
@@ -74,5 +74,5 @@ export const nested = [
 ];
 
 describe('fixtures', () => {
-  it('should pass', () => {})
+  it('should pass', () => {});
```
zacharyfmarion: ?

ntucker: formatting fix. This was discussed in another pr as to why there's an empty test. (basically so I can put a file in __tests__

PR #147, opened 2019-09-25
Update auth.md to point to superagent import

mjmaurer:
Was pointing at rest-hooks


ntucker:
Request actually is exported from Rest Hooks: https://github.com/coinbase/rest-hooks/blob/master/src/index.ts#L89

Do you think it would be better to suggest people to import from superagent even tho many of them will not have installed it themselves already?

mjmaurer:
Whoops missed that. No I think it's better as is.

PR #149, opened 2019-09-25
[enhance] Improve useResource() error handling conditions

ntucker:
### Motivation
https://github.com/coinbase/rest-hooks/pull/109 incorporated some useful improvements to error mechanisms. However, after careful consideration it seems like not the best approach, so this should incorporate those improvements and some others.

### Solution
* Never generate errors when params are null (previously could create fake 404 when resource wasn't found)
* Improve some typings to factor in conditional of param being null
* multi-resource requests now immediately throw error on first error and stop suspending.

### DIFF
--- a/src/react-integration/__tests__/useResourceNew.tsx
+++ b/src/react-integration/__tests__/useResourceNew.tsx
@@ -137,23 +137,20 @@ describe('useResourceNewNew()', () => {
         );
         return article;
       } catch (e) {
+        // TODO: we're not handling suspense properly so react complains
+        // When upgrading test util we should be able to fix this as we'll suspense ourselves.
         if (typeof e.then === 'function') {
           return e;
         } else {
-          // TODO: we're not handling suspense properly so react complains
-          // When upgrading test util we should be able to fix this as we'll suspense ourselves.
-          if (e.name === 'Invariant Violation') {
-            return null;
-          } else {
-            throw e;
-          }
+          throw e;
         }
       }
     }
-    const { rerender, result, waitForNextUpdate } = renderRestHook(
-      MultiResourceTester,
-    );
+    const { rerender, result } = renderRestHook(MultiResourceTester);
     const firstPromise = result.current;
+    expect(firstPromise).toBeDefined();
+    expect(typeof firstPromise.then).toBe('function');
+
     jest.advanceTimersByTime(50);
     rerender();
     expect(result.current).toBe(firstPromise);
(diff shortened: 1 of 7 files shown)


src/react-integration/hooks/useResourceNew.ts
```diff
@@ -55,18 +42,27 @@ function useManyResources<A extends ResourceArgs<any, any, any>[]>(
...
     // only wait on promises without results
     .map((p, i) => !hasUsableData(resources[i], resourceList[i][0]) && p);
 
+  // throw first valid error
+  for (let i = 0; i < resourceList.length; i++) {
+    const [fetchShape, params] = resourceList[i];
+    const resource = resources[i];
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const error = useError(fetchShape, params, resource);
+    if (error && !promises[i]) throw error;
```
ntucker: this logic is now identical to the singular case, since we store the same conditional in this map.

src/react-integration/__tests__/useResourceNew.tsx
```diff
@@ -137,23 +137,20 @@ describe('useResourceNewNew()', () => {
         );
         return article;
       } catch (e) {
+        // TODO: we're not handling suspense properly so react complains
+        // When upgrading test until we should be able to fix this as we'll suspense ourselves.
```
zacharyfmarion: not sure this english makes sense?

src/react-integration/hooks/useResource.ts
```diff
@@ -1,49 +1,36 @@
...
-    typeof maybePromise.then === 'function'
-  )
-    throw maybePromise;
   const error = useError(fetchShape, params, resource);
+
+  if (!hasUsableData(resource, fetchShape) && maybePromise) throw maybePromise;
   if (error) throw error;
 
-  return resource as NonNullable<typeof resource>;
+  return resource as any;
```
zacharyfmarion: seems like a type regression, why did you do this?

src/react-integration/hooks/useResourceNew.ts
```diff
@@ -1,49 +1,36 @@
...
->(
-  resource: RequestResource<ReadShape<S, Params, Body>> | null,
-  fetchShape: ReadShape<S, Params, Body>,
-) {
-  return !(
-    (fetchShape.options && fetchShape.options.invalidIfStale) ||
-    !resource
-  );
-}
-
```
zacharyfmarion: useResourceNew? I'm confused

src/react-integration/hooks/useResource.ts
ntucker: basically i need to force typescript to say resource is not null. but this is only in the case params is not null. anyway, this is least verbose way to do it because i can just put the actual return type in the function def.

src/react-integration/hooks/useResourceNew.ts
ntucker: Did you read the rest hooks update doc? also: https://github.com/coinbase/rest-hooks/commit/53c1922f2aa6fcc9b9dbf0889aebf78516581a85

Issue #118, opened 2019-08-05
pk() type should be string | number | undefined (not null)


ntucker:
One of the main purposes of `SimpleResource.pk()` is to pass as a key in React. However, React types keys as possibly being undefined - not null so the type should reflect that here.

PR #151, opened 2019-10-01
Changed PK type to include 'undefined'

ChristianRoper:
Fixes # .


ntucker:
Thanks for the PR. #118 actually refers to https://github.com/coinbase/rest-hooks/blob/master/src/resource/SimpleResource.ts#L19 though.

Be sure to fill in the PR description and update the docs to complete the PR!

ChristianRoper:
Ah, oops. Will do, thanks for the tip!

PR #152, opened 2019-10-01
Changed pk() types, replace 'null' with 'undefined'

### LINKED ISSUE #118: pk() type should be string | number | undefined (not null)

ChristianRoper:
Fixes #118  .

### Motivation

I believe this is exactly what was asked for. I updated the docs as well and think everything is in order.

### Solution

"One of the main purposes of SimpleResource.pk() is to pass as a key in React. However, React types keys as possibly being undefined - not null so the type should reflect that here."

### DIFF
--- a/docs/api/Resource.md
+++ b/docs/api/Resource.md
@@ -96,7 +96,7 @@ the instance in construction.
 
 ## Be sure to always provide:
 
-### pk: () => string | number | null
+### pk: () => string | number | undefined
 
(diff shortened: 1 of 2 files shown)


ntucker:
There's also a bunch of example pk() definitions in both the docs and tests that use `| null` now

PR #154, opened 2019-10-09
[enhance] Improve normalizr type expressiveness

ntucker:
### Solution

- Fake nominal typing for schema types (since they have pretty similar signatures).
  - future: should change to tracking return value of normalize as that can encode proper typings. We can use the schema provided and build on that to pretty convincingly manifest the algo. This can replace the giant ResultType and Denormalized block.
- Add missing readonly member schema
- Include Array in Schema and SchemaList def
- Add greedy resolution to Entity in ResultType and Denormalized generics.
  - This means if we only have SchemaDetail or SchemaList it will resolve to T and T[] respectively. While we don't actually know this - it enables a broad base class definition for our REST shapes. In the future a better solution is to simply make those shapes a mixin and encourage more often writing own shapes from scratch.

### Open questions

Should we just try to fix the normalize/denormalize algo now?

### DIFF
--- a/src/resource/normal.d.ts
+++ b/src/resource/normal.d.ts
@@ -11,7 +11,7 @@ declare namespace schemas {
     parent: any,
     key: string,
   ) => S;
-  type EntityMap<T = any> = { [key: string]: Entity<T> };
+  export type EntityMap<T = any> = { [key: string]: Entity<T> };
 
   export class Array<S extends Schema = Schema<any>> {
     private _identifier: 'Array';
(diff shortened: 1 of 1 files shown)


src/resource/normal.d.ts
```diff
@@ -13,25 +13,29 @@ declare namespace schemas {
   ) => S;
   type EntityMap<T = any> = { [key: string]: Entity<T> };
 
-  export class Array<T = any> {
-    constructor(
-      definition: Schema<T>,
-      schemaAttribute?: string | SchemaFunction,
-    );
+  export class Array<S extends Schema = Schema<any>> {
+    private _identifier: 'Array';
```
ntucker: faking nominal typing.

PR #171, opened 2019-11-04
[breaking] Drop IE support (out of the box)

ntucker:
### Motivation

- IE is becoming less and less common
- Noone uses it out of the box as a developer
- Can still be built to target it by building this package inline
- This has a pretty significant reduction in shipping codesize.

### Solution

Set browserslist that ignores IE as well as using @rest-hooks/normalizr version that does as well

### DIFF
--- a/src/.babelrc
+++ b/src/.babelrc
@@ -3,7 +3,8 @@
     [
       "@anansi/babel-preset",
       {
-        "typing": "typescript"
+        "typing": "typescript",
+        "legacyDecorators": true
       }
     ]
   ]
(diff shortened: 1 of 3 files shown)


src/.babelrc
```diff
@@ -3,7 +3,8 @@
     [
       "@anansi/babel-preset",
       {
-        "typing": "typescript"
+        "typing": "typescript",
+        "legacyDecorators": true
```
zacharyfmarion: what is this

ntucker: better name in latest anansi preset

PR #182, opened 2019-11-19
enhance: Move testing modules to own package

ntucker:
BREAKING CHANGE: rest-hooks/test -> @rest-hooks/test

### Motivation

Having test package be part of main package is incredibly awkward. This also allows us to publish extremely optional extensions like legacy mode easily.

### Solution

Lerna + yarn workspaces

`@rest-hooks/test` will be compatible with 3 and 4, so people can start using it immediately.

The most interesting piece was making the sort of circular dependency of rest-hooks and test work out both locally and with published versions.

#### Changelog generation

Moving from auto-changelog and release-it to conventional commits integrated into lerna's only publish powers.

#### TypeScript
During development the tsconfig for test remaps rest-hooks to point to the src dir. However, when compiled these path mappings are gone (tsconfig.compile.json). This makes it so in the package form it will correctly reference rest-hooks and pick up the peerdep.

#### Jest
Jest config is only used during development so that mapping was consistently set. Configured per-package. (only rest-hooks for now)

#### Eslint
Configured at top level with adjusted patterns to pick things up correctly

#### Build
Configured per-package. Fairly straightforward split

#### Coverage
Will need separate env variables for each coverage repo token, but for now only rest-hooks is there.

### Open questions

Currently building @rest-hooks/test and jest are a little jank as they both require basic lib build of rest-hooks first. Not sure how to best fix this but this works for now so we can wait to figure this out later.


.commitlintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
ntucker: started using this in anansi recently

.eslintrc
```diff
@@ -24,7 +24,8 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/explicit-member-accessibility": "off",
     "@typescript-eslint/no-use-before-define": "off",
-    "@typescript-eslint/unbound-method": "warn"
+    "@typescript-eslint/unbound-method": "warn",
+    "@typescript-eslint/no-unnecessary-type-assertion": ["error", { "typesToIgnore": ["any"] }]
```
ntucker: this was removing 'as any' that was actually needed

lerna.json
```diff
@@ -0,0 +1,14 @@
+{
+  "lerna": "0.0.1",
+  "npmClient": "yarn",
+  "useWorkspaces": true,
+  "version": "independent",
+  "command": {
```
ntucker: since yarn does the hard work, publishing becomes the most interesting part of lerna

packages/rest-hooks/README.md
```diff
@@ -0,0 +1,83 @@
+# ![🛌🎣 Rest hooks](./rest_hooks_logo_and_text.svg?sanitize=true)
```
ntucker: literally just moved from root

packages/rest-hooks/jest.config.js
```diff
@@ -7,17 +10,21 @@ module.exports = {
...
     '^.+\\.tsx?$': 'ts-jest',
     '^.+\\.jsx?$': 'babel-jest',
   },
   testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.tsx?$',
-  coveragePathIgnorePatterns: ['test', 'react-integration/hooks/useSelection'],
+  coveragePathIgnorePatterns: ['react-integration/hooks/useSelection'],
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
   moduleNameMapper: {
-    '^~/(.*)$': '<rootDir>/src/$1',
+    ...pathsToModuleNameMapper(compilerOptions.paths, {
```
ntucker: this DRY's up the concept of ~/ to only need to be in the tsconfig and then it's copied here.

packages/rest-hooks/rollup.config.js
```diff
@@ -60,19 +60,4 @@ export default [
       commonjs({ extensions }),
     ],
   },
-  // test utils commonjs build
```
ntucker: this part is just moved to the test package script - nothing changed

packages/rest-hooks/src/react-integration/__tests__/useCache.tsx
```diff
@@ -4,8 +4,7 @@ import {
   PaginatedArticleResource,
 } from '../../__tests__/common';
 import { useCache } from '../hooks';
-import makeRenderRestHook from '../../test/makeRenderRestHook';
-import { makeCacheProvider } from '../../test/providers';
+import { makeRenderRestHook, makeCacheProvider } from '../../../../test/src';
```
ntucker: haha so many more layers now

tsconfig.json
```diff
@@ -12,8 +12,8 @@
     // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
     // "sourceMap": true,                     /* Generates corresponding '.map' file. */
     // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    "outDir": "lib",                         /* Redirect output structure to the directory. */
-    "rootDir": "./src",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
+    //"outDir": "lib",                         /* Redirect output structure to the directory. */
```
ntucker: moving all the path stuff to each package, but want to keep general config to extend from here

README.md
```diff
@@ -61,7 +61,7 @@ For the small price of 8kb gziped. &nbsp;&nbsp; [🏁Get started now](https://re
 
 ## Features
 
-- [x] ![TS](./scripts/typescript.svg?sanitize=true) Strong [Typescript](https://www.typescriptlang.org/) types
+- [x] ![TS](./packages/rest-hooks/typescript.svg?sanitize=true) Strong [Typescript](https://www.typescriptlang.org/) types
```
zacharyfmarion: did you remove the scripts folder?

packages/rest-hooks/src/react-integration/__tests__/useCache.tsx
zacharyfmarion: yeah would be cool if you could reference the test package without a relative import

README.md
ntucker: just has testSetup now. I had originally put the svg in there to try to make there not be too many top-level files, but now that it's in packages, it's better served there

packages/rest-hooks/src/react-integration/__tests__/useCache.tsx
ntucker: i suppose i could add another path alias, but i'd rather try to keep that as clean as possible for now

PR #187, opened 2019-11-28
feat: New @rest-hooks/legacy package

### LINKED ISSUE #180: Move useStatefulResource into the library itself
The useStatefulResource hook is not that big, can probably be tree-shaken, and feels like it belongs in rest-hooks itself (esp now that it relies on internals). The reason I'm asking this is that the types seem to break periodically, and importing my own implementation feels kind of weird.

(Don't worry: I'm actually using this for stuff that doesn't make too much sense with suspense, like showing loading states in tables)

Make useStatefulResource an export from 'rest-hooks'

ntucker:
Fixes #180.

### Motivation

Internal implementation can change and it's best to let people easily upgrade.

### Solution

Having a separate package to keep this an optional include.


website/versioned_docs/version-3.0/guides/no-suspense.md
```diff
@@ -10,49 +10,18 @@ In these cases you might want a boolean `loading` instead of yielding to Suspens
 In any case, here's a sample hook you can adapt to use that information in any
```
ntucker: As with @rest-hooks/test - this library works with 3.0 as well so let's update the recommendation

packages/legacy/jest.config.js
```diff
@@ -0,0 +1,54 @@
...
+
+const { options } = readTsConfig();
+
+module.exports = {
+  roots: ['<rootDir>/src'],
+  /**
+   *  If you comment this out, you will get error unexpected token with optional chaining because you are using babel in your project
+   *  When using babel together with ts-jest in a project, you need to let ts-jest know about it
+   */
+  globals: {
```
zacharyfmarion: can you put this comment in the other place you do this

packages/legacy/src/__tests__/useStatefulResource.tsx
```diff
@@ -0,0 +1,86 @@
...
+    expect(result.current.data).toBe(undefined);
+    expect(result.current.error).toBe(undefined);
+    expect(result.current.loading).toBe(true);
+    await waitForNextUpdate();
+    expect(result.current.loading).toBe(false);
+    expect(result.current.error).toBeUndefined();
+    expect(result.current.data).toEqual(CoolerArticleResource.fromJS(payload));
+  });
+
+  /*it('should return errors on bad network', async () => {
```
zacharyfmarion: add comment

packages/legacy/jest.config.js
ntucker: already in the other jest config

PR #188, opened 2019-12-01
internal: Use TypeScript project references

ntucker:
This simplifies and makes more robust monorepo configuration

### Motivation

Remapping based on /src was weird. Now paths are only needed for IDE integration, and should be ignored at build.

### Solution

https://www.typescriptlang.org/docs/handbook/project-references.html improve typescript speed, while eliminating the need for weird src remapping. TypeScript source maps allow editors to reference the .d.ts files while still going to the ./src when looking up in IDEs.

- lint: use wildcard tsconfig to pick up each directory's config
- circle: build from root and let tsc orchestrate build order for types via `--build` option
- publishing: run `ttsc --build` before lerna publish to build types first
  - OPEN QUESTION: perhaps entire build should be done first in publish script and no build done in each dir?
- add src to published dirs: `--build` option adds sourcemaps that are used to return actual source from ambient declaration files. this will improve user EXP
- `build:clean` needs to delete .tsbuildinfo  files else types won't be regen.

#### Footnote

Getting this right will help make the migration of coinbase/www to TypeScript smooth

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,7 @@ jobs:
 
       # run tests!
       - run: yarn run lint packages/*/src
-      - run:
-          command: |
-            cd packages/rest-hooks
-            yarn run build
-      - run:
-          command: |
-            cd packages/test
-            yarn run build
+      - run: yarn run build
       - run: yarn run test:ci -- -- --maxWorkers=2
 
       # only run coverage if repo token is available (so third parties don't fail the build)
--- a/packages/rest-hooks/tsconfig.json
+++ b/packages/rest-hooks/tsconfig.json
@@ -1,12 +1,12 @@
 {
-  "extends": "../../tsconfig",
+  "extends": "../../tsconfig-base",
   "compilerOptions": {
     "outDir": "lib",
-    "rootDir": "./src",
-    "baseUrl": "./src",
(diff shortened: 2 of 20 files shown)


.circleci/config.yml
```diff
@@ -29,14 +29,7 @@ jobs:
...
       - run: yarn run lint packages/*/src
-      - run:
-          command: |
-            cd packages/rest-hooks
-            yarn run build
-      - run:
-          command: |
-            cd packages/test
-            yarn run build
+      - run: yarn run build
```
ntucker: glorious orchestration by `ttsc --build`

packages/rest-hooks/tsconfig.json
```diff
@@ -1,12 +1,12 @@
...
-  "extends": "../../tsconfig",
+  "extends": "../../tsconfig-base",
   "compilerOptions": {
     "outDir": "lib",
-    "rootDir": "./src",
-    "baseUrl": "./src",
+    "rootDir": "src",
     "paths": {
-      "~/*": ["*"]
+      "~/*": ["packages/rest-hooks/src/*"],
```
ntucker: can't use @rest-hooks/* for now until @rest-hooks/normalizr is brought in

PR #189, opened 2019-12-02
internal: Centralize babel config & common test

ntucker:
### Motivation

This is the right way to go for monorepo. A lot of things need to be done at top level, so when projects references each other it's easier if they're already accounting for non-project-relative configs.

### Solution

- Use babel.config.js
- Pull __tests__/common up to repo root


tsconfig-base.json
```diff
@@ -51,8 +51,10 @@
     "rootDir": ".",
     "outDir": "lib",
     "paths": {
-      "@rest-hooks/*": ["packages/*/src"],
-      "rest-hooks": ["packages/rest-hooks/src"]
+      "~/*": ["packages/rest-hooks/src/*"],
+      //"@rest-hooks/*": ["packages/*/src"],
```
zacharyfmarion: Can you add a comment

PR #199, opened 2019-12-15
enhance: Resource.fetch uses fetch instead of superagent

ntucker:
BREAKING CHANGE:
* fetchPlugin -> fetchOptionsPlugin, which has different signature
* No more SuperagentResource export
* New overridable Resource.resolveFetchData()

### Motivation

- People mostly know fetch these days
- Smaller bundle out of the box
- SSR for fetch has a decent implementation now with https://github.com/github/fetch
- Fetch has actually been improved greatly over the years offering interesting capabilities like streaming and cancellation.
- JSONP is probably not needed as it's been a long time since CORS was available.

### Solution

- [x] Add polyfill IE instructions to getting started installation guide
- [x] Add instructions for making SSR work that's linked to in testing guides and setup
- [x] Update existing solutions that are fetch-specific to include both fetch and superagent examples
  - [x] auth
  - [x] pagination
- [x] Update custom networking library to show default implementation and update the superagent one

`Resource.resolveFetchData()` provides a new customization hook that is much less verbose. Its objective is to return fetch responses into the correctly resolved data type. This opens it up for easy extensions like using headers, different data types, and streaming support.


packages/legacy/src/__tests__/useStatefulResource.tsx
```diff
@@ -18,32 +18,40 @@ afterAll(() => {
 describe('useStatefulResource()', () => {
   let renderRestHook: ReturnType<typeof makeRenderRestHook>;
 
-  beforeEach(() => {
-    renderRestHook = makeRenderRestHook(makeCacheProvider);
-    nock('http://test.com')
+  beforeAll(() => {
```
ntucker: a lot of these tests had to be updated to respond to CORS requests

packages/rest-hooks/src/react-integration/__tests__/subscriptions.tsx
```diff
@@ -33,13 +35,28 @@ for (const makeProvider of [makeCacheProvider, makeExternalCacheProvider]) {
     });
 
     beforeEach(() => {
-      nock('http://test.com')
+      nock(/.*/)
```
ntucker: i'm not sure why 'test.com' string doesn't work anymore

packages/rest-hooks/src/resource/__tests__/resource.ts
```diff
@@ -312,17 +328,65 @@ describe('Resource', () => {
...
+        );
+      } catch (e) {
+        error = e;
+      }
+      expect(error).toBeDefined();
+      expect(error.status).toBe(400);
     });
 
-    it('should use fetchPlugin if defined', async () => {
+    it('should return raw response if status is 204 No Content', async () => {
```
ntucker: added extra tests for all the special cases we were already handling before.

packages/rest-hooks/src/resource/Resource.ts
```diff
@@ -20,21 +28,38 @@ export default abstract class Resource extends SimpleResource {
...
-}
 
-export const isInvalidResponse = (res: request.Response): boolean => {
-  // Empty is only valid when no response is expect (204)
-  const resEmptyIsExpected = res.text === '' && res.status === 204;
-  const resBodyEmpty = Object.keys(res.body).length === 0;
-  return !(res.type.includes('json') || resEmptyIsExpected) && resBodyEmpty;
-};
+  /** Resolve response into correct data type */
+  static resolveFetchData?(response: Response) {
```
ntucker: new overridable 'hole'!

docs/guides/pagination.md
```diff
@@ -105,6 +103,32 @@ We show the custom listShape() below. All other parts of the above example remai
...
+  ReadShape,
+  SchemaList,
+  AbstractInstanceType,
+} from 'rest-hooks';
+
+export default class ArticleResource extends Resource {
+  // same as above....
+
+  /** Resolve response into correct data type */
+  static async resolveFetchData(response: Response) {
```
ntucker: so sexy this is way cleaner!

PR #203, opened 2019-12-18
feat: export SimpleRecord which has the data methods of Resource

ntucker:
### Motivation

Resource needs to be split up anyway, and I realized that this is one piece that already works.

### Solution

Pull data methods into SimpleRecord that SimpleResource extends.

### Open questions

Naming....I don't really like SimpleRecord, but not sure what else to call it. I don't want to call it Record since that's a typescript generic util type.

### DIFF
--- a/packages/rest-hooks/src/resource/SimpleRecord.ts
+++ b/packages/rest-hooks/src/resource/SimpleRecord.ts
@@ -1,0 +1,82 @@
+import { AbstractInstanceType } from '~/types';
+
+const DefinedMembersKey = Symbol('Defined Members');
+type Filter<T, U> = T extends U ? T : never;
+interface SimpleResourceMembers<T extends typeof SimpleRecord> {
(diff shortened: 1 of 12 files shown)


packages/rest-hooks/src/resource/SimpleRecord.ts
```diff
@@ -0,0 +1,82 @@
+import { AbstractInstanceType } from '~/types';
+
+const DefinedMembersKey = Symbol('Defined Members');
+type Filter<T, U> = T extends U ? T : never;
+interface SimpleResourceMembers<T extends typeof SimpleRecord> {
+  [DefinedMembersKey]: Filter<keyof AbstractInstanceType<T>, string>[];
+}
+
+/** Immutable record that keeps track of which members are defined vs defaults. */
+export default abstract class SimpleRecord {
```
ntucker: this is all copy/pasted from SimpleResource, except the ABC check was moved out (since it used `pk` checking which isn't valid for this class).

PR #226, opened 2020-01-04
feat: Entity class added to Resource hierarchy

ntucker:
### Motivation

This makes it easier to create entities that are not necessarily used by Resources

- Useful for nested entities that don't have endpoints like LatestPrice
- Useful for non-REST style APIs like CSF and GraphQL. This makes building their FetchShapes much easier.
- This also makes it easier to integrate full denormalization

### Solution

New hierarchy:

```
 SimpleRecord
      |
   Entity
      |
SimpleResource
      |
   Resource
```

Deprecations:

- Resource.getKey() -> Resource.key
- Resource.getEntitySchema() -> Resource.asSchema()
- Entity.define() -> override Entity.schema

Changes:
* Normalizr: top level key sent to getId is undefined not null
* pk() now takes additional parent, and key optional args
* pk() no longer accepts number return value (run .toString())


docs/api/Resource.md
```diff
@@ -198,11 +207,21 @@ on network error. This can be useful to override to really customize your transp
...
 Used in `fetch()`. Resolves the HTTP [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response).
 
-### static getEntitySchema() => [schema.Entity](https://github.com/ntucker/normalizr/blob/master/docs/api.md#entitykey-definition---options--)
+### static asSchema() => [Entity](./Entity)
 
-Returns the [shape of the data](https://github.com/ntucker/normalizr/blob/master/docs/api.md#schema)
-when requesting one resource at a time. Defaults to a plain object
-containing the keys. This can be useful to override if your response is in a different form.
+Returns this Resource as an [Entity](./Entity) with the TypeScript type set properly. Using
+this method instead of `this` directly key to getting correct typing from the hooks.
```
kmartinezmedia: Might want to make more like `asSchema` description. Maybe something like below

```
Returns this Resource as an `Entity` with the TypeScript type set properly. Using `asSchema()` instead of
`this` directly is key to getting correct typing from the hooks.
```

docs/api/Entity.md
```diff
@@ -0,0 +1,154 @@
...
+#### undefined value
+
+A `undefined` can be used as a default to indicate the entity has not been created yet.
+This is useful when initializing a creation form using [Entity.fromJS()](#static-fromjst-extends-typeof-simplerecordthis-t-props-partialabstractinstancetypet-abstractinstancetypet)
+directly. If `pk()` resolves to null it is considered not persisted to the server,
+and thus will not be kept in the cache.
+
+#### Other uses
+
+While the `pk()` definition is key (pun intended) for making the normalized cache work;
```
outfitted: lol

PR #237, opened 2020-01-16
feat: Add indexes to Entity for improved performance

### LINKED ISSUE #232: Improve performance with arbitrary indexes

Specify arbitrary indexes in entities.

Define indexes like so. The function param would already be typed as `AbstractIntanceType<T>`

```typescript
class AssetInfoResource extends Resource {
  static indexes = [
    [this.detailShape(), value => ({ slug: value.slug })],
    [this.symbolShape(), value => ({ symbol: value.symbol })],
  ];
}
```

Any `receive` action would use this to update the reducer state like so:

```typescript
// action.payload is the Entity itself
{
  entities: {
    [AssetInfoResource.key]: {
      [AssetInfoResource.pk(payload)]: AssetInfoResource.fromJS(payload),
    }
  },
  results: {
    [indexes[0][0].getFetchKey(indexes[0][1](payload))]: buildInferredResults(index[0][0].schema, action.payload),
  },
}
```

No meta information would be set for that result entry. This means the data is always considered 'stale'. Thus, in normal uses of useResource() it will use the Stale-While-Revalidate policy.

**Additional context**
[We already index on pk by this mechanism](https://github.com/coinbase/rest-hooks/blob/master/packages/rest-hooks/src/state/selectors/buildInferredResults.ts)

ntucker:
Fixes #232.


kmartinezmedia:
🚀

PR #251, opened 2020-01-27
enhance: Keep referential equality in list views

### LINKED ISSUE #249: listShape lacks reference equality across rerenders
**React version** 16.12.0

**Concurrent mode** no

**Describe the bug**
The return value of useResource(MyResource.listShape(), {}) is referentially different on every rerender, with some of the returned MyResource instances also being referentially different. This makes it difficult to optimize things.

**To Reproduce**
1. Make a resource
2. `const myResourceList = useResource(MyResource.listShape(), {});`
3. `useEffect(() => console.log(myResourceList), [myResourceList]);`
4. Trigger a rerender on that component (setting state, etc)
5. Observe `myResourceList` being logged 

**Expected behavior**
`myResourceList` should be referentially stable across renders if it is being pulled from cache.

**Additional context**
Any console errors? What does the networking inspector show?

ntucker:
Fixes #249 .

### Motivation

- Performance
- Hooks world expects this - like useEffect, etc

### Solution

- Track all entities captured in denormalize
- Serialize them
- Use that serialization as memo check for results

### Open questions

How does this affect performance overall? Could this be degrading non-list cases?

### DIFF
--- a/packages/normalizr/src/index.js
+++ b/packages/normalizr/src/index.js
@@ -155,41 +155,44 @@ const getUnvisit = entities => {
   const cache = {};
   const getEntity = getEntities(entities);
 
-  return function unvisit(input, schema) {
-    if (!schema) return [input, true];
+  return [
+    function unvisit(input, schema) {
+      if (!schema) return [input, true];
 
-    if (
-      typeof schema === 'object' &&
-      (!schema.denormalize || typeof schema.denormalize !== 'function')
-    ) {
-      const method = Array.isArray(schema)
-        ? ArrayUtils.denormalize
-        : ObjectUtils.denormalize;
-      return method(schema, input, unvisit);
-    }
-
-    // null is considered intentional, thus always 'found' as true
-    if (input === null) {
-      return [input, true];
-    }
-
-    if (
-      typeof schema.getId === 'function' &&
-      typeof schema.normalize === 'function'
-    ) {
-      // unvisitEntity just can't handle undefined
-      if (input === undefined) {
-        return [input, false];
+      if (
+        typeof schema === 'object' &&
+        (!schema.denormalize || typeof schema.denormalize !== 'function')
+      ) {
+        const method = Array.isArray(schema)
+          ? ArrayUtils.denormalize
+          : ObjectUtils.denormalize;
+        return method(schema, input, unvisit);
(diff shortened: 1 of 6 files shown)


packages/normalizr/src/index.js
```diff
@@ -155,41 +155,44 @@ const getUnvisit = entities => {
   const cache = {};
   const getEntity = getEntities(entities);
 
-  return function unvisit(input, schema) {
```
ntucker: this whole block until denormalizeAndTracked  is actually just returning `[previousReturn, cache]`

Formatting changed each line due to the tabbing over.

packages/rest-hooks/src/resource/SimpleRecord.ts
```diff
@@ -8,6 +8,13 @@ interface SimpleResourceMembers<T extends typeof SimpleRecord> {
 
 /** Immutable record that keeps track of which members are defined vs defaults. */
 export default abstract class SimpleRecord {
+  // a 'unique' identifier to make referential equality comparisons easy
+  declare readonly _unq: string;
```
ntucker: this is sort of a hack for react not wanting variable sync comparisons. we'll be serializing this later

packages/rest-hooks/src/state/selectors/useDenormalized.ts
```diff
@@ -63,13 +63,31 @@ export default function useDenormalized<
...
       schema,
       entities,
     );
-    return [denormalized, entitiesFound] as any;
+
+    // this enables us to keep referential equality based on entities contained within
+    const entitiesList = Object.values(cache)
+      .map(Object.values)
+      .reduce((a: any[], b: any[]) => a.concat(b), [])
+      .join(',');
```
ntucker: serializing this list so react doesn't complain. this is somewhat dangerous as people not using `Entity` will have troubles.

kmartinezmedia:
Looks good to me. Would like to know if we can answer, `How does this affect performance overall? Could this be degrading non-list cases?` or if this is negligible.

ntucker:
I could create some synthetic benchmarks to run. For non-list cases since the entity list will be incredibly small, while it will have some non-zero impact, it probably won't be large

PR #254, opened 2020-01-28
feat: Ability to dynamically set nested schema type

ntucker:
### Motivation

https://github.com/paularmstrong/normalizr/pull/415

Ability to 'normalize entity nested inside object using property from parent'

### Solution

In Object (or plain) the members can be functions that resolve to schemas based on the input.


ntucker:
Not sure this feature is worth the complexity.

PR #267, opened 2020-02-13
feat: Support non JSON (aka binary) fetches

### LINKED ISSUE #266: Fetching ArrayBuffer via Resource
This is more of a question than necessarily a bug.

I stumbled across fetching a photo for a specific user which is send as an `ArrayBuffer`.
The endpoint to request the photo is: `/users/{userId}/photo`.

I created a separate `UserPhotoResource` additional to my `UserResource` due to problems with missing keys that are not present on the photo return etc.
The `UserPhotoResource` looks like this:

```typescript
import { Resource } from 'rest-hooks';

export default class UserPhotoResource extends Resource {
  pk() {
    return 'user_photo';
  }

  static url<T extends typeof Resource>(
    this: T,
    urlParams: { userId: string },
  ): string {
    if (urlParams) {
      return `/users/${urlParams.userId}/photo`;
    }

    throw new Error('UserPhoto requires userId to retrieve');
  }

  static photoShape<T extends typeof Resource>(this: T) {
    return {
      ...this.detailShape(),
      getFetchKey: ({
        userId,
      }: {
        userId: string;
      }) => {
        return `/users/${userId}/photo`;
      },
      fetch: async ({
        userId,
      }: {
        userId: string;
      }) => {
        const response = await fetch(
          `/users/${userId}/photo`,
        );
        const photoArrayBuffer = await response.arrayBuffer();

        return photoArrayBuffer;
      },
    };
  }
}
```

If I call it like this:
```typescript
const userPhoto = useResource(UserPhotoResource.photoShape(), {
  userId,
});

console.log(userPhoto);
```
It logs:
```
UserPhotoResource {_unq: "0.608616043130813", __ownerID: 1337, Symbol(Defined Members): Array(0)}
```

It would be super helpful if you could lead me in the right direction how to fetch a `ArrayBuffer` resource or even explain how to integrate the `/photo` endpoint in the `UserResource` so i don't need a additional `UserPhotoResource`

ntucker:
Fixes #266.

### Motivation

With Rest Hooks already useful for JSON type data, it should allow for other response types like images, ArrayBuffers, etc.

### Solution

Sample FetchShape:
```typescript
export const photoShape = {
  type: 'read' as const,
  schema: null as ArrayBuffer | null,
  getFetchKey({ userId }: { userId: string }) {
    return `/users/${userId}/photo`;
  },
  fetch: async ({ userId }: { userId: string }) => {
    const response = await fetch(`/users/${userId}/photo`);
    const photoArrayBuffer = await response.arrayBuffer();

    return photoArrayBuffer;
  },
};
```

Usage:
```tsx
const photo = useResource(photoShape, { userId });
```

### Open questions

useResource() ensures a response, but we're still stuck with the | null union. This is needed however to have sane defaults. Currently the only distinction is based on entities, which this cannot be so would require some rethinking from normalizr design.

Perhaps a better solution is to simply allow schema to be undefined, and then take the return type of fetch.

### DIFF
--- a/__tests__/common.ts
+++ b/__tests__/common.ts
@@ -296,6 +296,20 @@ export class NestedArticleResource extends OtherArticleResource {
   };
 }
 
+export const photoShape = {
+  type: 'read' as const,
+  schema: null as ArrayBuffer | null,
+  getFetchKey({ userId }: { userId: string }) {
+    return `/users/${userId}/photo`;
+  },
+  fetch: async ({ userId }: { userId: string }) => {
+    const response = await fetch(`http://test.com/users/${userId}/photo`);
+    const photoArrayBuffer = await response.arrayBuffer();
+
+    return photoArrayBuffer;
+  },
+};
+
 export function makeErrorBoundary(cb: (error: any) => void) {
   return class ErrorInterceptor extends React.Component<any, { error: any }> {
     state = { error: null };
--- a/docs/guides/binary-fetches.md
+++ b/docs/guides/binary-fetches.md
@@ -1,0 +1,41 @@
+---
+title: Fetching Media
+---
+
+After setting up Rest Hooks for structured data fetching, you might want to incorporate
+some media fetches as well to take advantage of suspense and concurrent mode support.
+
+[Resource](../api/Resource) and [Entity](../api/Entity) should not be used in this case, since they both represent
+string -> value map structures. Instead, we'll define our own simple [FetchShape](../api/FetchShape)
+with a schema set to null, but with a type including what we expect in the response.
+
+Schemas with literal types like null simply pass through the response, but their value is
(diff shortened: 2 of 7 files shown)


__tests__/common.ts
```diff
@@ -296,6 +296,20 @@ export class NestedArticleResource extends OtherArticleResource {
...
+  getFetchKey({ userId }: { userId: string }) {
+    return `/users/${userId}/photo`;
+  },
+  fetch: async ({ userId }: { userId: string }) => {
+    const response = await fetch(`http://test.com/users/${userId}/photo`);
+    const photoArrayBuffer = await response.arrayBuffer();
+
+    return photoArrayBuffer;
+  },
+};
```
kmartinezmedia: Nice! This is a great feature!

PR #274, opened 2020-02-24
fix: type SimpleRecord._unq should be private

ntucker:
### Motivation

SimpleRecord._unq is a tracker used in toString() to make referential comparisoins easier (#251 ).

Exposing this is problematic as it pollutes the structural definition of any classes inheriting from SimpleRecord. This might have been done previously due to limitations of parsing declare lines - but this seems to work now.

### Solution

Removing the type declaration altogether is necessary to avoid its pollution of type compatibility

https://www.typescriptlang.org/docs/handbook/type-compatibility.html#private-and-protected-members-in-classes

Technically this is a breaking change, but it increases compatibility rather than restricts it, so it likely won't cause issues.

### DIFF
--- a/packages/rest-hooks/src/react-integration/__tests__/hooks.tsx
+++ b/packages/rest-hooks/src/react-integration/__tests__/hooks.tsx
@@ -1,13 +1,14 @@
-import React, { Suspense, useEffect } from 'react';
-import { render } from '@testing-library/react';
-import { renderHook } from '@testing-library/react-hooks';
-import nock from 'nock';
 import {
   CoolerArticleResource,
   PaginatedArticleResource,
   ArticleResourceWithOtherListUrl,
   StaticArticleResource,
 } from '__tests__/common';
+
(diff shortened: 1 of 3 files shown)


packages/rest-hooks/src/resource/SimpleRecord.ts
```diff
@@ -1,15 +1,15 @@
-import { AbstractInstanceType } from 'rest-hooks/types';
+import { EntityInstance } from 'rest-hooks/types';
```
ntucker: this is the same thing, but wrapped in Readonly<>. Arguments are never modified in these methods, and this allows us to not pay attention to private members.

packages/rest-hooks/src/react-integration/__tests__/hooks.tsx
```diff
@@ -1,14 +1,15 @@
-import React, { Suspense, useEffect } from 'react';
```
ntucker: this is just import order

PR #275, opened 2020-02-24
feat: Export __INTERNAL__.actionTypes

ntucker:
### Motivation

Building custom Managers

### Solution

Exported under \_\_INTERNAL__ because action structure is still considered somewhat unstable.


ntucker:
Will export these from @rest-hooks/core when it's released

PR #281, opened 2020-03-03
fix: Add generic useResource parallel overloads for up to 16 arguments

### LINKED ISSUE #280: Parallel useResource should support more than 5 args
**React version** (e.g., 16.8.5) 16.12

**Concurrent mode** no

**Describe the bug**
If useResource is called with >5 args in TS, it fails to compile. This is a bit of a dumb request, but I think this should be supported up to something higher (like 16).

yunyu:
Fixes #280.

### Motivation
This is probably enough parallel resource load overrides to satisfy anybody.

### Solution
Generated with https://gist.github.com/yunyu/230161d5ddfdb1b7f7ad368662b73d04


ntucker:
Nice! Maybe I should put this in a build script instead of having it hard coded.

yunyu:
@ntucker There's a [better way to do this now](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#variadic-tuple-types) with TS 4. I'm not sure about the backwards compatibility story, but would you be interested in potentially switching to it if it's not backwards compatible?

ntucker:
@yunyu Yes. Will have to wait to land til after v5 is released as I don't want to add that to the breaking changes

PR #291, opened 2020-03-12
fix: Don't throw when RN tries to poll

ntucker:
### Motivation

React Native doesn't have addEventListener. Furthermore, it has it's own way of detecting online/offline. 

### Solution

For now we'll just not activate offline/online detection when addEventListener isn't available.

### Open questions

- How can we cleanly add online/offline detection for RN
- We should figure out how to run the tests in a React Native environment so things like this never happen again.

### DIFF
--- a/packages/rest-hooks/src/state/PollingSubscription.ts
+++ b/packages/rest-hooks/src/state/PollingSubscription.ts
@@ -150,7 +150,9 @@ export default class PollingSubscription implements Subscription {
         }
         this.update();
       }, this.frequency);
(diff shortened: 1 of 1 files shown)


dgcoffman:
There are other uses of `addEventListener` in this code which haven't been guarded, and calls to `removeEventListener` which I suspect will also throw.

Maybe further guarding would be appropriate.

dgcoffman:
ok looks like isOnline() will always return true for React Native, so the other add/removeEventListeners won't be hit.

PR #302, opened 2020-03-21
fix: Do not run reducer on non-optimistic fetches

### LINKED ISSUE #300: Various warnings in React 16.13.1
**React version** 16.13.1

**Concurrent mode** no

**Describe the bug**
React throws warnings when useResource is used anywhere:

```
Warning: Cannot update a component (`CacheProvider`) while rendering a different component (`PartnerRewards`). To locate the bad setState() call inside `PartnerRewards`, follow the stack trace as described in https://fb.me/setstate-in-render
```

ntucker:
Fixes #300  .

### Motivation

Even though the state does not update, React will warn about the reducer running from render (triggered from useRetrieve()). Although React should short-circuit and not actually re-render, it doesn't harm for us to short-circuit even earlier - after all it will run less code!

### Solution

Intercept all fetch dispatches in the NetworkManager that cannot cause state updates. (No optimistic response)

### DIFF
--- a/packages/rest-hooks/src/react-integration/hooks/useError.ts
+++ b/packages/rest-hooks/src/react-integration/hooks/useError.ts
@@ -20,9 +20,7 @@ export default function useError<
     if (!meta.error) {
       // this means we probably deleted the entity found in this result
       const err: any = new Error(
-        `Resource not found in cache ${
-          params ? fetchShape.getFetchKey(params) : ''
-        }`,
+        `Resource not found in cache ${fetchShape.getFetchKey(params)}`,
       );
       err.status = 404;
       return err;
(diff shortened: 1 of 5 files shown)


packages/rest-hooks/src/react-integration/hooks/useError.ts
```diff
@@ -20,9 +20,7 @@ export default function useError<
     if (!meta.error) {
       // this means we probably deleted the entity found in this result
       const err: any = new Error(
-        `Resource not found in cache ${
-          params ? fetchShape.getFetchKey(params) : ''
```
ntucker: this was actually not needed - as params is already discriminated above.

PR #309, opened 2020-03-29
internal: Prepare tests to run in React Native env

ntucker:
### Motivation

To ensure the same quality level for React Native as web has experienced, we should run all tests in a react native environment.

### Solution

Jest has a preset for this. We can't do normal react render tests, but most are just hooks or function tests anyway.

Test with .web.ts will only be tested on web.
Test with .native-test.ts will only be tested on native.
Without the above, both web and native will run the test.

Currently this won't actually run in RN as their setup files are all borked by being hardcoded in flowtype. This just prepares for the day when that is fixed and it will be an easy mod here.

### Open questions

What about any oddities are fetch? Since we can't actually test react-native's implementation


jest.config.js
```diff
@@ -55,6 +55,29 @@ const projects = [
...
+    transformIgnorePatterns: [
+      '/node_modules/(?!(jest-)?react-native|@react-native-community)', //from RN preset
+      '<rootDir>/.*__tests__/[^/]+\\.web\\.(j|t)sx?$',
+    ],
+    setupFiles: [
+      '<rootDir>/node_modules/react-native/jest/setup.js', //from RN preset
+      '<rootDir>/scripts/testSetupNative.js',
+    ],
+    transform: {
+      //'^.+\\.js$': '<rootDir>/node_modules/react-native/jest/preprocessor.js',  setup.js needs to be transformed, but preprocessor screws everything else up
```
ntucker: https://github.com/facebook/react-native/issues/21075

PR #310, opened 2020-04-01
enhance: Infer returntype of useFetcher()'s functions

ntucker:
### Motivation

Previously it was just 'any', which is often not very helpful.

### Solution

We'll grab from the fetch function as that's the simplest approach and most accurate.

### DIFF
--- a/packages/rest-hooks/src/react-integration/hooks/useResource.ts
+++ b/packages/rest-hooks/src/react-integration/hooks/useResource.ts
@@ -1,4 +1,3 @@
-import { useMemo, useContext } from 'react';
 import {
   ReadShape,
   Denormalize,
@@ -7,6 +6,8 @@ import {
 } from 'rest-hooks/resource';
 import { useDenormalized } from 'rest-hooks/state/selectors';
 import { StateContext } from 'rest-hooks/react-integration/context';
+
+import { useMemo, useContext } from 'react';
 
 import useRetrieve from './useRetrieve';
 import useError from './useError';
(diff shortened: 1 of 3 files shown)


packages/rest-hooks/src/react-integration/hooks/useResource.ts
```diff
@@ -61,7 +62,7 @@ function useManyResources<A extends ResourceArgs<any, any>[]>(
     // only wait on promises without results
     .map(
       (p, i) =>
-        !hasUsableData(denormalizedValues[i][1], resourceList[i][0]) && p,
+        !hasUsableData(denormalizedValues[i][1], resourceList[i][0]) && !!p,
```
ntucker, line 65: this unified the types which made this work cleaner

packages/rest-hooks/src/react-integration/hooks/useRetrieve.ts
```diff
@@ -1,15 +1,15 @@
...
 
 import useFetcher from './useFetcher';
 import useExpiresAt from './useExpiresAt';
 
 /** Request a resource if it is not in cache. */
-export default function useRetrieve<
-  Params extends Readonly<object>,
-  S extends Schema
->(fetchShape: ReadShape<S, Params>, params: Params | null) {
+export default function useRetrieve<Shape extends ReadShape<any, any>>(
```
ntucker, line 9: parameter of Shape rather than a part gives us more information. this is used in useFetcher() which infers its return type

PR #319, opened 2020-04-24
enhance: useResource array parameters as readonly

skoging:
Makes it easier to pass tuples defined with "as const" assertion as paramters.

Fixes # .

### Motivation

```ts

export const compResources= [
  [SomeResource.listShape(), {}],
  [OtherResource.listShape(), {}]
] as const;

const Comp = () => {
  const [someData, otherData] = useResource(...compResources);
}

```

The arguments passed to `useResource()` will be `readonly [S, P]`, which errors due to the useResource array arguments being defined without readonly modifier.

### Solution

Add the `readonly` modifier to all array arguments of useResource: `useResource(readonly [S1, P1], ...)`.
There's no reason not to add this modifier, as these arrays should not be mutated inside this function anyway.


skoging:
That was quick! Thanks 😄

ntucker:
May want to have the same for all the hooks

ntucker:
Released in 5.0.0-beta.0 (0 is after beta.4 - accidentally missed it)

Issue #277, opened 2020-02-26
Entity.denormalize() is not yet written, so denormalizing a nested Resource is not possible


chrisapplegate:
**React version** (e.g., 16.8.5)

16.12

**Concurrent mode** yes/no

No

**Describe the bug**

Trying to take advantage of the nesting & normalization features detailed [here](https://resthooks.io/docs/guides/nested-response), however it seems that while data is normalized correctly, it is not denormalized.

**To Reproduce**

 if I have two resources set up as e.g. a `PostResource` and an `AuthorResource`:

```
export class AuthorResource extends Resource {
  id = null;
  name;
  pk() {
    return this.id?.toString();
  }
  static urlRoot = "/api/author";
}

export class PostResource extends Resource {
  id;
  author;
  title;
  pk() {
    return this.id?.toString();
  }
  static urlRoot = "/api/post";
  static schema = {
    author: AuthorResource.asSchema()
  }
}
```

And I then fetch `PostResource` using: `const posts = useResource(PostResource.listShape(), {})`, I note that in the Redux store I am using, the `PostResource` and `AuthorResource` objects are correctly normalized and stored under different stoage paths. But when retrieving the `PostResource`, the `AuthorResource` is not correctly denormalized and put back into the object. I will just get something like e.g.:

```
console.log(posts[0]);
>> {
  id: 1,
  author: "1000"
  ...
}
```

I thought I was doing something wrong until I looked at the source code and found in [Entity.ts](https://github.com/coinbase/rest-hooks/blob/master/packages/rest-hooks/src/resource/Entity.ts) a `TODO: Add denormalizing capability` comment on the `denormalize` static function.

**Expected behavior**

Expect the returned object to be something like:

```
{
  id: 1,
  author: {
    id: 1000,
    name: "John Doe"
  }
  ...
}
```

**Additional context**

Should I just not bother using normalizing or specifying a schema for nested objects until this is TODO fixed? If so can we get an update to the documentation saying denormalizing is not currently supported?

chrisapplegate:
Incidentally, I wrote a hacky shim to override `denormalize()` that seems to work:

```
class DenormalizableResource extends Resource {
  static denormalize(entity, unvisit) {
    Object.keys(this.schema).forEach((key) => {
      if (key in entity) {
        entity[key] = unvisit(entity[key], this.schema[key])[0];
      }
    });
    return [entity, true];
  }
}
```

[...]

Issue #165, opened 2019-10-23
Include nested entities in useCache() response


ntucker:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

ntucker:
#328

PR #328, opened 2020-04-29
feat: Entity fully denormalizes & Entity -> FlatEntity for non-nested

### LINKED ISSUE #301: [docs] Immutability
I noticed that there isn't much documentation, at least that I could find, that details how one might use immutability with a 3rd party library, i.e. immutable-js. Is there a way to intercept the data coming back from the hook or should I just create a wrapper around the hook calls?

ntucker:
Fixes #301, #277, #165.

### Motivation

Lacking full denormalization is a regression from recent consolidation. This allows for that.

### Solution

- Entity now fully supported nested denormalization - including recursion.
  - Use this as basis for tests now so they have similar results as before.
  - This does not guarantee referential equality every denorm call. This will need to be solved to eliminate FlatEntity. (this is the old schema.Entity behavior)
- FlatEntity replicated previous Entity behavior, which is what rest-hooks uses to build Resource
- Fixed up immutablejs usage to still result in Entity construction.
- All ids in normalizr tests are now strings - as that is the only allowed `pk` type.

To maintain backwards compatibility, exports from 'rest-hooks' package itself will be `NestedEntity` and `Entity` - that have previous behavior. New projects should import `Entity` and `FlatEntity` from `@rest-hooks/normalizr` directly.

### Open questions

The next step is to make referential equality last more than one denorm call.

Entity, and it's schema members will need to be the keys mapping into a fully denormalized object. These should be freed from memory as soon as all other references to the fully denormalized object are lost.
```
{
  this: this,
  draftedBy: USERA,
  publishedBy USERB,
} -> fullyDenorm(members)
```


packages/normalizr/src/entities/__tests__/Entity.test.ts
```diff
@@ -233,17 +233,16 @@ describe(`${Entity.name} normalization`, () => {
   });
 });
 
-/* TODO: This isn't supported quite yet
```
ntucker, line 236: note we re-enable this entire block. This has a lot of interested recursively nested denormalization cases.

packages/rest-hooks/src/index.ts
```diff
@@ -32,14 +26,19 @@ import useSelectionUnstable from './react-integration/hooks/useSelection';
 import hasUsableData from './react-integration/hooks/hasUsableData';
 import { StateContext, DispatchContext } from './react-integration/context';
 
+export {
+  Entity as NestedEntity,
```
ntucker, line 30: remapping so there's no actual breaking changes for 'rest-hooks' itself (for now)

PR #334, opened 2020-05-08
feat: Allow SimpleRecords to be schemas

ntucker:
### Motivation

Plain objects assume the value is there. SimpleRecords allow for defaults to be provided. There are times when we want to use these on things that aren't entities (with proto3).

### Solution

Add normalize/denormalize that just pulls on 'object' version.

```ts
const schema = {
  cards: [] as LoyaltyCard[],
  pageDescription: new PageDescription(),
  cardLocation: LoyaltyCardLocation.HOME_CAROUSEL,
}
```
to

```ts
class ListLoyaltyCards extends SimpleRecord {
  readonly cards: LoyaltyCard[] = [];
  readonly pageDescription = new PageDescription();
  readonly cardLocation = LoyaltyCardLocation.HOME_CAROUSEL;
}
const schema = ListLoyaltyCards.asSchema();
```

### DIFF
--- a/packages/normalizr/src/entities/Entity.ts
+++ b/packages/normalizr/src/entities/Entity.ts
@@ -1,7 +1,7 @@
-import SimpleRecord from './SimpleRecord';
+import SimpleRecord, { SimpleRecordSchema } from './SimpleRecord';
 import { isImmutable, denormalizeImmutable } from '../schemas/ImmutableUtils';
 import * as schema from '../schema';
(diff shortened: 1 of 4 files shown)


landakram:
Maybe we want to add a test for `SimpleRecord`? I see there's some for `Entity`: https://github.com/coinbase/rest-hooks/blob/master/packages/normalizr/typescript-tests/entity.ts#L19

PR #350, opened 2020-05-31
feat: SimpleRecord can have optional entities by setting default to null

ntucker:
### Motivation

A way to express nested members that aren't required.

### Solution

For both Entity, and SimpleRecord:

- for all members of `static schema`, if default is set, and that default is falsy - field is considered optional
- optional fields don't need to be 'found'
- fromJS() works with Records sent

### DIFF
--- a/packages/normalizr/src/entities/__tests__/__snapshots__/Entity.test.ts.snap
+++ b/packages/normalizr/src/entities/__tests__/__snapshots__/Entity.test.ts.snap
@@ -80,7 +80,7 @@ Array [
     },
     "id": "2",
   },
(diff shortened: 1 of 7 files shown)


packages/normalizr/src/entities/__tests__/__snapshots__/Entity.test.ts.snap
```diff
@@ -80,7 +80,7 @@ Array [
     },
     "id": "2",
   },
-  true,
+  false,
```
ntucker, line 83: this was actually a bug 😲

PR #352, opened 2020-06-03
fix: Normalizr errors reject and throw in useResource()

ntucker:
### Motivation

- Catching errors earlier is helpful when building and understanding network endpoints.
- Also pretty key that normalize gets what it needs, or denormalize won't be able to get the correct results.
- We actually need to ensure we don't throw in the reducer or the provider will unmount.

### Solution

We *must* handling the error in the reducer, or React gets confused about state updates (we don't want to redispatch in middle of a render). Therefore, we catch processing receives (where normalizr should be able to throw errors) and process those as network errors (setting the meta error). To ensure invariants with usage in functions, we check for this condition in NetworkManager so it can reject the promise.

### Open questions

We should probably get rid of other error handling routes in NetworkManager for simplicity - since they will all result in state updates.

We need to fix usage of `act` in tests, since we now rely on getState(). Note, this does work properly (to reflect batching behavior) - but `act()` is needed for tests to behave in the same manner.

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -39,6 +39,7 @@ export default class NetworkManager implements Manager {
 
     this.middleware = <R extends React.Reducer<any, any>>({
       dispatch,
+      getState,
     }: MiddlewareAPI<R>) => {
       return (next: Dispatch<R>) => (
         action: React.ReducerAction<R>,
@@ -58,7 +59,14 @@ export default class NetworkManager implements Manager {
             // only receive after new state is computed
             return next(action).then(() => {
               if (action.meta.key in this.fetched) {
-                this.handleReceive(action);
+                // Note: meta *must* be set by reducer so this should be safe
+                const error = getState().meta[action.meta.key]?.error;
+                // processing errors result in state meta having error, so we should reject the promise
+                if (error) {
+                  this.handleReceive(createReceiveError(error, action.meta));
+                } else {
+                  this.handleReceive(action);
+                }
               }
             });
           case RESET_TYPE:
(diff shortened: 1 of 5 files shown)


packages/core/src/state/reducer.ts
```diff
@@ -55,41 +60,37 @@ export default function reducer(
     }
     case RECEIVE_TYPE: {
       if (action.error) {
+        return reduceError(state, action, action.payload);
+      }
+      try {
```
ntucker, line 65: diff here is very weird (showed sensible things in vscode).

this is actually just putting everthing in try/catch

above we extracted the error handling case so we could reuse it when this throws

packages/core/src/state/actions/createReceiveError.ts
```diff
@@ -15,14 +16,14 @@ interface Options<S extends Schema = any>
 export default function createReceiveError<S extends Schema = any>(
   error: any,
   { schema, key, type, errorExpiryLength }: Options<S>,
-): ResponseActions {
+): typeof type extends 'delete' ? PurgeAction : ReceiveAction {
```
ntucker, line 19: just improving the specificity of return types here

packages/core/src/state/NetworkManager.ts
```diff
@@ -58,7 +59,14 @@ export default class NetworkManager implements Manager {
             // only receive after new state is computed
             return next(action).then(() => {
               if (action.meta.key in this.fetched) {
-                this.handleReceive(action);
+                // Note: meta *must* be set by reducer so this should be safe
+                const error = getState().meta[action.meta.key].error;
```
ntucker, line 63: if reducer turns this into an error, we should reject the promise. In the future perhaps we should *only* use this method to detect rejections, rather than having a separate code path for rejecting when the promise also rejects. this would greatly simplify the concept.

PR #360, opened 2020-07-06
feat: Deletes and invalidates trigger suspense always

ntucker:
BREAKING CHANGE:
- denormalize has third boolean value to track deletion
- deletes no long remove entities, but replace them with DELETE symbol (exported from normalizr)
- schema of delete shape should be the `new schemas.Delete()`
- useInvalidator()'s function calls will always suspend - even without invalidIfStale
- deleted entities that are required by a useResource() will now cause it to suspend rather than throwing `404`
- required entities missing from network response will now throw error in useResource() just like other unexpected deserializations
- FetchShape type is now just 'read' | 'mutate'. No more 'delete'. (use schema.Delete())

# HEP # 5

### Motivation

- deleted entities should suspend (invalidate)
- missing expected entities from response should throw error indicating malformed response
- imperative invalidation should work on arbitrary Endpoints that may in other cases not want to invalidate
- schema with no required entities never suspends

Overall this should be much more intuitive behavior and generally not require odd implementation detail explanations.

#### Hook behavior

- [x]  useRetrieve() is forced to fetch when:
    - expiresAt is old (imperative invalidation)
    - entity was deleted (sent as third arg)
- [x]  useInvalidator()
    -  set expiresAt to 0. this triggers useRetrieve (already true)
    - deletes results entry. this only allows direct read from the cache if inferred results.
    - sets meta.invalidated to true. this is used to determine whether to throw promise
- [x] useResource() throwing promise is thrown if
    - not ready (inferred results has not enough)
    - deleted entity that is required
    - meta is marked as invalidated
- [x] useResource() throws error if
    - error in meta
    - required entity is missing AND not invalidated (this is the case where response isn't what is expected)
- [x] useDenormalized(), useCache(), useError(), hasUsableData() should always make sense irrespective of anything else (call order should not matter)

#### Internals

- [x] No special deletion internals. Deletes are now just mutates. Deletion is managed by a new type of normalizr schema.
- [x]  add: schemas.Delete(schema) - all entities are replaced with DELETE symbol
    - This means more sophisticated deletes like batch deleted can be implemented
    - Those using Resource without refinement will continue with same deleteShape() interface with no breakages
- [x]  deletion replaces entity with special symbol instead of removing

Now one can represent batch deletes like so

```typescript
    {
      type: 'mutate',
      schema: [new schemas.Delete(this)],
      options,
      getFetchKey: (params: readonly string[]) => {
        return 'DELETE ' + this.url(params);
      },
      fetch: (params: readonly string[]) => {
        return this.fetch('delete', this.url(params)).then(() => params.map(id => ({ id })));
      },
    };
```

### Future work

Get rid of invalidIfStale options

### Special Thanks

Design based on discussion with bruno & @zacharyfmarion


packages/core/src/react-integration/__tests__/integration.web.tsx
```diff
@@ -39,7 +39,9 @@ afterEach(() => {
     removeEventListener('error', onError);
 });
 
-for (const makeProvider of [makeCacheProvider, makeExternalCacheProvider]) {
+for (const makeProvider of [
+  makeCacheProvider /*, makeExternalCacheProvider*/,
```
zacharyfmarion, line 43: why is this commented out?

packages/core/src/react-integration/hooks/hasUsableData.ts
```diff
@@ -2,11 +2,15 @@ import { FetchShape } from '@rest-hooks/core/endpoint';
 
 /** If the invalidIfStale option is set we suspend if resource has expired */
 export default function hasUsableData(
-  cacheReady: boolean,
   fetchShape: Pick<FetchShape<any>, 'options'>,
+  cacheReady: boolean,
+  notDeleted: boolean,
```
zacharyfmarion, line 7: kind of a strange boolean name - why not use `deleted`?

packages/core/src/react-integration/hooks/useCache.ts
```diff
@@ -2,25 +2,60 @@ import { ReadShape, ParamsFromShape } from '@rest-hooks/core/endpoint';
...
   const expiresAt = useExpiresAt(fetchShape, params);
+
+  const state = useContext(StateContext);
+  const [denormalized, ready, notDeleted] = useDenormalized(
+    fetchShape,
+    params,
+    state,
+  );
+  const error = useError(fetchShape, params, ready);
+  const trigger = !notDeleted && !error;
```
zacharyfmarion, line 28: yeah !notDeleted just feels real weird to me

zacharyfmarion: Also trigger what? A little unclear about what this variable name means

packages/core/src/react-integration/hooks/useError.ts
```diff
@@ -3,28 +3,44 @@ import { Schema } from '@rest-hooks/normalizr';
...
     if (!meta) return;
-    if (!meta.error) {
-      // this means we probably deleted the entity found in this result
-      const err: any = new Error(
-        `Resource not found in cache ${fetchShape.getFetchKey(params)}`,
-      );
-      err.status = 404;
-      return err;
+    // this means the response is missing an expected entity
+    if (!meta.error && !meta.invalidated) {
```
zacharyfmarion, line 24: Nice!

packages/core/src/react-integration/hooks/useResource.ts
```diff
@@ -29,14 +30,29 @@ function useOneResource<
...
+  if (
+    !hasUsableData(
+      fetchShape,
+      ready,
+      notDeleted,
+      useMeta(fetchShape, params)?.invalidated,
+    ) &&
+    maybePromise
+  )
+    throw maybePromise;
```
zacharyfmarion, line 54: nit: Personally I'm a big fan of eslint/ prettier enforcing braces if the condition is multi-line

packages/core/src/react-integration/hooks/useRetrieve.ts
```diff
@@ -9,8 +9,11 @@ import { DispatchContext } from '../context';
 export default function useRetrieve<Shape extends ReadShape<any, any>>(
   fetchShape: Shape,
   params: ParamsFromShape<Shape> | null,
+  trigger = false,
```
zacharyfmarion, line 12: again, trigger what?

packages/legacy/package.json
```diff
@@ -59,7 +59,7 @@
     "@rest-hooks/test": "^2.0.0-beta.2"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.0.0-beta.0",
+    "@rest-hooks/core": "^1.0.0-delta.0",
```
zacharyfmarion, line 62: what are we, starting a sorority 😛

packages/core/src/react-integration/__tests__/integration.web.tsx
ntucker: ack! was putting console logs in the test and didn't want to get spammed. great catch!

packages/core/src/react-integration/hooks/useResource.ts
ntucker: mm, excellent point. whcih rule is that, i'll upstream to anansi's root config.

packages/legacy/package.json
ntucker: haha. i realized that with so many packages with continual breaking changes along beta release that I needed to be able to distinguish breaking changes *within* a pre-release. going through greek letters allows me to control that.

packages/core/src/react-integration/hooks/useRetrieve.ts
ntucker: trigger teh retrieval, i.e., fetch. This part is kinda uncomfortable because it extends this hook interface when I'd prefer not to. Would love an alternative way - but essentially I need to make sure that when something is deleted its useResource() throws - which can only happen if useRetrieve() triggers a new fetch. maybe i should just layer another fetch if this one doesn't happen instead of conflating useretreive.

```typescript
maybePromise = useMemo(() => {
  // null params mean don't do anything
  if (!triggerFetch || !params | maybePromise) return;
  return fetch(params);
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [!!maybePromise, triggerFetch]);
```

packages/core/src/react-integration/hooks/hasUsableData.ts
ntucker: yeah,  i should probably change this. initially wanted to keep true/false concept similar to cacheReady. maybe a better name can solve it so it's still true when things are 'ok'.

PR #362, opened 2020-07-07
feat: Resource.fetch() arguments reflect browser fetch()

### LINKED ISSUE #273: Request for more examples: Adding authentication header to every request
I'm new to Rest Hooks, so please forgive me if I'm overlooking something obvious. I'm trying to figure out what's the best way of enhancing all requests (fetch, mutations, etc.) with a custom `Authorization-Token` header. 

I have an `<AuthenticationProvider>` context provider that works with a `useAuthContext` hook like this:
```ts
const MyApp = () => (
  <AuthenticationProvider>
    <CacheProvider>
      ... // the app
    </CacheProvider>
  </AuthenticationProvider>
);
```

And then in one of my screens I can access the session and access token like this:
```ts
const MyScreen = ({ id }: { id: number }) => {
  const { session } = useAuthContext();
  const fetchUser = useResource(PostResource.detailShape(), { id });
  return ( ... );
}
```

I'm having trouble understanding how I can pass the `sesion.accessToken` to my resource on every fetch. I've walked through the documentation and found various pointers such as writing a custom `Manager`, using `fetchOptionsPlugin` or implementing a custom `Resource`. I also considered creating my own `useAuthResource()` hook. However, I fail to understand how all these pieces exactly fit together to achieve what I'm trying to do: Set an `Access-Token` HTTP header on all outgoing requests based on the value of `session.accessToken`.

Ideally I would want to configure the `CacheProvider` (or create my own implementation if necessary) that is aware of the authentication context and automatically sets the `Access-Token` HTTP header if logged in.

It would be great if there would be some additional examples for how to use `rest-hooks` in real-world applications. Overall the documentation is good and helpful, but I simply don't see how everything fits together yet.

ntucker:
This allows using hooks in Resource.getFetchInit(), to
make context-based authentication easy.

BREAKING CHANGE:
- Removed Resource.fetchOptionsPlugin()
- Added Resource.getFetchInit() which is called in shape generators
- Resourece.fetch() interface changed to match browser fetch()

Fixes #273 .

### Motivation

- This allows using hooks in Resource.getFetchInit(), to make context-based authentication easy.
- More standardized interface
- Allows for body payloads that aren't just JSON.

### Solution

- Removed Resource.fetchOptionsPlugin()
- Added Resource.getFetchInit() which is called in shape generators
- Changed Resource.fetch() to `static fetch(input: RequestInfo, init: RequestInit): Promise<any>`

```typescript
class AuthdResource extends Resource {
  static getFetchInit = (init: RequestInit) => {
    const { session } = useAuthContext();
    return {
    ...options,
      headers: {
        ...options.headers,
        'Access-Token': session,
      },
    }
  };
}
```

### Open questions

This does make overriding fetch just to change the url kinda annoying.

### DIFF
--- a/__tests__/common.ts
+++ b/__tests__/common.ts
@@ -79,7 +79,9 @@ export class ArticleResource extends Resource {
       getFetchKey: (params: Readonly<Record<string, string>>) =>
         baseShape.getFetchKey({ ...params, includeUser: true }),
       fetch: (params: object) =>
-        this.fetch('get', this.url({ ...params, includeUser: true })),
+        this.fetch(this.url({ ...params, includeUser: true }), {
+          method: 'get',
+        }),
       schema: this,
     };
   }
@@ -93,7 +95,9 @@ export class ArticleResource extends Resource {
       getFetchKey: (params: Readonly<Record<string, string>>) =>
         baseShape.getFetchKey({ ...params, includeUser: true }),
       fetch: (params: object) =>
-        this.fetch('get', this.listUrl({ ...params, includeUser: 'true' })),
+        this.fetch(this.listUrl({ ...params, includeUser: 'true' }), {
+          method: 'get',
+        }),
       schema: [this],
     };
   }
@@ -153,7 +157,10 @@ export class ArticleResourceWithOtherListUrl extends ArticleResource {
     return {
       ...this.listShape(),
       getFetchKey,
-      fetch: (_params: object) => this.fetch('get', getFetchKey()),
+      fetch: (_params: object) =>
+        this.fetch(getFetchKey(), {
+          method: 'get',
+        }),
     };
   }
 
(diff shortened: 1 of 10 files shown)


__tests__/common.ts
```diff
@@ -80,7 +80,9 @@ export class ArticleResource extends Resource {
       getFetchKey: (params: Readonly<Record<string, string>>) =>
         baseShape.getFetchKey({ ...params, includeUser: true }),
       fetch: (params: object) =>
-        this.fetch('get', this.url({ ...params, includeUser: true })),
+        this.fetch(this.url({ ...params, includeUser: true }), {
+          method: 'get',
+        }),
```
zacharyfmarion, line 85: wow it's wild how happy this makes me

PR #403, opened 2020-08-20
enhance: Maintain referential equality globally

ntucker:
#### SSR

To make SSR easy, the cache itself should be easily serializable/deserializable. Currently the cache stores instances of Entities, which means any deserialization construct would have to deal with knowing how to load the Entity classes. While this may be achievable, this amount of complexity while maintaining codesplitting seems undesirable.

#### Entities nested in entities

Since the flat entities are stored as results of fromJS() (normalized form) - they simply contain ids to their nested counterparts. This makes the typing somewhat odd, but more importantly means that denormalizing them on the fly violates referential equality expectations of any entity. In the past this hasn’t been a huge deal because we mostly don’t use nested entities in retail.

#### Update Performance

Referential guarantees are currently provided by a useDenormalized() hook. Using useMemo() ensures concurrent mode compatibility - but it also means all caches are distinct. This has several downsides.
- requiring recomputation even for repeated denormalizations
- no referential equality guarantees cross-component (though this is unlikely to cause problems, and hasn’t yet, it would be safer to provide global guarantees.)
- This ties the pub/sub system to React itself, limiting our ability to use things like recoil or context selector systems that can improve performance needed for react native, and potentially web with faster updates

### Solution

Denormalize function will take in an extra optional argument that is a denormalization cache. This will be used to short-circuit computations if the pieces exist. This also means useDenormalized() will be simplified - possibly removed later.

The denormalized cache will be provided via context, but maintain referential equality due to its mutable nature.

#### Design constraints

Since concurrent mode compatibility is necessary, extra care is needed when creating a cache that is global. Unlike component-level caches which only need the last value, global caches need to be able to provide answers to questions from disjoint state trees.

Hence, we need a mechanism to look up based on the current state tree, which cache result might exist. Furthermore, once React has discarded a tree, and thus the state representation - these cache answers should be released to be collected from the GC.

This makes [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) an ideal supplementary data structure. However, there is an additional challenge to provide a list of keys that will be outlined below.

#### Cache data structure
```ts
const denormalizedCacheInstance = {
  entities: {
    [key]: {
      [pk]: WeakListMap([og, ...nestedEntities]
               -> { author: { id: 1, name: 'bob' }, content: 'hi' } (fromJS() result)
            )
    }
  }
  results: {
    [key]: WeakListMap([og, ...entities] -> EntireDenormalizedResponse)
  }
}
```
This structure mirrors the state cache itself, with the values of entities, and results being supplemented with a WeakMap.

#### WeakListMap

Mapping a list of referential objects to a value is not actually provided by the WeakMap interface. WeakMap can only map a singular object. To create the necessary data structure, we pull from lisps’ [`cons` data construct](http://www.lispworks.com/documentation/HyperSpec/Body/f_cons.htm). We can recursively apply a WeakMap to each entry in the list to achieve the appropriate mapping. This also resembles the [Trie](https://en.wikipedia.org/wiki/Trie) data structure.

This structure
```ts
WM([og, entitya, entityb] -> denorm(og, entitya, entityb))
```

Is represented by
```ts
WM(og -> WM(entitya -> WM(entityb -> denorm(og, entitya, entityb)))
```

### Future work
- [ ] Store pojos, not Entity instances

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -16,6 +16,9 @@
     "./typescript.svg",
     "./rest_hooks_logo_and_text.svg"
   ],
+  "engines": {
+    "node": ">=6.17.1"
+  },
   "scripts": {
     "build:lib": "cross-env NODE_ENV=production ROOT_PATH_PREFIX='@rest-hooks/core' babel --root-mode upward src --out-dir lib --source-maps inline --extensions '.ts,.tsx,.js' --ignore '**/__tests__/**' --ignore '**/*.d.ts'",
     "build:bundle": "rollup -c",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -17,7 +17,7 @@
     "./rest_hooks_logo_and_text.svg"
   ],
   "engines": {
-    "node": ">=6.17.1"
+    "node": ">=0.12"
   },
   "scripts": {
     "build:lib": "cross-env NODE_ENV=production ROOT_PATH_PREFIX='@rest-hooks/core' babel --root-mode upward src --out-dir lib --source-maps inline --extensions '.ts,.tsx,.js' --ignore '**/__tests__/**' --ignore '**/*.d.ts'",
--- a/packages/core/src/state/selectors/useDenormalized.ts
+++ b/packages/core/src/state/selectors/useDenormalized.ts
@@ -1,7 +1,13 @@
 import { State } from '@rest-hooks/core/types';
 import { ReadShape, ParamsFromShape } from '@rest-hooks/core/endpoint';
 import { DenormalizeNullable } from '@rest-hooks/endpoint';
-import { isEntity, Schema, denormalize } from '@rest-hooks/normalizr';
+import {
+  isEntity,
+  Schema,
+  denormalize,
+  DenormalizeCache,
+  WeakListMap,
+} from '@rest-hooks/normalizr';
 import { useMemo } from 'react';
 
 import buildInferredResults from './buildInferredResults';
@@ -13,6 +19,7 @@ import buildInferredResults from './buildInferredResults';
  * using params and schema. This increases cache hit rate for many
  * detail shapes.
  *
+ *
  * @returns [denormalizedValue, ready]
  */
 export default function useDenormalized<
@@ -21,6 +28,7 @@ export default function useDenormalized<
   { schema, getFetchKey, options }: Shape,
   params: ParamsFromShape<Shape> | null,
   state: State<any>,
+  denormalizeCache: DenormalizeCache = { entities: {}, results: {} },
 ): [
   DenormalizeNullable<Shape['schema']>,
   typeof params extends null ? false : boolean,
@@ -44,18 +52,12 @@ export default function useDenormalized<
   const needsDenormalization = useMemo(() => schemaHasEntity(schema), [schema]);
 
   // Compute denormalized value
-  const [
-    denormalized,
-    entitiesFound,
-    entitiesDeleted,
-    entitiesList,
-  ] = useMemo(() => {
+  return useMemo(() => {
     if (!needsDenormalization)
-      return [cacheResults, cacheResults !== undefined, false, ''] as [
+      return [cacheResults, cacheResults !== undefined, false] as [
         DenormalizeNullable<Shape['schema']>,
         any,
         boolean,
-        string,
       ];
     // Warn users with bad configurations
     /* istanbul ignore next */
@@ -76,26 +78,19 @@ export default function useDenormalized<
     // inferred results are considered stale
     if (options && options.invalidIfStale && !cacheResults) entities = {};
 
+    if (params && !denormalizeCache.results[getFetchKey(params)])
+      denormalizeCache.results[getFetchKey(params)] = new WeakListMap();
+
     // second argument is false if any entities are missing
     // eslint-disable-next-line prefer-const
-    let [denormalized, entitiesFound, entitiesDeleted, cache] = denormalize(
+    return denormalize(
       results,
       schema,
       entities,
-    );
+      denormalizeCache.entities,
+      params ? denormalizeCache.results[getFetchKey(params)] : undefined,
+    ) as [DenormalizeNullable<Shape['schema']>, boolean, boolean];
 
-    // this enables us to keep referential equality based on entities contained within
-    const entitiesList = Object.values(cache)
-      .map(Object.values)
-      .reduce((a: any[], b: any[]) => a.concat(b), [])
-      .join(',');
-
-    return [denormalized, entitiesFound, entitiesDeleted, entitiesList] as [
-      DenormalizeNullable<Shape['schema']>,
-      boolean,
-      boolean,
-      string,
-    ];
     // TODO: would be nice to make this only recompute on the entity types that are in schema
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
@@ -105,14 +100,6 @@ export default function useDenormalized<
     cacheResults,
     needsDenormalization,
     options && options.invalidIfStale,
-  ]);
-
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  return useMemo(() => [denormalized, entitiesFound, entitiesDeleted], [
-    entitiesFound,
-    entitiesDeleted,
-    results,
-    entitiesList,
   ]);
 }
 
--- a/packages/core/src/state/selectors/useDenormalized.ts
+++ b/packages/core/src/state/selectors/useDenormalized.ts
@@ -19,7 +19,6 @@ import buildInferredResults from './buildInferredResults';
  * using params and schema. This increases cache hit rate for many
  * detail shapes.
  *
- *
  * @returns [denormalizedValue, ready]
  */
 export default function useDenormalized<
--- a/packages/normalizr/package.json
+++ b/packages/normalizr/package.json
@@ -30,7 +30,7 @@
   "typings": "lib/index.d.ts",
   "sideEffects": false,
   "engines": {
-    "node": ">=6.17.1"
+    "node": ">=0.12"
   },
   "scripts": {
     "build": "yarn run build:lib && yarn run build:bundle",
--- a/packages/normalizr/package.json
+++ b/packages/normalizr/package.json
@@ -29,6 +29,9 @@
   "module": "lib/index.js",
   "typings": "lib/index.d.ts",
   "sideEffects": false,
+  "engines": {
+    "node": ">=6.17.1"
+  },
   "scripts": {
     "build": "yarn run build:lib && yarn run build:bundle",
     "build:bundle": "run-p build:js:*",
--- a/packages/normalizr/src/WeakListMap.ts
+++ b/packages/normalizr/src/WeakListMap.ts
@@ -1,0 +1,65 @@
+/** Link in a chain */
+class Link<K extends object, V> {
+  children = new WeakMap<K, Link<K, V>>();
+  declare value: V | undefined;
+}
+
+class KeySize extends Error {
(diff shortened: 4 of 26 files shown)


packages/normalizr/src/__tests__/WeakListMap.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    for (const attempt of attempts) {
+      expect(wlm.get(attempt.key)).toBe(attempt.value);
+      expect(wlm.has(attempt.key)).toBe(true);
+    }
+
+    expect(wlm.get([a])).toBeUndefined();
+    expect(wlm.has([a])).toBe(false);
+  });
+
+  it('consideres empty key list invalid', () => {
```
nickcherry, line 76: "considers"

fredbuluku (approved):

packages/normalizr/src/WeakListMap.ts
```diff
@@ -0,0 +1,65 @@
+/** Link in a chain */
+class Link<K extends object, V> {
+  children = new WeakMap<K, Link<K, V>>();
```
ljharb, line 3: What are normalizr's supported engine targets? WeakMap is not polyfillable, so you'd need to fall back to `Map` if you wanted to support engines that lacked it.

packages/normalizr/src/denormalize.ts
```diff
@@ -21,72 +32,131 @@ const unvisitEntity = (
     return [entity, false, false];
   }
 
-  if (!cache[schema.key]) {
-    cache[schema.key] = {};
+  if (!localCache[schema.key]) {
+    localCache[schema.key] = {};
   }
```
ljharb, line 37: if `schema.key` could be any string (or Symbol) value, then you may want to use another approach so a key of, say, `toString` doesn't return the wrong value.

packages/normalizr/src/denormalize.ts
```diff
@@ -21,72 +32,131 @@ const unvisitEntity = (
...
-      isImmutable(entity) || entity instanceof FlatEntity
-        ? entity
-        : schema.fromJS(entity);
-
+  if (!localCache[schema.key][id]) {
+    // make unvisit that tracks entity list here
+    const globalKey: EntityInterface[] = [entity];
+    // every time we nest, we want to unwrap back to the top.
+    // this is due to only needed the next level of nested entities for lookup
+    const originalUnvisit = unvisit.original ? unvisit.original : unvisit;
```
ljharb, line 46:
```suggestion
    const originalUnvisit = unvisit.original || unvisit;
```

packages/normalizr/src/denormalize.ts
```diff
@@ -21,72 +32,131 @@ const unvisitEntity = (
...
-  ] as const;
+    return [input, true, false];
+  }
+
+  // make unvisit that tracks entity list here
+  const globalKey: EntityInterface[] = [];
+  const wrappedUnvisit = (input: any, schema: any) => {
+    const ret: [any, boolean, boolean] = unvisit(input, schema);
+    // pass over undefined in key
+    if (ret[0] && schema && isEntity(schema)) globalKey.push(ret[0]);
```
ljharb, line 140: `ret[0]` here has a type of `any`, and is then pushed into `globalKey`, which means it should have a type of `EntityInterface`. What's the reason not to do:
```suggestion
    const ret: [EntityInterface, boolean, boolean] = unvisit(input, schema);
    // pass over undefined in key
    if (ret[0] && schema && isEntity(schema)) globalKey.push(ret[0]);
```
?

packages/normalizr/src/denormalize.ts
```diff
@@ -21,72 +32,131 @@ const unvisitEntity = (
...
+    return ret;
+  };
+  wrappedUnvisit.original = unvisit;
+
+  return (
+    input: any,
+    schema: any,
+  ): [denormalized: any, found: boolean, deleted: boolean] => {
+    globalKey.push(input);
```
ljharb, line 149: same here?
```suggestion
    input: EntityInterface,
    schema: any,
  ): [denormalized: any, found: boolean, deleted: boolean] => {
    globalKey.push(input);
```

packages/normalizr/src/entities/__tests__/Entity.test.ts
```diff
@@ -445,6 +446,31 @@ describe(`${Entity.name} denormalization`, () => {
...
+    const resultCache = new WeakListMap();
+
+    const first = denormalize('1', Menu, entities, entityCache, resultCache)[0];
+    const second = denormalize(
+      '1',
+      Menu,
+      entities,
+      entityCache,
+      resultCache,
+    )[0];
```
ljharb, line 469: nit:
```suggestion
    const [first] = denormalize('1', Menu, entities, entityCache, resultCache);
    const [second] = denormalize(
      '1',
      Menu,
      entities,
      entityCache,
      resultCache,
    );
```

packages/normalizr/src/WeakListMap.ts
ntucker: Looks like WeakMap starts in v6, which is well beyond what most people support now. I'll add an engine entry to ensure this.

packages/normalizr/src/denormalize.ts
ntucker: isEntity(schema) determines whether it should be pushed. otherwise it is not expected to be EntityInterface.

packages/normalizr/src/denormalize.ts
ntucker: `unvisit` will run on any part of input, which can be things that aren't entities

e.g., in the input:
`[{ car: { id: 1, name: 'tesla' }, items: 5 }, { car: { id: 1, name: 'ford' }, items: 2 }]`

each level will run, so we'll see array, then { car, items}, then finally the { id, name} piece. only that last one is an entity.

packages/normalizr/src/denormalize.ts
ntucker: would `!== undefined` be just as good vs the prototype.call? Just to save some bytes

packages/normalizr/src/denormalize.ts
ljharb: ah, so only the properly typed thing will be pushed into `globalKey`?

packages/normalizr/src/denormalize.ts
ljharb: only as long as nobody's done `Object.prototype.something = undefined`, but i suppose in that case it would still work fine.

packages/normalizr/src/WeakListMap.ts
ljharb: Note that adding an engines entry is semver-major.

packages/normalizr/src/denormalize.ts
ntucker: exactly, that's what the if statement is for

packages/core/package.json
```diff
@@ -16,6 +16,9 @@
     "./typescript.svg",
     "./rest_hooks_logo_and_text.svg"
   ],
+  "engines": {
+    "node": ">=6.17.1"
```
ljharb, line 20:
```suggestion
    "node": ">=6.5"
```
note that this also prevents IE 10, and somewhat IE 11, from being able to use rest-hooks, and a *lot* of people still have to support IE 11.

packages/core/src/state/selectors/useDenormalized.ts
```diff
@@ -13,6 +19,7 @@ import buildInferredResults from './buildInferredResults';
  * using params and schema. This increases cache hit rate for many
  * detail shapes.
  *
+ *
```
ljharb, line 22:
```suggestion
```

packages/normalizr/package.json
```diff
@@ -29,6 +29,9 @@
   "module": "lib/index.js",
   "typings": "lib/index.d.ts",
   "sideEffects": false,
+  "engines": {
+    "node": ">=6.17.1"
```
ljharb, line 33:
```suggestion
    "node": ">=0.12"
```
apparently WeakMap is in node 0.12, so might as well have maximal back compat?

packages/normalizr/src/__tests__/index.test.js
```diff
@@ -602,3 +603,171 @@ describe('denormalize', () => {
...
+      entityCache,
+      resultCache,
+    )[0];
+    const second = denormalize(
+      result,
+      [Tacos],
+      entities,
+      entityCache,
+      resultCache,
+    )[0];
```
ljharb, line 631: nit:
```suggestion
    const [first] = denormalize(
      result,
      [Tacos],
      entities,
      entityCache,
      resultCache,
    );
    const [second] = denormalize(
      result,
      [Tacos],
      entities,
      entityCache,
      resultCache,
    );
```

packages/normalizr/src/__tests__/index.test.js
```diff
@@ -602,3 +603,171 @@ describe('denormalize', () => {
...
+    expect(first).toBe(second);
+
+    const third = denormalize(
+      [...result],
+      [Tacos],
+      entities,
+      entityCache,
+      resultCache,
+    )[0];
```
ljharb, line 640: nit:
```suggestion
    const [third] = denormalize(
      [...result],
      [Tacos],
      entities,
      entityCache,
      resultCache,
    );
```

packages/normalizr/src/__tests__/index.test.js
```diff
@@ -602,3 +603,171 @@ describe('denormalize', () => {
...
+              body: 'This article is great.',
+              comments: ['comment-123-4738'],
+              id: '123',
+              title: 'A Great Article',
+            },
+          },
+        },
+        entityCache,
+        resultCache,
+      )[0];
```
ljharb, line 767:
```suggestion
      const [first] = denormalize(
        result,
        { data: Article },
        entities,
        entityCache,
        resultCache,
      );
      const [second] = denormalize(
        result,
        { data: Article },
        {
          ...entities,
          Article: {
            123: {
              author: '8472',
              body: 'This article is great.',
              comments: ['comment-123-4738'],
              id: '123',
              title: 'A Great Article',
            },
          },
        },
        entityCache,
        resultCache,
      );
```

packages/core/package.json
ntucker: https://caniuse.com/?search=JavaScript%20WeakMap what's the 'somewhat' ie11 about? ie10 isn't supported.

packages/rest-hooks/package.json
```diff
@@ -16,6 +16,9 @@
     "./typescript.svg",
     "./rest_hooks_logo_and_text.svg"
   ],
+  "engines": {
+    "node": ">=6.17.1"
```
ljharb, line 20:
```suggestion
    "node": ">=0.12"
```

ljharb (approved):
LGTM but would be good to get more eyes on it if possible

packages/core/package.json
ljharb: http://kangax.github.io/compat-table/es6/ indicates that IE 11 has partial WeakMap support.

ljharb: (caniuse is not an effective choice for language things; the kangax table is the best source)

ntucker: node v4 is needed for some of the weakmap() features according to https://node.green/

ntucker: Looks like the only potential for IE11 is if people somehow try to inject frozen objects. That's not recommended anywhere in the docs so probably a fine tradeoff for them to consider.

ljharb: For iterator closing, sure, but you're not using that feature. It'd be node v6.5 if you needed that feature.

PR #407, opened 2020-09-04
feat: useFetchDispatcher()

ntucker:
### Motivation

Fetching unknown number of items? Never fear, this lets you use one hook with many different fetches!

We'll be using this for our fetch-as-render SSR + concurrent mode routing architecture.

#### SSR Concept

This hook is always mounted.
```ts
import { dispatchFetch } from 'rest-hooks';
import { useLocale } from 'i18n';

function useRouteContext() {
  const locale = useLocale();
  const dispatchFetch = useFetchDispatcher();
  return { locale, dispatchFetch };
}
```

Each of these routes resolveData will be run on navigation even stack.
```ts
const routes = [
  {
    path: '/home',
    modulePath: 'Home',
    resolveData: ({ dispatchFetch, locale }) => {
      return dispatchFetch(
        AssetResource.listShape(),
        {
          locale: locale?.currency,
          localeCode: locale?.localeCode
        }
      );
    },
  },
  {
    path: '/careers/:position-id',
    modulePath: 'Careers',
    resolveData: (({ dispatchFetch, params: { positionId })) => {
      return dispatchFetch(PositionResource.detailShape(), { positionId });
    },
  },
]
```

This design allows for multiple, variable number and types of endpoints to fetch without any re-renders.

### Open questions

Should this be the new default, or simply supplement useFetcher()? It doesn't add verbosity really and is more powerful.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -16,6 +16,7 @@ export { useDenormalized } from './state/selectors';
 export {
   useCache,
   useFetcher,
+  useFetchDispatcher,
   useRetrieve,
   useResource,
   useSubscription,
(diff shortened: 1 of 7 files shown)


kmartinezmedia:
oh i really like this

PR #410, opened 2020-09-08
fix: Only block suspense if errors are synthetic

### LINKED ISSUE #408: Surprising behavior using `errorExpiryLength`
We've been surprised by the behavior of network errors, specifically that they're cached. We thought we'd worked around a problem, described below, by using `errorExpiryLength` however it ended up working in a way that was surprising at first.

Our use case is basically:

1. We try and load a `list` using `useResource`
2. The list network fall fails for whatever reason
3. We show a helpful error, using `NetworkErrorBoundary`, and allow the user to try again.
4. We make the same `list` call, but the error is now cached and thus rethrown by `useResource`, so the `NetworkErrorBoundary` is rendered again.

We tried setting `errorExpiryLength` to `0`, with the thought that this would disable caching of errors. This didn't work, and looking at the source of `useError` seems to indicate that it doesn't have any interaction with `errorExpiryLength` at all - you can only invalidate the resource's error using `useInvalidator`. Setting `errorExpiryLength` does seem to cause the resource to re-fetch, but by then it has already thrown the error and we've rendered the error boundary.

This seems un-intuitive: our thought was that by setting `errorExpiryLength` errors would not be cached, but it doesn't look like there's any way to do that. Our current workaround is to call `useResetter` at the error boundary, completely nuking the cache, but ideally we could just have a way of configuring resources to not cache errors. Perhaps a counterpart to `invalidIfStale` for errors specifically?

ntucker:
Fixes #408 .

### Motivation

In https://github.com/coinbase/rest-hooks/pull/360 we solve an infinite loop problem when the network response was invalid. This required detecting and throwing errors before suspending so it doesn't just keep suspending infinitely.

However, this means that any genuine network errors will now prevent suspending for new content.

### Solution

For the one case of response value, we'll mark the error as `synthetic`. Only `synthetic` errors throw before considering suspense.

Also, Fixtures (used in various testing pieces) now support faking errors, by sending error: true.

### DIFF
--- a/packages/core/src/react-integration/__tests__/useResource.web.tsx
+++ b/packages/core/src/react-integration/__tests__/useResource.web.tsx
@@ -377,6 +377,65 @@ describe('useResource()', () => {
     expect((result.error as any).status).toBe(403);
   });
 
+  it('should suspend when already has a network error', async () => {
+    const error: any = { message: 'network error', status: 400 };
+    const FS = { ...CoolerArticleResource.detailShape() };
+    FS.options = { ...FS.options, errorExpiryLength: -100 };
+    const { result, waitForNextUpdate } = renderRestHook(
+      () => {
+        return useResource(CoolerArticleResource.detailShape(), {
+          title: '0',
+        });
+      },
+      {
+        results: [
+          {
+            request: FS,
+            params: { title: '0' },
+            result: error,
+            error: true,
+          },
+        ],
+      },
+    );
+    expect(result.current).toBe(null);
+    expect(result.error).toBe(null);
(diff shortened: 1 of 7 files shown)


packages/core/src/react-integration/__tests__/useResource.web.tsx
```diff
@@ -377,6 +377,65 @@ describe('useResource()', () => {
     expect((result.error as any).status).toBe(403);
   });
 
+  it('should suspend when already has a network error', async () => {
```
ntucker, line 380: @praxxis This is your use case

PR #422, opened 2020-10-29
fix: Clear promises on cleanup

ntucker:
### Motivation

In React Native, when the app is closed, the react tree unmounts. However, the JS runtime still persists. Because of this, module defined NetworkManagers will not be reset, but will have clean called.

Additionally, this race condition also exists with RESETs

### Solution

Clean should also prepare NetworkManager to be still used.


ljharb (approved):
seems reasonable

jarrettkong (approved):

PR #427, opened 2020-11-20
feat: Add RestEndpoint type

### LINKED ISSUE #423: Incorrect output from useResource() when overriding schema in listShape
**React version** 16.13.1

**Concurrent mode** no

**Describe the bug**
I am using rest-hooks 4.5.9. I am working with an API that wraps a list in an object, just like [this example in the documentation](https://resthooks.io/docs/guides/endpoints#example-schema). That is, for the list URL, I get something like:
```
{ data: [item1, item2, ...] }
```
I followed the instructions in the documentation and overrode the `schema` in the `listUrl` in my resource class, let's call it `ItemResource`. I would now expect `useResource(ItemResource.listShape(), {})` to return an `ItemResource[]`. The TypeScript typings seem to agree with me. However, what it actually returns is `{ data: ItemResource[] }`.

**To Reproduce**
1. Create a subclass of `Resource` that overrides `listShape()`, overwriting the `schema` property with `{ data: [this.asSchema()] }`, as show in [the documentation](https://resthooks.io/docs/guides/endpoints#resource-definition).
2. Try to use this resource to fetch data from an API (or mocked API) that returns data in the corresponding format.

**Expected behavior**
`useResource()` should always return an array of the resource regardless of how the `schema` is defined.

**Additional context**
Great library, big fan!

ntucker:
Fixes #423 .

### Motivation

Explicitly typing return values is encouraged by common eslint rules and helps TypeScript performance as well as reducing recursion limit hit rate. Let's make it both as easy as possible to do, and clearly explain how to do it.

### Solution

- We are adding a new convenience type `RestEndpoint` as well as `RestFetch` to make explicit endpoint return types for Resource easier
- Adding thorough documentation about typing Resource endpoints

### DIFF
--- a/docs/api/types.md
+++ b/docs/api/types.md
@@ -128,7 +128,7 @@ for easy customization.
 
 ```typescript
 export interface RestEndpoint<
-  F extends FetchFunction = FetchFunction,
+  F extends FetchFunction = RestFetch,
   S extends Schema | undefined = Schema | undefined,
   M extends true | undefined = true | undefined
 > extends EndpointInstance<F, S, M> {
@@ -142,4 +142,17 @@ export interface RestEndpoint<
   method: string;
   signal: AbortSignal | undefined;
 }
+```
+
+### RestFetch
+
+Fetch function for Resources. Unlike [FetchFunction](#fetchfunction), these require the params variable
+as [Resource](./resource) expects it
+
+```typescript
+export type RestFetch<
+  P = object,
+  B = RequestInit['body'] | Record<string, any>,
+  R = any
+> = (params: P, body?: B) => Promise<R>;
 ```
--- a/docs/api/types.md
+++ b/docs/api/types.md
@@ -1,0 +1,145 @@
+---
+title: Types
+---
+
+## @rest-hooks/core
+
+### Manager
+
+```typescript
+interface Manager {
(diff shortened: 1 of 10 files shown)


docs/guides/rest-types.md
```diff
@@ -0,0 +1,147 @@
+---
```
ntucker, line 1: This is for rest hooks 5.

website/versioned_docs/version-4.5/guides/rest-types.md
```diff
@@ -0,0 +1,126 @@
+---
```
ntucker, line 1: this is 4.5 rest hooks version

ntucker:
@michael-swarm

docs/api/types.md
```diff
@@ -0,0 +1,158 @@
...
+  getFetchInit: (
+    this: any,
+    body?: RequestInit['body'] | Record<string, any>,
+  ) => any;
+  method: string;
+  signal: AbortSignal | undefined;
+}
+```
+
+### RestFetch
```
ntucker, line 147: Or should this be called RestFunction?

michael-swarm:
Sorry for the late reply, this looks awesome, thank you!

PR #439, opened 2020-12-16
feat: Track and use entity resolution time

ntucker:
### Motivation

Tracking entity specific resolution time helps improve entity freshness in cases where custom data-specific merging tracking is not provided. There are several cases that will shortly be using the improved tracking of data age at an entity level:

- stale-while-revalidation staleness guarantees when inferring results
- resolving fully normalized persistence caches (@0xcaff )

### Solution

Add entityMeta as top level `state` member. This tracks the same shape as entities, but with the data of { date: number }.

```typescript
  entityMeta: {
    readonly [entityKey: string]: {
      readonly [pk: string]: { readonly date: number };
    };
  };
```

This affects merges by potentially changing the order of arguments sent to [SimpleRecord.merge()](https://resthooks.io/docs/next/api/SimpleRecord#static-merget-extends-typeof-simplerecordfirst-instancetypet-second-instancetypet--instancetypet). The second argument is considered 'priority' over the first - as far as the rest hooks cache is considered. User-implemented merge functions can either factor that in to their merge or choose to ignore it.

Previously, the currently processed data was sent as the second 'priority' argument to take precedence over any existing entities. However, now if there is any staleness information, it is used to inform said ordering.

For example, the default merge() copies the 'incoming' replacing 'existing' properties when found.

```typescript
  static merge<T extends typeof SimpleRecord>(
    this: T,
    existing: AbstractInstanceType<T>,
    incoming: AbstractInstanceType<T>,  // this is the priority
  ) {
    const props = Object.assign(
      this.toObjectDefined(existing),
      this.toObjectDefined(incoming),
    );
    return this.fromJS(props);
  }
```

### Alternatives considered

We did consider storing the recency data in the entity itself.

Pros:
- potentially more efficient updates
- simpler data traversal

Cons:
- we infect 'normalizr' code with concerns of the cache implementation
- awkward extra data in what should be an object that most exactly reflects what the user expects.

Another note is that in the future when Entities themselves are not actually stored in state, we could shuffle around where the date information is stored.

### DIFF
--- a/packages/core/src/state/merge/mergeDeepCopy.ts
+++ b/packages/core/src/state/merge/mergeDeepCopy.ts
@@ -5,7 +5,12 @@ import isMergeableObject from './isMergeable';
 /**
  * Deep merge two objects or arrays. Uses static merge function if exists.
  */
-export default function mergeDeepCopy<T1, T2>(target: T1, source: T2): T1 & T2 {
+export default function mergeDeepCopy<T1, T2>(
+  target: T1,
+  source: T2,
+  targetMeta?: any,
+  sourceDate?: any,
+): T1 & T2 {
   const sourceIsArray = Array.isArray(source);
   const targetIsArray = Array.isArray(target);
   const sourceAndTargetTypesMatch = sourceIsArray === targetIsArray;
@@ -13,7 +18,12 @@ export default function mergeDeepCopy<T1, T2>(target: T1, source: T2): T1 & T2 {
 
   if (target && Static && isMergeable(Static)) {
     if (isMergeable((target as any).constructor)) {
-      return Static.merge(target, source) as any;
+      // second argument takes priority over first
+      // if either of these is undefined, it resolves to 'false' which
+      // means we fallback to 'newer' (source) takes priority
+      return (targetMeta?.date > sourceDate
+        ? Static.merge(source, target)
+        : Static.merge(target, source)) as any;
     } else {
       return source as any;
     }
@@ -27,6 +37,8 @@ export default function mergeDeepCopy<T1, T2>(target: T1, source: T2): T1 & T2 {
         destination[key] = mergeDeepCopy(
           destination[key],
           (source as any)[key],
+          targetMeta?.[key],
+          sourceDate,
         );
       });
       return destination;
--- a/packages/core/src/react-integration/provider/__tests__/__snapshots__/provider.tsx.snap
+++ b/packages/core/src/react-integration/provider/__tests__/__snapshots__/provider.tsx.snap
@@ -13,6 +13,13 @@ Object {
       },
     },
   },
+  "entityMeta": Object {
+    "http://test.com/article-cooler/": Object {
+      "5": Object {
+        "date": 50,
+      },
+    },
+  },
   "indexes": Object {},
   "meta": Object {
     "http://test.com/article-cooler/5": Object {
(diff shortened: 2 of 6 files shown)


packages/core/src/state/merge/mergeDeepCopy.ts
```diff
@@ -5,15 +5,25 @@ import isMergeableObject from './isMergeable';
 /**
  * Deep merge two objects or arrays. Uses static merge function if exists.
  */
-export default function mergeDeepCopy<T1, T2>(target: T1, source: T2): T1 & T2 {
+export default function mergeDeepCopy<T1, T2>(
+  target: T1,
+  source: T2,
+  targetMeta?: any,
```
ljharb, line 11: couldn't this have a more useful type?
```suggestion
  targetMeta?: {
    data?: T2['date'],
    [k: string]: unknown,
  },
```

ntucker: that breaks in all usages in the function. this is recursively evaluated but since its not publicly exposed just making the typings simple seemed to work best. Furthermore, typescript disallows > comparisons for things possibly undefined, but we actually want to use the javascript behavior here to resolve that to false for fallbacks.

ljharb: gotcha. would `unknown` be better than `any`?

ntucker: unknown doesn't let me do ?. on it

ljharb: hm, it should - i guess TS types are broken for optional chaining.

ntucker: My theory is that they just think treating types as loosely as javascript does (where everything is an object) is not a good idea. They also do this for instance where I cannot do < comparison on undefined, even tho that has a valid result.

ntucker: I subscribed to that issue now :)

ljharb (approved):
not a fan of `any` anywhere, but i suppose this is the best it can be for now due to TS limitations.

ntucker:
Ya, I'm there with you. I feel less bad since its not exposed even as a 'hidden' export. They have made a lot of improvements about expressing recursive types recently so progress is being made.

packages/core/src/state/merge/mergeDeepCopy.ts
```diff
@@ -5,15 +5,25 @@ import isMergeableObject from './isMergeable';
...
   const Static: StaticType<T2> = source && (source as any).constructor;
 
   if (target && Static && isMergeable(Static)) {
     if (isMergeable((target as any).constructor)) {
-      return Static.merge(target, source) as any;
+      // second argument takes priority over first
+      // if either of these is undefined, it resolves to 'false' which
+      // means we fallback to 'newer' (source) takes priority
+      return (targetMeta?.date > sourceDate
+        ? Static.merge(source, target)
```
0xcaff, line 25: perhaps it is worthwhile to send more information from the resolution (maybe an object with date and which manager fired the action or the entire action) and let the a default implementation of merge control the behavior.

i can't think of an example where this would be useful though so I don't think we should worry about it.

0xcaff (approved):
yay! a step towards per-entity caching

packages/core/src/state/merge/mergeDeepCopy.ts
ntucker: Interesting idea. I prefer to minimize public interfaces as well as coupling of concerns without a clear need. If/when this need is realized the interface for such a thing will be more clear. It is easier to add than change.

I'll keep this in mind though!

PR #440, opened 2020-12-16
feat: Add FixtureManager to improve testing imperative fetch

### LINKED ISSUE #415: Example: Unit test for useFetcher() and Resource.createShape()
You have a nice example on how to unit test `useResource()` together with `Resource.detailShape()`: https://resthooks.io/docs/api/makeRenderRestHook#example

There is no example on how to do it for `useFetcher()` and `Resource.createShape()`

Similar example as mentioned above.

ntucker:
Fixes #415.

### Motivation

In many cases, mocking network itself is inconvenient - which is why MockProvider was created. However, it is incredibly limited in its capability as it cannot handle any updates, thus cannot model imperative pieces.

We wish to have an easy, but universal solution to mocking with fixtures.

### Solution

FixtureManager is initialized with fixtures it can handle and automatically dispatches receives in the normal manner when it receives a 'fetch' dispatch.

If it does not find a matching fixture for a given request, it will pass through to other managers. To fallback to normal behavior, simply include those other managers.

### Usage

`fixtures.ts`
```typescript
export default {
  full: [
    {
      request: ArticleResource.list(),
      params: { maxResults: 10 },
      result: [
        {
          id: 5,
          content: 'have a merry christmas',
          author: 2,
          contributors: [],
        },
        {
          id: 532,
          content: 'never again',
          author: 23,
          contributors: [5],
        },
      ],
    },
   {
      request: ArticleResource.create(),
      params: { },
      result: {
          id: 51,
          content: 'new item created',
          author: 2,
          contributors: [],
        },
    },
  ],
}
```

`articleStory.tsx`
```typescript
import { select } from '@storybook/addon-knobs';
import { CacheProvider } from '@rest-hooks/core';
import { FixtureManager } from '@rest-hooks/test';
import ArticleList from 'ArticleList';
import options from './fixtures';

const label = 'Data';
const defaultValue = options.full;
const groupId = 'GROUP-ID1';

storiesOf('name', module).add('Name', () => {
  const results = select(label, options, defaultValue, groupId);
  const managers = [new FixtureManager(results)];
  return (
    <CacheProvider managers={managers}>
      <ArticleList maxResults={10} />
    </MockProvider>
  );
});
```

### Future work

This would be a breaking change, but would better consolidate:
- Make MockProvider be a simple proxy to simply use FixtureNetworkManager, so it's one less line to use.
- Make this work with hooks testing as well so people can choose to use this with hook tests rather than mocking with things like 'nock'

### TODO for this PR:

- [ ] Test
- [ ] Update docs

### DIFF
--- a/packages/test/src/FixtureManager.ts
+++ b/packages/test/src/FixtureManager.ts
@@ -1,0 +1,67 @@
+import { MiddlewareAPI, Middleware } from '@rest-hooks/use-enhanced-reducer';
+import { NetworkManager, actionTypes, ReceiveAction } from '@rest-hooks/core';
+import type { Dispatch, Manager } from '@rest-hooks/core';
+
+import { Fixture, actionFromFixture } from './mockState';
+
+/** Resolves requests matching provided fixtures
+ *
+ */
+export default class FixtureManager implements Manager {
+  protected mockResults: { [key: string]: ReceiveAction } = {};
+  protected declare middleware: Middleware;
+
+  constructor(fixtures: Fixture[]) {
+    for (const fixture of fixtures) {
+      const { key, action } = actionFromFixture(fixture);
+      this.mockResults[key] = action;
+    }
+
+    this.middleware = <R extends React.Reducer<any, any>>({
+      dispatch,
+      getState,
+    }: MiddlewareAPI<R>) => {
+      return (next: Dispatch<R>) => (
+        action: React.ReducerAction<R>,
+      ): Promise<void> => {
+        switch (action.type) {
+          case actionTypes.FETCH_TYPE: {
+            const { key } = action.meta;
+
+            if (key in this.mockResults) {
+              dispatch(this.mockResults[key] as any);
+              return Promise.resolve();
+            }
+            console.warn(
+              `FixtureManager received a dispatch:
+        ${JSON.stringify(action, undefined, 2)}
+        for which there is no matching fixture.
+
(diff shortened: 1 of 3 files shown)


packages/test/src/mockState.ts
```diff
@@ -24,34 +24,36 @@ export interface ErrorFixture {
 
 export type Fixture = SuccessFixture | ErrorFixture;
 
+export function actionFromFixture(fixture: Fixture) {
```
ntucker, line 27: extracted this to be shared, but nothing really changed

packages/test/src/FixtureManager.ts
```diff
@@ -0,0 +1,67 @@
...
+      return (next: Dispatch<R>) => (
+        action: React.ReducerAction<R>,
+      ): Promise<void> => {
+        switch (action.type) {
+          case actionTypes.FETCH_TYPE: {
+            const { key } = action.meta;
+
+            if (key in this.mockResults) {
+              dispatch(this.mockResults[key] as any);
+              return Promise.resolve();
```
ljharb, line 33: should this be an async function, and this line just be `return`?

ntucker: We dont' use async functions to avoid the huge added code when transpiling with babel to target older browsers since this is a library.

ljharb: ah makes sense.

ljharb (approved):

PR #444, opened 2020-12-24
Expose BodyFromShape from core library

jarrettkong:
### Motivation

When extracting the request type from a fetch shape, `get` and `mutate` requests have the requests in either Params or Body of the fetch shape. `ParamsFromShape` was exposed from the package, but `BodyFromShape` wasn't, necessitating the method be copied into the codebase directly until it is exposed.

### Solution

One line change to export `BodyFromShape` from `packages/core`.

### Open questions

What is the appropriate way to update the version of the package?

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -42,5 +42,6 @@ export type { FetchShape, ReadShape, MutateShape } from './endpoint/shapes';
 export type {
   SetShapeParams,
   ParamsFromShape,
+  BodyFromShape,
   ReturnFromShape,
 } from './endpoint/types';


ljharb:
Isn't this already exposed with `import { BodyFromShape } from '@rest-hooks/core/lib/endpoint/types'` https://unpkg.com/browse/@rest-hooks/core@1.0.0-beta.12/lib/endpoint/types.d.ts?

ntucker:
LGTM. Tho my write access still hasn't be re-established.

jarrettkong:
> Isn't this already exposed with `import { BodyFromShape } from '@rest-hooks/core/lib/endpoint/types'` https://unpkg.com/browse/@rest-hooks/core@1.0.0-beta.12/lib/endpoint/types.d.ts?

@ljharb it might be, but `ParamsFromShape` is directly exported from `rest-hooks`: `import { ParamsFromShape } from 'rest-hooks'` so I want to expose `BodyFromShape` in the same way. cc @ntucker this is the right way to do this right?

ntucker:
Generally I would consider any exports `@rest-hooks/core/lib...` to be dangerous as the file layout or even the fact that they export to lib is considered 'internal' information and subject to change without a semver bump. Because of this, I agree with the PR very much.

ntucker:
@jarrettkong You might want to make this not a draft if you want it to be reviewed

jarrettkong:
Left it as a draft since I hadn't updated the version yet and thus considered it WIP still

ljharb:
@ntucker to be clear, without the "exports" field, nothing is internal and it's technically semver-major to remove any of those files or imports. Totally fine to export it here tho and encourage that as a more reliable interface.

ntucker:
The node version uses main.... Which is one giant file. Lib is only for es6. So without es6 you can't even use it, especially since it uses the export syntax

ntucker:
Exports looks cool but seems new. When was it introduced and does yarn one support it, and webpack?

ljharb:
It was added in node v12.17+, v13.7+, and v14+. Yarn is irrelevant, since yarn/npm don't deal with require or import. I'm pretty sure webpack does support it.

Adding it is a breaking change, but once it's added, it defines the entrypoints of your package in a standard way.

ntucker:
Can't find reference in webpack 4. (Searching just goes to webpack 5 site even when on the v4 version) I remember being excited about this to solve transpilation targetting problems but hit a brick wall with support. Might just be because node 10 is still widely supported (I think consumer still runs on it maybe? or maybe we upgraded right before I left, can't remember)

ljharb:
You can easily design your package tho so that it’s backwards compatible (which I’d recommend). The benefit is that for the first time, packages can have internals.

ntucker:
exports overwrites main and module? But other tools look at those if they don't understand? If so will definitely add this in another PR

ljharb:
"module" isn't a thing, and yes, "exports" replaces "main" (you'd still want a "main" for back compat), but also "main" has never been the only entrypoint for a package - "every possible specifier" is. Try `npx ls-exports package rest-hooks` for an example :-), and `npx ls-exports path .` before/after adding "exports" to the package.

ljharb (approved):

PR #445, opened 2020-12-25
fix: FixtureManager should dispatch receive async to not break react

ntucker:
### Motivation

The fetch comes down the render pass in children components, so the parent was already rendered at that point. We need to schedule another render call async or React complains.

### Solution

Settimeout with 0

### DIFF
--- a/packages/test/src/FixtureManager.ts
+++ b/packages/test/src/FixtureManager.ts
@@ -26,12 +26,15 @@ export default class FixtureManager implements Manager {
       ): Promise<void> => {
         switch (action.type) {
           case actionTypes.FETCH_TYPE: {
-            const { key } = action.meta;
+            const { key, resolve, reject } = action.meta;
 
             if (key in this.mockResults) {
               // All updates must be async or React will complain about re-rendering in same pass
               setTimeout(() => {
-                dispatch(this.mockResults[key] as any);
+                const action: any = this.mockResults[key];
+                dispatch(action);
+                const complete = action.error ? reject : resolve;
+                complete(action.payload);
               }, 0);
               return Promise.resolve();
             }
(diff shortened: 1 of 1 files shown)


packages/test/src/FixtureManager.ts
```diff
@@ -29,7 +29,10 @@ export default class FixtureManager implements Manager {
             const { key } = action.meta;
 
             if (key in this.mockResults) {
-              dispatch(this.mockResults[key] as any);
+              // All updates must be async or React will complain about re-rendering in same pass
+              setTimeout(() => {
+                dispatch(this.mockResults[key] as any);
+              }, 0);
               return Promise.resolve();
```
ljharb, line 36:
```suggestion
              return new Promise((resolve) => {
                resolve(dispatch(this.mockResults[key] as any));
              });
```
that way if it throws or returns a promise, it'll be caught/handled

ljharb (approved):
Is there a regression test for this?

packages/test/src/FixtureManager.ts
ntucker: Yeah, got handling rejection in next commit. I'll just push that as part of this PR as well

packages/test/src/FixtureManager.ts
```diff
@@ -26,10 +26,16 @@ export default class FixtureManager implements Manager {
...
           case actionTypes.FETCH_TYPE: {
-            const { key } = action.meta;
+            const { key, resolve, reject } = action.meta;
 
             if (key in this.mockResults) {
-              dispatch(this.mockResults[key] as any);
+              // All updates must be async or React will complain about re-rendering in same pass
+              setTimeout(() => {
+                const action: any = this.mockResults[key];
+                dispatch(action);
```
ljharb, line 35: this will still not reject if `dispatch` throws

ntucker: reducer will never throw due to try/catch. I guess if a user created manager throws this could cause suspense to never resolve. not sure we should send that error to the component tho. I'll put it in a finally to be sure

ljharb (approved):

PR #446, opened 2020-12-30
feat: Add <MockResolver />

### LINKED ISSUE #415: Example: Unit test for useFetcher() and Resource.createShape()

ntucker:
Fixes #415.

### Motivation

Mocking needs to work for imperative endpoints like `update`, not just `read` endpoints as requested in #415.

Initially a new manager was added in #440 (FixtureManager) to solve this problem. However, in practice this became cumbersome for three reasons:

- Full Provider, ErrorBoundary, and Suspense would need to be established in each story.
- Integration with a users' customer managers would need to be re-specified.
- Any changing of fixtures required resetting the entire Provider.

While the first two are only somewhat annoying, the third seemed like a pretty big deal breaker.

### Solution

Use `dispatch interceptor pattern` via `<MockResolver fixtures={myMockData} />`. 

Since all fetches using rest hooks require an established context, we can bind a new context to intercept dispatches before they
even reach the managers.

This design is greatly beneficial for the following reasons:

- Completely decoupled with the rest of rest-hooks usage (both Provider, as well as components).
- Can be composed in normal React manner to override any other fixtures
- Maintains any manager customizations even in this mock environment.

```tsx
const results = [
  {
    request: ArticleResource.list(),
    params: { maxResults: 10 },
    result: [
      {
        id: 5,
        content: 'have a merry christmas',
        author: 2,
        contributors: [],
      },
      {
        id: 532,
        content: 'never again',
        author: 23,
        contributors: [5],
      },
    ],
  },
];
const Template: Story = () => (
  <MockResolver fixtures={results}><MyComponentToTest /></MockResolver>
);
export const MyStory = Template.bind({});
```

### Open questions

- Plan to remove FixtureManager in subsequent PR
- Is there still a use case for MockProvider?
  - It does have the advantage of instantly resolving vs one-tick resolving. This is meaningful in a node-test environment. However, `mockInitialState()` can simply be used in its absence.

### Future work
- Add resolution time mocking to fixture specification (like `nock` does)

### Testing
- Fully tested in `yarn link` storybook environment.
- Automated testing for test package may come later.

### DIFF
--- a/docs/api/MockResolver.md
+++ b/docs/api/MockResolver.md
@@ -1,35 +1,41 @@
 ---
-title: <MockProvider />
+title: <MockResolver />
 ---
 
 ```typescript
-function MockProvider({
+function MockResolver({
   children,
   results,
 }: {
-  children: React.ReactChild;
+  children: React.ReactNode;
   results: Fixture[];
 }): JSX.Element;
 ```
 
-\<MockProvider /> is a simple substitute provider to prefill the cache with fixtures so the 'happy path'
-can be tested. This is useful for [storybook](../guides/storybook.md) as well as component testing.
+\<MockResolver /> enables easy loading of fixtures to see what different network responses might look like.
+This is useful for [storybook](../guides/storybook.md) as well as component testing.
 
-> Deprecated: Use [<MockResolver />](./mockResolver) instead as it also supports [imperative fetches](../api/useFetcher) like [create](../api/resource#create-endpoint) and [update](../api/resource#update-endpoint).
-
-> Note: \<MockProvider /> disables dispatches, thus no fetches will occur. To simply initalize the
-> cache, use [mockInitialState()](./mockInitialState) to construct initialState for the normal [\<CacheProvider />](./CacheProvider)
 
 ## Arguments
 
-### results
+### fixtures
 
 ```typescript
-interface Fixture {
-  request: ReadEndpoint;
+export interface SuccessFixture {
+  request: ReadShape<Schema, object>;
   params: object;
   result: object | string | number;
+  error?: false;
 }
+
+export interface ErrorFixture {
+  request: ReadShape<Schema, object>;
+  params: object;
+  result: Error;
+  error: true;
+}
+
+export type Fixture = SuccessFixture | ErrorFixture;
 ```
 
 This prop specifies the fixtures to use data from. Each item represents a fetch defined by the
@@ -45,8 +51,8 @@ Renders the children prop.
 
 ## Example
 
-```typescript
-import { MockProvider } from '@rest-hooks/test';
+```tsx
+import { MockResolver } from '@rest-hooks/test';
 
 import ArticleResource from 'resources/ArticleResource';
 import MyComponentToTest from 'components/MyComponentToTest';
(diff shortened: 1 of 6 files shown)


docs/api/MockResolver.md
```diff
@@ -0,0 +1,86 @@
...
+    result: [
+      {
+        id: 5,
+        content: 'have a merry christmas',
+        author: 2,
+        contributors: [],
+      },
+      {
+        id: 532,
+        content: 'never again',
```
ljharb, line 73: a horror movie, in two parts

packages/test/src/MockResolver.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * Place below <CacheProvider /> and above any components you want to mock.
+ */
+export default function MockResolver({ children, fixtures }: Props) {
+  const fetchToReceiveAction = useMemo(() => {
+    const actionMap: Record<string, ReceiveAction> = {};
+    for (const fixture of fixtures) {
+      const { key, action } = actionFromFixture(fixture);
+      actionMap[key] = action;
+    }
+    return actionMap;
```
ljharb, line 30:
```suggestion
    const actionMap: Record<string, ReceiveAction> = fromEntries(fixtures.map((fixture) => {
      const { key, action } = actionFromFixture(fixture);
      return [key, action];
    }));
    return actionMap;
```
with `import fromEntries from 'object.fromentries';`, but i understand if you prefer this approach to adding the dep, despite being very useful and clean :-)

packages/test/src/MockResolver.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, [fixtures]);
+
+  const dispatch = useContext(DispatchContext);
+  const dispatchInterceptor = useCallback(
+    (action: ActionTypes) => {
+      if (action.type === actionTypes.FETCH_TYPE) {
+        const { key, resolve, reject } = action.meta;
+        if (key in fetchToReceiveAction) {
```
ljharb, line 38: what happens if the key is `'toString'`?
```suggestion
        if (Object.prototype.hasOwnProperty.call(fetchToReceiveAction, key)) {
```

packages/test/src/MockResolver.tsx
```diff
@@ -0,0 +1,85 @@
...
+  {
+  request: ArticleResource.list(),
+  params: { maxResults: 10 },
+  result: [],
+  }`,
+        );
+        return dispatch(action);
+      } else {
+        return dispatch(action);
+      }
```
ljharb, line 75:
```suggestion
      }
      return dispatch(action);
```

ljharb (approved):
LGTM overall

packages/test/src/MockResolver.tsx
ntucker: ya keeping the library small is a primary objective.

packages/test/src/MockResolver.tsx
ntucker: key is typed is always be an actual string - so in practice this is less bytes to send over wire (and easier to read).

ljharb: right, but if it's a string that's one of the keys on `Object.prototype`, like `toString`, then this will break.

PR #447, opened 2020-12-30
enhance: Remove FixtureManager in favor of MockResolver

ntucker:
### Motivation

#446 is the favored approach since it makes changing fixtures in a component easy

### Solution

Remove FixtureManager (use MockResolver instead)


ntucker:
This is forked from the other branch so will rebase on merge

ljharb:
This is labeled "enhance", but it seems like it's a breaking change?

ntucker:
Breaking is marked by 'BREAKING CHANGE' in commit message. However - since the removed item was never released in a version (even beta), this actually breaks nothing.

ljharb:
Ah, ok great, thanks for clarifying.

ljharb (approved):

PR #457, opened 2021-01-05
internal: Add lerna options to  cmd

ntucker:
### Motivation

To avoid committing directly to master, we need to have a release process that only pushes branches.

### Solution

New release process:

1) `git checkout -b release`
2) `lerna version` (this will commit to the current branch and push tags, etc, but not publish)
3) open PR on github for version branch & land PR
4) publish: `yarn clean && yarn build && lerna publish from-git`

Due to this, we are adding the options to lerna version in the config.

### Future work

Automate step 4

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -10,7 +10,15 @@
       "preDistTag": "beta",
       "changelogPreset": "anansi",
       "createRelease": "github",
-      "message": "internal: publish",
+      "message": "internal: publish %s",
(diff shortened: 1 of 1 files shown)


ljharb:
Want to make the release PR? :-) then i can land it and run step 4 (are there other steps around releasing the website?)

ntucker:
Lerna version still creates tags and releases so it still must be done on main repo even if not main branch. Thus, it requires write access

PR #461, opened 2021-01-10
internal: Fix publish message

ntucker:
### Motivation

`%s` only works when all packages match versions


packages/core/src/react-integration/hooks/useResource.ts
```diff
@@ -2,7 +2,9 @@ import { ReadShape, ParamsFromShape } from '@rest-hooks/core/endpoint';
 import { Denormalize, DenormalizeNullable } from '@rest-hooks/endpoint';
 import { useDenormalized } from '@rest-hooks/core/state/selectors';
 import { StateContext } from '@rest-hooks/core/react-integration/context';
-import { useMemo, useContext } from 'react';
+import { useMemo, useContext, useEffect } from 'react';
```
ntucker, line 5: These were just extraneous imports

ljharb (approved):

PR #464, opened 2021-01-14
enhance: Inferred endpoints expiry based on entities

### LINKED ISSUE #460: Tracking entity freshness and avoiding additional requests
**React version** 16.13.1

**Concurrent mode** no

**Rest Hooks version** 4.5.9

**Describe the bug**
I have a list component that fetches a list of entities.
When clicking on entity, I navigate to that entity page and fetch that single entity by id.

I noticed there's a network request for the single entity, even if it's already in cache.

**To Reproduce**
1. Call to `useResource` to fetch a list of entities `listShape`.
2. Call to `useResoruce` to fetch a single entity that already loaded in `listShape`.
3. A request is made to `GET /resource/:id`.

**Expected behavior**
I'm expecting that no request will be made to fetch entity by id since it's already in cache.

**Additional context**
In my current API, the shape of each entity in the list is exactly the same as in the single entity.
But it may be that other APIs return partial entity data in the list request and full entity in the single entity request so we may need some way to signal if entity is partial and re-fetch even if in cache.

ntucker:
In other words: don't extraneously refetch fresh data.

BREAKING CHANGE: useResource() inferred endpoint will sometimes
not trigger a fetch if entities are fresh enough

Fixes #460 .
Also requested by: @melissafzhang @dgcoffman 
Continuation of: #439

### Motivation

Allows liberal use of useResource()

- When a detail (inferrable) useResouce() is in a nested component of a list result, will not trigger `N` fetches
- Any other case where manual useCache() tracking was used.

### Solution

- Previously inferred results were set to expiry time of 0, which meant they would try to fetch if they could
- Now this is computed based on the 'oldest' (or soonest to expire) entity of those used.
- If any entities are missing, the full result was not inferred and thus expiry is 0.
- Also went back to previous denormalize() design (from previous commit) that returns the localCache of entities. However, this time we have a slightly cleaner interface to getUnvisit().
- entitiesExpireAt added to several hooks `useRetrieve()`, `useExpiresAt()`. This defaults to 0 so previous behavior is maintained.
- entitiesExpireAt is now returned from `useDenormalized()`

### DIFF
--- a/packages/core/src/react-integration/__tests__/useResource-endpoint.web.tsx
+++ b/packages/core/src/react-integration/__tests__/useResource-endpoint.web.tsx
@@ -254,10 +254,18 @@ describe('useResource()', () => {
   });
   it('should NOT suspend even when result is stale and options.invalidIfStale is false', () => {
     const { entities, result } = normalize(payload, CoolerArticleResource);
+    const entityMeta: any = {};
+    for (const k in entities) {
+      entityMeta[k] = {};
+      for (const pk in entities[k]) {
+        entityMeta[k][pk] = { date: 0, expiresAt: 0 };
+      }
+    }
     const fetchKey = CoolerArticleResource.detail().getFetchKey(payload);
     const state = {
       ...initialState,
       entities,
+      entityMeta,
       results: {
         [fetchKey]: result,
       },
@@ -292,12 +300,20 @@ describe('useResource()', () => {
     const fetchKey = InvalidIfStaleArticleResource.detail().getFetchKey(
       payload,
     );
+    const entityMeta: any = {};
+    for (const k in entities) {
+      entityMeta[k] = {};
+      for (const pk in entities[k]) {
+        entityMeta[k][pk] = { date: 0, expiresAt: 0 };
+      }
+    }
     const state = {
       ...initialState,
       entities,
       results: {
         [fetchKey]: result,
       },
+      entityMeta,
       meta: {
         [fetchKey]: {
           date: Infinity,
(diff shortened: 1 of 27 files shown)


packages/legacy/package.json
```diff
@@ -59,7 +59,7 @@
     "@rest-hooks/test": "^2.0.0-k.2"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.0.0-delta.0",
+    "@rest-hooks/core": "^1.0.0-k.6",
```
ntucker, line 62: since we changed some internals

packages/core/src/react-integration/__tests__/useResource-endpoint.web.tsx
```diff
@@ -327,12 +343,20 @@ describe('useResource()', () => {
     const fetchKey = InvalidIfStaleArticleResource.detail().getFetchKey(
       payload,
     );
+    const entityMeta: any = {};
+    for (const k in entities) {
+      entityMeta[k] = {};
+      for (const pk in entities[k]) {
+        entityMeta[k][pk] = { date: 0, expiresAt: 0 };
```
ljharb, line 350: perhaps this repeated logic could be factored into a shared function?

packages/core/src/state/selectors/useDenormalized.ts
```diff
@@ -74,31 +78,47 @@ export default function useDenormalized<
...
+
+    // oldest entity dictates age
+    let expiresAt = Infinity;
+    if (found) {
+      // using Object.keys ensures we don't hit `toString` type members
+      for (const key of Object.keys(resolvedEntities)) {
+        for (const pk of Object.keys(resolvedEntities[key])) {
+          expiresAt = Math.min(expiresAt, entityMeta?.[key]?.[pk].expiresAt);
+        }
+      }
```
ljharb, line 107: you'll get a bit more perf in some engines by using `.forEach` here rather than for..of loops, which invokes the iterator protocol.

ljharb (approved):

packages/core/src/react-integration/__tests__/useResource-endpoint.web.tsx
ntucker: oh ya, i did that in another test file. i'm so liberal with copypasta in tests :P

packages/core/src/state/selectors/useDenormalized.ts
ntucker: TIL

ntucker: This is the total opposite of python

PR #471, opened 2021-01-15
docs: Update nested response docs with behaviors of rest package

ntucker:
### Motivation

With #469, the recommended usage includes nested entities in response.

### Solution

Reflect new behavior in doc guide for nested resources.

### DIFF
--- a/packages/rest/src/__tests__/resource-endpoint.ts
+++ b/packages/rest/src/__tests__/resource-endpoint.ts
@@ -39,17 +39,19 @@ describe('Resource', () => {
   });
 
   it('should init', () => {
+    const author = UserResource.fromJS({ id: 5 });
     const resource = CoolerArticleResource.fromJS({
       id: 5,
       title: 'happy',
-      author: 5,
+      author,
     });
     expect(resource.pk()).toBe('5');
     expect(CoolerArticleResource.pk(resource)).toBe('5');
     expect(resource.title).toBe('happy');
     expect(resource.things).toBe('happy five');
     expect(resource.url).toBe('http://test.com/article-cooler/5');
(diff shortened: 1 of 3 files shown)


packages/rest/src/__tests__/resource-endpoint.ts
```diff
@@ -39,17 +39,19 @@ describe('Resource', () => {
...
+      author,
     });
     expect(resource.pk()).toBe('5');
     expect(CoolerArticleResource.pk(resource)).toBe('5');
     expect(resource.title).toBe('happy');
     expect(resource.things).toBe('happy five');
     expect(resource.url).toBe('http://test.com/article-cooler/5');
-    expect(resource.author).toBe(5);
+    expect(resource.author).toBe(author);
+    expect(resource.author?.pk()).toBe('5');
```
ljharb, line 54: why the optional chain?

ljharb (approved):

packages/rest/src/__tests__/resource-endpoint.ts
ntucker: In this particular case, user is not always defined as defined `readonly author: UserResource | null = null;`  in ` __tests__/new.ts `

Nested members always existing is also supported.

ljharb: ah, the previous assertion doesn’t type narrow, got it

ntucker: ya for some reason expect() doesn't integrate with typescript. if it threw, i think since 3.7 of typescript it would actually help, but it's not marked that way

PR #474, opened 2021-01-17
internal: Bump commitlint

ntucker:
https://github.com/conventional-changelog/commitlint/blob/master/CHANGELOG.md drops node <10 and lerna <3. Added engine in root package (not published but used for dev) that we need node 10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
   },
+  "engines": {"node": ">=0.12"},
   "browserslist": [
     "extends @anansi/browserslist-config"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
     }
   },
+  "engines": {"node": ">=0.12"},
```
ljharb, line 26: did you mean this to be `>= 10`?

ljharb (approved):

package.json
ntucker: oops! yes. copy/paste issue

PR #476, opened 2021-01-18
internal: Use github action to enforce commitlint

ntucker:
### Motivation

Currently we lint on commit, which is a fairly large amount of friction for contributions. What we really need is to simply enforce conventions for a PR merge.

### Solution

Adding this GH action. If this works well, will remove the per-commit enforcement

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,12 @@
+name: Lint Commit Messages
+
+on: [pull_request]
+
+jobs:
+  commitlint:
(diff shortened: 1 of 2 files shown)


.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,14 @@
+name: Lint Commit Messages
+on:
+  push:
+    branches:
+      - master
```
ljharb, line 5: why not just run it on every branch?

ljharb (approved):
much better than the wildly-insecure husky

PR #479, opened 2021-01-18
enhance: Widen RestFetch types to make overriding not break

ntucker:
### Motivation

Ensure override Resource patterns work in all cases.

### Solution

- RestFetch params: object -> any
  - object no longer widens, so overrides break with `{ id: number }` or the like (even though this matches, due to the directionality)
- useFetcher matches 'any' properly

Also:
- updated tests & docs with all known cases to ensure our recommendations work

### DIFF
--- a/__tests__/new.ts
+++ b/__tests__/new.ts
@@ -4,6 +4,8 @@ import {
   EndpointExtraOptions,
   FetchFunction,
   Index,
+  SchemaDetail,
+  SchemaList,
 } from '@rest-hooks/endpoint';
 import {
   Resource,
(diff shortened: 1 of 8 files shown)


__tests__/new.ts
```diff
@@ -151,6 +153,46 @@ export class CoolerArticleResource extends ArticleResource {
   }
 }
 
+export class TypedArticleResource extends CoolerArticleResource {
+  get tagString() {
```
ljharb, line 157: why make this a getter? that adds a lot of indirection and a performance hit, and only saves you typing `()` on the consumer side.

ljharb (approved):
lots of any's :-(

__tests__/new.ts
ntucker: It's a test - eventually I want to filter out getters in my type definitions. I don't think there's a way to currently do it.

ljharb: What for? a getter, an own property, and a method all should be showing up in type definitions.

ntucker: when i send a payload to update a Resource, I do not expect it to include methods.

```js
//good
update({ id }, {title: 'hi'});

// bad
update({id}, {title: 'hi', tagString: '123'});
```

ljharb: Ok - why not make tagString a normal method then? getters are generally to be avoided.

ntucker: I'm not sure how I could ban users from using getters.

ljharb: lol no, of course you can’t, but why would the use them in the first place?

Or by “experiment” do you mean you’re trying to make it ergonomic for them to do so?

ntucker: We try not to have opinions about things that aren't relevant to the principals of the library. If their types are wrong because they choose to use a getter, this would be a fault of the library.

PR #487, opened 2021-01-21
enhance: Add jsdocs deprecations to deprecated methods

ntucker:
Integration into VS code came after these were deprecated. This should give nice indications in IDEs that support.

### DIFF
--- a/packages/normalizr/src/entities/SimpleRecord.ts
+++ b/packages/normalizr/src/entities/SimpleRecord.ts
@@ -145,6 +145,7 @@ export default abstract class SimpleRecord {
     return [this.fromJS(object) as any, found, deleted];
   }
 
+  /** @deprecated */
   /* istanbul ignore next */
   static asSchema<T extends typeof SimpleRecord>(this: T) {
     /* istanbul ignore next */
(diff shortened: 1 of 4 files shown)


ljharb (approved):

yunyu:
@ntucker I'm mulling over an upgrade to v5, and don't see the getEndpointExtra exported in the SimpleResource that's under the rest-hooks package. Is this intentional?

ntucker:
@yunyu `getEndpointExtra()` and endpoints in general are the new way of doing things, and thus are in the `@rest-hooks/rest` package. To make upgrading easy, the `Resource` in `rest-hooks` will remain as a legacy implementation. You can see the list of changes for the `rest-hooks` resource in https://github.com/coinbase/rest-hooks/releases/tag/rest-hooks%405.0.0 (just some removals of stuff previously deprecated)

My plan is to write a migration doc this coming weekend, to not only outline getting on 5.0 but also how to easily move to @rest-hooks/rest - which is the future. Generally I'd recommend just getting on 5.0 first, and then moving that piece.

Just off the top of my head, some of the large breaking changes, but improvements in `@rest-hooks/rest`
- New methods that use the new `Endpoint`
- Streamlined fetch method directly reflecting API of browser fetch
- Supports nested entities

PR #490, opened 2021-01-22
fix: MockResolver intercepts subscriptions matching fixtures

ntucker:
### Motivation

Subscriptions cause fetches - which means components using subscriptions that are mocked will still cause a normal fetch

### Solution

Since fixtures currently don't support any time-series based changes, simply blocked the mocked subscription dispatches will produce desired results.


ljharb (approved):
LGTM but should probably have a regression test?

PR #497, opened 2021-01-24
pkg: react-hooks-testing-library major

ntucker:
BREAKING CHANGE:
- result.current, result.error is now `undefined` after suspense, rather than `null`
-  interval will now default to 50ms in async utils
-  timeout will now default to 1000ms in async utils
-  suppressErrors has been removed from async utils
- Adjust types so that react renderer exports don't required extra generic parameter
- Importing from renderHook and act from @testing-library/react-hooks will now auto-detect which renderer to used based on the project's dependencies
    - peerDependencies are now optional to support different dependencies being required
    - This means there will be no warning if the dependency is not installed at all, but it will still warn if an incompatible version is installed
    - Auto-detection won't work with bundlers (e.g. Webpack). Please use as specific renderer import instead
(see https://github.com/testing-library/react-hooks-testing-library/releases/tag/v5.0.0)

### Motivation

Keep up with the latest

### Solution

Upgrade react-hooks-testing-library from 3.2 to 5
Had to change all the checks for suspense to undefined from null.
Deprecated `cleanup()` as it already happens on unmount.
Added cleanup() to externalprovider in tests to reflect behavior of cacheprovider.
Updated redux docs to include calling manager's init

Waiting on https://github.com/testing-library/react-hooks-testing-library/issues/554 to be able to use this version.

### DIFF
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -46,7 +46,7 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "dependencies": {
-    "@testing-library/react-hooks": "~3.5.0"
+    "@testing-library/react-hooks": "~5.0.3"
   },
   "peerDependencies": {
     "@rest-hooks/core": "^1.0.0-beta.1",
--- a/docs/api/makeRenderRestHook.md
+++ b/docs/api/makeRenderRestHook.md
@@ -119,7 +119,7 @@ it('should resolve useResource()', async () => {
   const { result, waitForNextUpdate } = renderRestHook(() => {
     return useResource(ArticleResource.detail(), payload);
   });
-  expect(result.current).toBe(null);
+  expect(result.current).toBeUndefined();
   await waitForNextUpdate();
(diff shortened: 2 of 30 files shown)


ntucker:
Might have to consider forking or discontinuing use of this library as it doesn't seem to be maintained well.

packages/test/package.json
```diff
@@ -46,7 +46,7 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "dependencies": {
-    "@testing-library/react-hooks": "~3.5.0"
+    "@testing-library/react-hooks": "~5.0.3"
```
ljharb, line 49: why ~ ?

ljharb (approved):

packages/test/package.json
ntucker: As outlined in a previous PR - there was an uncaught breaking change in a minor upgrade to this library. This lets us validate all minor upgrades before our users potentially get broken things.

ljharb: oof, gross

PR #498, opened 2021-01-24
fix: Disallow breaking 3.6 version of @testing-library/react-hooks

ntucker:
### Motivation

3.6 accidentally introduced a breaking change. More here: https://github.com/testing-library/react-hooks-testing-library/issues/554

### Solution

Disallow major versions beyond 3.5.

### DIFF
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -46,7 +46,7 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "dependencies": {
-    "@testing-library/react-hooks": "^3.2.1"
+    "@testing-library/react-hooks": "~3.5.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


ljharb (approved):
lol, this library (－‸ლ)

mpeyper:
> lol, this library (－‸ლ)

@ljharb, I don't understand what this means. Can you please elaborate?

If it's negative, I'd much rather take the feedback directly.

ljharb:
@mpeyper nothing that can really be fixed; RTL's approach means it can't test server-rendering, using test IDs is a brittle approach (and has been for well over a decade), and "simulating events" doesn't actually do so, so the main premises of the entire library are what I consider flawed. I'm not interested in debating it in a public forum, however, so i'm going to hide both of our comments.

PR #504, opened 2021-01-28
enhance: Expand matches peerdeps for endpoint

ntucker:
### Motivation

Allow matching of v1. Noticed this when using npm instead of yarn and it tried to resolve with v0.6. Realized v1 is breaking from 0.x

### Solution

Add a union for peerDep of endpoint.


ljharb (approved):
indeed, even v0.7 is breaking from v0.6.x

ntucker:
Circle says everything's passing but we have `build Expected — Waiting for status to be reported ` 🤔

PR #505, opened 2021-01-28
fix: Ensure Endpoint doesn't trigger CSP

ntucker:
### Motivation

[CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) disallows calling Function(), which means in some scenarios the Endpoint class would not work.

### Solution

Using `Object.setPrototypeOf(self, new.target.prototype);` we can still make instances inherit from Function without using `this` and thus, being forced to call `super()`, which is disallowed.

This does mean we cannot use class assignments, so moved those to constructor. This should work on children though.

A benefit of this method is we no longer have the hacky eval-like string.

### TODO:

This won't work in IE, so we need a detection mechanism to use the old method

### DIFF
--- a/packages/endpoint/src/endpoint.js
+++ b/packages/endpoint/src/endpoint.js
@@ -15,11 +15,17 @@ function runCompat(endpoint, options) {
 
 export default class Endpoint extends Function {
   constructor(fetchFunction, options) {
-    super('return arguments.callee.fetch.apply(arguments.callee, arguments)');
-    if (fetchFunction) this.fetch = fetchFunction;
-    Object.assign(this, options);
+    const self = (...args) => self.fetch(...args);
     /** The following is for compatibility with FetchShape */
-    runCompat(this, options);
+    self.getFetchKey = params => self.key(params);
+
+    if (fetchFunction) self.fetch = fetchFunction;
+    Object.assign(self, options);
+    Object.setPrototypeOf(self, new.target.prototype);
+
+    /** The following is for compatibility with FetchShape */
+    runCompat(self, options);
+    return self;
   }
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):

packages/endpoint/src/endpoint.js
```diff
@@ -13,13 +13,32 @@ function runCompat(endpoint, options) {
...
+  CSP = true;
+}
+
 export default class Endpoint extends Function {
   constructor(fetchFunction, options) {
-    super('return arguments.callee.fetch.apply(arguments.callee, arguments)');
-    if (fetchFunction) this.fetch = fetchFunction;
-    Object.assign(this, options);
+    let self;
+    if (CSP) {
```
ljharb, line 26: why go with the complex approach instead of doing the same non-eval thing for everyone?

ljharb: ah, i see you say it won’t work in IE - but neither should extends Function work there :-/

ntucker: It's transpiled. Does es5 inheritance work with function?

ljharb: no. Extending builtins requires native class syntax.

ntucker: oh hmm....so there's really no way to make this work in IE then?

ljharb: not that i know of.

praxxis:
As another data point, this code/pattern caused us to avoid using a particular testing library https://github.com/DevExpress/testcafe-hammerhead/issues/2439

PR #506, opened 2021-01-28
docs: Added @rest-hooks/rest package to installation guide

ranyefet:
### Motivation
Current docs does not mention that you need to install `@rest-hooks/rest` to use the library.


ntucker (approved):
Thanks!

ntucker:
Can you add `docs: ` to the front of the commitmessages? This is used for autogenerating changelogs

PR #507, opened 2021-01-28
Fixed code example of extending detail endpoint

ranyefet:
In custom endpoints docs, the code example is wrong:

```js
export default class CommentResource extends Resource {
  static detail<T extends typeof Resource>(this: T) {
    return super.detail.extend({
      schema: { data: this },
    });
  }
```

Updated **.detail** to **.detail()**

```js
export default class CommentResource extends Resource {
  static detail<T extends typeof Resource>(this: T) {
    return super.detail().extend({
      schema: { data: this },
    });
  }
```

### DIFF
--- a/docs/guides/extending-endpoints.md
+++ b/docs/guides/extending-endpoints.md
@@ -136,7 +136,7 @@ import { Resource } from '@rest-hooks/rest';
 
 export default class CommentResource extends Resource {
   static detail<T extends typeof Resource>(this: T) {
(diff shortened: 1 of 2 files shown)


ntucker:
Thanks for the fix! Can you also apply this to the website/versioned_docs folder for v5?

ranyefet:
Done

ntucker (approved):
Thanks!

PR #508, opened 2021-01-30
internal: publish

ntucker:
- @rest-hooks/core@1.0.2
 - @rest-hooks/endpoint@1.0.2
 - rest-hooks@5.0.2
 - @rest-hooks/rest@1.0.1

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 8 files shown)


ljharb (approved):
why bother updating the ones that don’t have any changes but deps?

ntucker:
Then you ensure you get the latest version of their transitive dependencies.

ljharb:
won’t that already happen when someone runs the update command, or reinstalls the base package? that’s kind of the whole point of semver ranges.

ntucker:
- This is a lerna behavior, so you might ask the babel people to get their idea
- For me, running update is scary - I like to move things incrementally, and it's also easier to know that I can update something when the package I actually have installed changes. I wouldn't know to upgrade otherwise unless I did the scary `outdated` commands which show way too much stuff that I don't even understand half the time.

ljharb:
Things like renovate and dependabot usually handle that kind of updating, but understood.

PR #513, opened 2021-02-03
pkg: Relax @babel/runtime requirement to ^7.7.2

ntucker:
### Motivation

Having multiple @babel/runtimes with nesting creates some interesting issues when building. 

After a more thorough investigation of https://github.com/babel/babel/pull/10853 I have realized that 7.12 is simply the requirement to work with webpack 5, but interop with babel/core will work. This is due to the new package exports directives - which are processed by webpack 5 only.

Thus, users of webpack 5, should ensure they run only babel 7.12+ and also hoist it to ensure a single version. Other users should not care.

### Solution

To reduce the chance of non-hoisted runtimes, we relax our requirements to what 4.5.9 was. This should enable no new introduced of non-hoisted runtimes.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -66,7 +66,7 @@
     "@rest-hooks/test": "^3.0.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.12.0",
+    "@babel/runtime": "^7.7.2",
     "@rest-hooks/endpoint": "^1.0.2",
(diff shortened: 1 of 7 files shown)


ljharb (approved):
not sure why this would be an issue - nothing in the entire ecosystem should be depending on babel/runtime without a `^` - but seems fine

PR #533, opened 2021-02-09
internal(pkg): Remove husky as it is now unused

ntucker:


ljharb:
husky is very dangerous, so removing it is def the best option.

ntucker:
I think it's used by the commitlint which still needs to run in the github action

PR #540, opened 2021-02-15
enhance: legacy Resource should only warn on validation problems

ntucker:
### Motivation

Since `Resource`s in 'rest-hooks' package (vs `@rest-hooks/rest`) are only for legacy, ease upgrade path.

It's still very unlikely to cause problems, but this should make things absolutely easy.

### Solution

`warn` still so people are aware, but if they were using it before it's likely working correctly.

### Open Questions

In the case of building new Resources off 'rest-hooks' import, it won't bring as many benefits; but more important to get everyone on 5 smoothly.

### DIFF
--- a/__tests__/legacy.ts
+++ b/__tests__/legacy.ts
@@ -1,0 +1,66 @@
+import { Resource } from 'rest-hooks';
+
+export class UserResource extends Resource {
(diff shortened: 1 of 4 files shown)


ljharb (approved):
seems innocuous, but is there a test for it?

PR #552, opened 2021-02-21
fix: Support non-JSON payloads (like FormData)

### LINKED ISSUE #550: Content-Type: application/json is for multipart form uploads

17.0.1

**Concurrent mode** yes/no

no

5.0.4

**Describe the bug**

Incorrect Content-Type header when PUT'ing multipart form data - [looks to be hard-coded here](https://github.com/coinbase/rest-hooks/blob/28b79aea6d280aa74abeea1d4d2ab14678edf9de/packages/rest-hooks/src/resource/Resource.ts#L35).

**To Reproduce**

I defined a custom endpoint for uploading photos.  The only thing I overrode was `url()` to something like `/users/${id}/photo`.  I used:
```javascript
const addPhoto = useFetcher(User.addPhoto())
const form = new FormData()
form.append('photo', newPhoto)
await addPhoto({id}, form)
```

**Expected behavior**

Some options:

1. Don't specify Content-Type and let `fetch()` auto-detect.
2. Specify based on the payload.
3. Provide a way for users to specify in custom endpoints or when calling the fetcher.

I suspect option 3 is already available to us but I couldn't figure it out.  I know how to modify fetch headers per Resource but not per Endpoint or per fetch.

**Additional context**

The app server (expressjs w/multer for multipart handling) complains with "PayloadTooLargeError: request entity too large".  I'm using FormData for the payload.  When using rest-hooks, I see `Content-Type: application/json` in the request headers.  When using native fetch, I see `Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryK3OGAXCjDS0o4nMX` which works.  I did not have to specify Content-Type - it was auto-detected correctly.

Thank you!

ntucker:
Fixes #550  .

### Motivation

For certain payloads this obviously isn't the correct header.

### Solution

- No body, or string body - set default 'Content-Type' header to 'application/json'
- User's init will always override this, which can be specified in many places
- Only for those using `Resource`

### DIFF
--- a/packages/rest-hooks/src/resource/Resource.ts
+++ b/packages/rest-hooks/src/resource/Resource.ts
@@ -33,7 +33,7 @@ export default abstract class Resource extends SimpleResource {
     };
     if (this.fetchOptionsPlugin) options = this.fetchOptionsPlugin(options);
     if (body) options.body = JSON.stringify(body);
-    if (!(body instanceof FormData)) {
+    if (!options.body || typeof options.body === 'string') {
       options.headers = {
         'Content-Type': 'application/json',
         ...options.headers,
--- a/packages/rest-hooks/src/resource/Resource.ts
+++ b/packages/rest-hooks/src/resource/Resource.ts
@@ -30,13 +30,15 @@ export default abstract class Resource extends SimpleResource {
   ) {
     let options: RequestInit = {
       method: method.toUpperCase(),
-      headers: {
-        'Content-Type': 'application/json',
-        // "Content-Type": "application/x-www-form-urlencoded",  -- maybe use this if typeof body is FormData ?
-      },
     };
     if (this.fetchOptionsPlugin) options = this.fetchOptionsPlugin(options);
     if (body) options.body = JSON.stringify(body);
+    if (!(body instanceof FormData)) {
+      options.headers = {
+        'Content-Type': 'application/json',
+        ...options.headers,
+      };
+    }
     return fetch(url, options)
       .then(response => {
         if (!response.ok) {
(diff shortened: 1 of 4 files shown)


packages/rest-hooks/src/resource/Resource.ts
```diff
@@ -30,13 +30,15 @@ export default abstract class Resource extends SimpleResource {
   ) {
     let options: RequestInit = {
       method: method.toUpperCase(),
-      headers: {
-        'Content-Type': 'application/json',
-        // "Content-Type": "application/x-www-form-urlencoded",  -- maybe use this if typeof body is FormData ?
-      },
     };
     if (this.fetchOptionsPlugin) options = this.fetchOptionsPlugin(options);
     if (body) options.body = JSON.stringify(body);
```
ljharb, line 35: what happens if i provide both `body` and `options.body`?

ljharb (approved):

packages/rest-hooks/src/resource/Resource.ts
ntucker: body will override that. this is the legacy way of doing things. @rest-hooks/rest maps much cleaner to normal 'fetch' interface patterns, and has many more extension points to specify user-defined behavior.

i agree it's not obvious how this will work, thus the change in design.

as for why it was done this way - body is called on a case-by-case basis, whereas fetchOptionsPlugin() is defined for endpoint statically. So body takes precedent due to being defined in a case with more dynamic information.

e.g., every call to `fetch(params, body)` can specify an exact body, given a fetch, which is defined at same time as `fetchOptionsPlugin()`

PR #571, opened 2021-02-24
enhance: Improve error type preciseness

ntucker:
### Motivation

For those using built in fetch, we can be assured the error might show up as NetworkError. If we expose this in a union at the very least they can test for it

### Solution

- Establish central NetworkError definition
- Spread through all error code
- Update docs

### Open Questions

Maybe we should have a way of allowing users to customize what their error types are?

### DIFF
--- a/packages/legacy/package.json
+++ b/packages/legacy/package.json
@@ -56,7 +56,7 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.0.0-k.6",
+    "@rest-hooks/core": "^1.0.4",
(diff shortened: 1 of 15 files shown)


ljharb (approved):

packages/legacy/package.json
```diff
@@ -56,7 +56,7 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.0.0-k.6",
+    "@rest-hooks/core": "^1.0.4",
```
ntucker, line 59: we use the new type export implicitly in the return type

PR #576, opened 2021-02-24
enhance: Explicitly defined Resource endpoint return types

ntucker:
### Motivation

- Reduce complexity of return types based on inference
  - This makes the types more readable to user
  - Also can help typescript performance - reducing chance of triggering recursion depth limit (I have theory that https://github.com/coinbase/rest-hooks/issues/569 was caused by this)
- Shows example in code of how to explicitly type return values of endpoints

### Solution

Add `this` definition to `RestFetch`

### DIFF
--- a/docs/api/types.md
+++ b/docs/api/types.md
@@ -160,8 +160,8 @@ as [Resource](./resource) expects it
 
 ```typescript
 export type RestFetch<
-  P = object,
+  P = any,
   B = RequestInit['body'] | Record<string, any>,
   R = any
-> = (params: P, body?: B) => Promise<R>;
+> = (this: RestEndpoint, params: P, body?: B) => Promise<R>;
 ```
--- a/packages/rest/src/SimpleResource.ts
+++ b/packages/rest/src/SimpleResource.ts
@@ -1,5 +1,5 @@
 import { Entity, schema } from '@rest-hooks/normalizr';
-import type { AbstractInstanceType } from '@rest-hooks/normalizr';
+import type { AbstractInstanceType, Schema } from '@rest-hooks/normalizr';
 import { Endpoint } from '@rest-hooks/endpoint';
 import type {
   EndpointExtraOptions,
@@ -10,6 +10,7 @@ import type {
 import { ReadShape, MutateShape } from './legacy';
 import { NotImplementedError } from './errors';
 import paramsToString from './paramsToString';
+import { RestEndpoint } from './types';
 
 /** Represents an entity to be retrieved from a server.
  * Typically 1:1 with a url endpoint.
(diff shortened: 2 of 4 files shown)


docs/api/types.md
```diff
@@ -160,8 +160,8 @@ as [Resource](./resource) expects it
 
 ```typescript
 export type RestFetch<
-  P = object,
+  P = any,
```
ljharb, line 163: why any?

packages/rest/src/SimpleResource.ts
```diff
@@ -157,11 +162,15 @@ export default abstract class SimpleResource extends Entity {
   }
 
   /** Base endpoint but for sideEffects */
-  protected static endpointMutate() {
+  protected static endpointMutate(): RestEndpoint<
+    (this: RestEndpoint, params: any, body?: any) => Promise<any>,
```
ljharb, line 166: same here - why not a generic that defaults to unknown?
```suggestion
    (this: RestEndpoint, params: T, body?: any) => Promise<T>,
```
(I’m not sure where to put the `<T = unknown>` here)

docs/api/types.md
ntucker: This was already discussed when it first landed. Due to typescripts limitations this is the only way to allow users to specify their own types. The alternative would be not allowing for precise typing. You started a whole issue on typescript  repo because of this

ntucker: https://github.com/coinbase/rest-hooks/pull/479

Again, this is unfortunately one of the things where Flow actually got it better than TypeScript.

Stated goals are a [Resource definition works without defining endpoints](https://resthooks.io/docs/getting-started/usage):

```typescript
import { Resource } from '@rest-hooks/rest';

export default class ArticleResource extends Resource {
  readonly id: number | undefined = undefined;
  readonly title: string = '';
  readonly author: number | null = null;
  readonly tags: string[] = [];

  pk() {
    return this.id?.toString();
  }

  static urlRoot = 'http://test.com/article/';
}
```

```typescript
ArticleResource.list()
ArticleResource.detail()
//....
```

And secondly, [overrides with precise types](https://resthooks.io/docs/guides/rest-types) are possible:

```typescript
class MyResource extends Resource {
  static detail<T extends typeof Resource>(
    this: T,
  ): RestEndpoint<RestFetch<{ id: string }>, T, undefined> {
    return super.list();
  }
}
```

ntucker: To clarify, this line is only a docs update to correctly match the types as I missed that in the previous PR

packages/rest/src/SimpleResource.ts
ntucker: Another note: We are not changing types here but defining them explicitly.

ljharb: `unknown` definitely does match everything - `any` is just `unknown` with typechecking disabled.

ljharb (approved):
I'm very skeptical that `any` ever needs to exist when `unknown` is available, but none of this is blocking.

ntucker:
In that case `unknown` should allow overrides, but does not allow for usage out of the box.

ljharb:
i guess i'm a bit unclear on that part; it seems like an explicit `any` would *force* users to cast it to get typechecking, which doesn't seem like the desired default inference behavior.

PR #583, opened 2021-02-28
enhance: Support 'undefined' as schema

ntucker:
### Motivation

`undefined` is a sensible symbol to demonstrate lack of processing.

### Solution

This is the canonical way to represent Endpoint schema's that should just 'pass through' the input. However, for backwards compatibility, we have have been setting schema to `null` if it is undefined.

Fully realizing support to undefined schemas is the first step to properly supported non-backcompat endpoints.

I chose not to convert the hooks FetchShapes as that stage of conversion will occur when they accept Endpoints instead of FetchShapes. Instead, all non-FetchShape places with schema are marked as Schema | undefined.

#### Why not just change Schema to include undefined?

There are many places where we want to only process *actual* schemas. It seems like a better naming convention to defined those as `Schema` and other places with the union. Furthermore, most places where we want to accept | undefined will simply be converted to EndpointInterface, which already includes that symbol.

### DIFF
--- a/__tests__/new.ts
+++ b/__tests__/new.ts
@@ -362,7 +362,7 @@ export class NestedArticleResource extends OtherArticleResource {
 }
 
 export const GetPhoto = new Endpoint(
-  async function ({ userId }: { userId: string }) {
+  async function ({ userId }: { userId: string }): Promise<ArrayBuffer> {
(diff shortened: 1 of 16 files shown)


ljharb (approved):
rubberstamp

PR #584, opened 2021-02-28
fix: Gracefully handle primitive entity responses

ntucker:
### Motivation

- We absolutely cannot push things into WeakMap that are not 'object'.

### Solution

- In dev mode, have useful error message (this indicates something is wrong with response, as entities are only supposed to objects (including arrays))
- In prod, simply ignore all primitives (this won't break anything - as we don't need any primitives to ensure cache consistency)

### DIFF
--- a/packages/normalizr/src/denormalize.ts
+++ b/packages/normalizr/src/denormalize.ts
@@ -20,7 +20,7 @@ const unvisitEntity = (
   getEntity: (
     entityOrId: Record<string, any> | string,
     schema: typeof Entity,
-  ) => any,
+  ) => EntityInterface | typeof DELETED,
   localCache: Record<string, Record<string, any>>,
   entityCache: DenormalizeCache['entities'],
 ): [
@@ -175,7 +175,7 @@ type DenormalizeReturn<S extends Schema> =
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export const denormalize = <S extends Schema>(
-  input: any,
+  input: unknown,
   schema: S,
   entities: any,
   entityCache: DenormalizeCache['entities'] = {},
@@ -228,7 +228,17 @@ function withTrackedEntities(
   const wrappedUnvisit = (input: any, schema: any) => {
     const ret: [any, boolean, boolean] = originalUnvisit(input, schema);
     // pass over undefined in key
-    if (ret[0] && schema && isEntity(schema)) globalKey.push(ret[0]);
+    if (ret[0] && schema && isEntity(schema)) {
+      /* istanbul ignore else */
+      if (typeof ret[0] === 'object') {
+        globalKey.push(ret[0]);
+      } else if (process.env.NODE_ENV !== 'production') {
+        throw new Error(
+          `Unexpected primitive found during denormalization\nFound: ${ret[0]}\nExpected entity: ${schema}`,
+        );
+      }
+    }
+
     return ret;
   };
   wrappedUnvisit.og = unvisit;
--- a/packages/endpoint/src/endpoint.js
+++ b/packages/endpoint/src/endpoint.js
@@ -17,12 +17,16 @@ let CSP = false;
 try {
   Function();
 } catch (e) {
+  /* istanbul ignore next */
   CSP = true;
 }
 
 export default class Endpoint extends Function {
   constructor(fetchFunction, options) {
     let self;
+    // TODO: Test the fallback?
+    /* istanbul ignore if */
+    /* istanbul ignore next */
     if (CSP) {
       self = (...args) => self.fetch(...args);
       Object.setPrototypeOf(self, new.target.prototype);
--- a/packages/normalizr/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/normalizr/src/__tests__/__snapshots__/index.test.js.snap
@@ -211,6 +211,18 @@ Array [
 ]
 `;
 
+exports[`denormalize throws when nested entities are primitives 1`] = `
+"Unexpected primitive found during denormalization
+Found: 4
+Expected entity: class Comment extends _IDEntity.default {
+      constructor(...args) {
+        super(...args);
+        this.comment = '';
+      }
+
+    }"
+`;
+
 exports[`normalize can use fully custom entity classes 1`] = `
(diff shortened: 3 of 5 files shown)


packages/normalizr/src/denormalize.ts
```diff
@@ -228,7 +228,17 @@ function withTrackedEntities(
   const wrappedUnvisit = (input: any, schema: any) => {
     const ret: [any, boolean, boolean] = originalUnvisit(input, schema);
     // pass over undefined in key
-    if (ret[0] && schema && isEntity(schema)) globalKey.push(ret[0]);
+    if (ret[0] && schema && isEntity(schema)) {
+      /* istanbul ignore else */
+      if (typeof ret[0] === 'object') {
```
ljharb, line 233:
```suggestion
      if (typeof ret[0] === 'object' || typeof ret[0] === 'function') {
```
alternatively and more robustly, `if (Object(ret[0]) === ret[0])`.

In the future, however, Symbols may be accepted as WeakMap keys, so this check will get a bit more complex.

packages/normalizr/src/denormalize.ts
```diff
@@ -228,7 +228,17 @@ function withTrackedEntities(
...
+        globalKey.push(ret[0]);
+      } else if (process.env.NODE_ENV !== 'production') {
+        throw new Error(
+          `Unexpected primitive found during denormalization\nFound: ${ret[0]}\nExpected entity: ${schema}`,
+        );
```
ljharb, line 238: i would expect it to fail in prod as well. why would it be desired to have code behave differently in dev and prod?

(note that if this code was merely *logging* in dev-only then i'd have no such concern)

packages/normalizr/src/denormalize.ts
ntucker: ret[0] is some value from a fetch response. While we need to ensure it is acceptable to use as a WeakMap key, we don't have to accept every possible type. I don't see purpose to allowing users to return both function or Symbol as they aren't JSON serializable nor some sort of data buffer.

As long as this check has no false positives, it is sufficient for these purposes. False negatives like banning function are quite acceptable.

packages/normalizr/src/denormalize.ts
ntucker: #### Dev vs Prod
The purpose of this check is to notify users of bad network definitions. They might be accidentally running code that gives unexpected results. In the case of a bad *definition*, it is expected to consistently fail such checks. However in prod, the response may change due to dynamic factors - like the server returns something weird. In these cases, because it's completely acceptable to have a primitive as a response value, we don't want to fail anything. It's only if they get a primitive when they don't *expect* to get one (the schema dictates otherwise) - that we want to trigger this case to help them define their schema. So because handling primitives as a response even when the schema doesn't match is okay for our processing, we don't want to break their application because of it.

#### Error vs logging
Since this nature of this warning is due to bad schema definitions - we want to be very loud. This is how we have handled all other mismatches.

All of these schema checks were prompted by a very common case of someone having a typo in their fetch url and then the server sends back garbage data and they spend hours confused as to why it's not working.

ljharb: I understand the arguments here. I don’t understand why you’d want it to *not* fail loudly in prod.

ntucker: The stole purpose is to help developers define correct schemas. In the case where they accidentally didn't mark something as optional, or some other dynamic definition problem, we don't want to break the user's experience when it can work fine. Furthermore, as far as simply logging errors - these checks have a cost in both runtime and bundle so we don't want to ship complex validation code to prod if not necessary.

ntucker: React does similar things like double renders during development mode as well as spamming dev console with thrown errors caught by error boundaries

ljharb: React makes a lot of subpar choices; it's not always best to follow them. (spamming the dev console is fine as that doesn't alter runtime behavior; that's the same reason propTypes don't throw errors in development, for example - which is what this seems analogous to). I'm not suggesting logging errors in production (or including that code in a prod build), i'm saying that the existence or absence of logging code doesn't alter the runtime behavior of a program.

ljharb (approved):
Not going to block on this ofc, but i still strongly believe that runtime-observable behavior (in particular, not counting console interaction) shouldn't be environment-dependent.

packages/normalizr/src/denormalize.ts
ntucker: To be clear, all the errors I throw are caught by myself. I'm using errors as a control flow mechanism - I do not want any errors to propagate to user-land. However, it does change the way results are processed so I can see that analogy. I do agree following another library like React on things should not be done blindly - it is simply an example of the same principal I am conveying here.

The values this is based on
- Handle bad cases as gracefully as possible
- When constraints cannot catch all user-errors - have mechanism to inform them of mistakes that they make
- Practicality beats purity

I can see the case for simply logging in the background. However, in practice what I found was many people missed those (you have to have the dev console open) and got confused. For practicality certain types of scenarios, I found that adoption and learning happened much faster with errors displayed in this manner.

I am also planning on two other strategies to help users with schemas:
- schema playground that lets people write inline and see how it processes data
- OpenAPI codegen so people don't even have to figure out the schemas themselves.

All of these reports were in response to getting feedback for 2 years and figuring out the biggest pain points in learning. This is all subject to new information, but based on what I've seen so far this is the most pragmatic helping point for new users.

ntucker: Let's keep this conversation open though. I'm sure there's a lot more data bout how this is used and I'd love to update practices to help developers as much as possible. I do agree that changing runtime behavior based on environment can be very dangerous. I mostly prefer solutions where the correct way to do something is obvious instead of trying to tell users they made a mistake. In practice this is hard to come to and often requires a bit of time and iteration. If you have any thoughts of making schema design easier, this would be much appreciated as well as I would definitely prefer not to do this :)

ntucker:
I definitely appreciate the feedback, and want to keep this conversation open. This is currently a conscious tradeoff that should always be questioned. If the intuitiveness of schema design increases enough I would love to move away from this mitigation solution.

PR #585, opened 2021-02-28
enhance: Throw useful error when schema key is missing from Entity

ntucker:
### Motivation

`static schema` keys should map to members of the entity - otherwise they do nothing.

### Solution

During development - throw an error in case a schema is missing in response. This should help users with any typos or if they don't understand how `static schema` should work.

### DIFF
--- a/packages/normalizr/src/entities/Entity.ts
+++ b/packages/normalizr/src/entities/Entity.ts
@@ -219,6 +219,20 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
           addEntity,
           visitedEntities,
         );
+      } else if (process.env.NODE_ENV !== 'production') {
+        const error = new Error(
+          `Schema key is missing in Entity
+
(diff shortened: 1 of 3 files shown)


packages/normalizr/src/entities/Entity.ts
```diff
@@ -219,6 +219,20 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
           addEntity,
           visitedEntities,
         );
+      } else if (process.env.NODE_ENV !== 'production') {
+        const error = new Error(
```
ljharb, line 223: same question as on the other PR; throwing an error (as opposed to just logging) in only dev seems like it would not be desired, since that would mean prod and dev actually do different things.

ntucker: Given the other PR, how do you feel about landing this for now but keeping conversation open?

ljharb (approved):

PR #586, opened 2021-02-28
feat: Add @rest-hooks/img

ntucker:
### Motivation

Faster getting start with images

### Solution

`import { Img, getImage } from '@rest-hooks/img';`


packages/img/rollup.config.js
```diff
@@ -0,0 +1,69 @@
...
+process.env.ROOT_PATH_PREFIX = '@rest-hooks/img';
+
+function isExternal(id) {
+  const ret = dependencies.includes(id);
+  if (!ret) {
+    for (const dep of dependencies) {
+      if (id.startsWith(dep)) return true;
+    }
+  }
+  return ret;
```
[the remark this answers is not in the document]
ntucker: Oh nice, we should do this to the other rollups as well

packages/img/src/Img.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { useResource } from '@rest-hooks/core';
+
+import getImage from './getImage';
+
+export default function Img(
+  props: ImgHTMLAttributes<HTMLImageElement>,
+): React.ReactElement {
+  const { src } = props;
+  useResource(getImage, src ? { src } : null);
+  return <img alt={props.alt} {...props} />;
```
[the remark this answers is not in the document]
ntucker: Does this help with GC load? I should probably change how I'm programming elsewhere if this is significant

packages/img/README.md
```diff
@@ -0,0 +1,83 @@
...
+<details open><summary><b>Profile.tsx</b></summary>
+
+```tsx
+import React, { ImgHTMLAttributes } from 'react';
+import { useResource, useRetrieve } from 'rest-hooks';
+import { Img, getImage } from '@rest-hooks/img';
+
+export default function Profile({ username }: { username: string }) {
+  const imageSrc = `/profile_images/${username}}`;
+  useRetrieve(getImage, { src: imageSrc });
```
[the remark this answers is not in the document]
ntucker: Good point - I need to validate how this works in node as well as react-native. This will run on server, so we need to make it a noop if Image isn't available. Is there a good way to test this?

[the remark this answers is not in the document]
ljharb: Setting jest’s environment to “node”, which can be done per-file via config or an override comment.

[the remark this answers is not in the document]
ntucker: okay it's ready

packages/img/src/getImage.ts
```diff
@@ -1,5 +1,9 @@
 import { Endpoint, EndpointInstance } from '@rest-hooks/endpoint';
 
+function isNode() {
+  return process !== undefined && process.release.name === 'node';
```
ljharb, line 4:
```suggestion
  return typeof process !== 'undefined' && process.release.name === 'node';
```
Otherwise a broken bundler (like webpack 5) will cause this to throw a reference error.

packages/img/src/getImage.ts
```diff
@@ -10,6 +14,8 @@ const getImage: EndpointInstance<
     { src }: { src: string },
   ): Promise<string> {
     return new Promise(resolve => {
+      // even if we polyfill, on server we don't want to actually wait to resolve the Image
```
ljharb, line 17: i don’t think the check you want is “is node” - Image only exists in a browser, so i think the check you want is “isBrowser” - namely, does `window` exist

ntucker: ya but if someone somehow polyfills RN or something I"m okay with them using this. VS on server, even if it's polyfilled it would be silly to do.

ljharb: in RN, window exists.

ntucker: didn't realize that - ok

ljharb: I’m not 100% but I’m pretty sure it’s like a browser environment

packages/img/src/getImage.ts
```diff
@@ -0,0 +1,34 @@
...
+  string,
+  undefined
+> = new Endpoint(
+  async function (
+    this: EndpointInstance,
+    { src }: { src: string },
+  ): Promise<string> {
+    return new Promise(resolve => {
+      // even if we polyfill, on server we don't want to actually wait to resolve the Image
+      if (typeof window === 'undefined' || !Image) resolve(src);
```
ljharb, line 14:
```suggestion
      if (typeof window === 'undefined' || typeof Image !== 'function') resolve(src);
```
without typeof it’s not safe to refer to globales that might not be there.

ljharb (approved):

packages/img/src/getImage.ts
ntucker: yeah, I know they don't mock everything and it creates a lot of weird scenarios.

PR #589, opened 2021-02-28
fix: Exclude browserslist config from published package

ntucker:
### Motivation

This breaks tooling that tries to transpile node_modules as it tries to looking for ` @anansi/browserslist-config` and cannot find it

### Solution

Previously, the build configuration for normalizr was separate, but this has changed so it will use the config from the root level package which is not published.


ljharb (approved):
not that it's ever safe or wise to transpile node_modules…

ntucker:
Found it from preact-cli

PR #590, opened 2021-03-01
enhance: Define package exports for modern tooling

ntucker:
BREAKING CHANGE:
- Requires node 10+ (and not node 13 less than 13.7)
- No nested exports - only allowed usage:
  - `import { /* something */ } from '@rest-hooks/test';`
  - `import packageJson from '@rest-hooks/test/package.json';`

### Motivation

- Some imports are used in context of storybook, which makes the node-only intended pieces screw up webpack
  - `Module not found: Error: Can't resolve 'redux' in '/home/ntucker/src/anansi/node_modules/@rest-hooks/test/lib'`
  - `WARNING in /home/ntucker/src/anansi/node_modules/@testing-library/react-hooks/lib/pure.js 53:11-42
Critical dependency: the request of a dependency is an expression`
- Trial run exports to eventually use in other packages

### Solution

References for behavior in node, webpack and typescript (not supported yet)
- https://webpack.js.org/guides/package-exports
- https://nodejs.org/docs/latest-v15.x/api/packages.html
- https://github.com/microsoft/TypeScript/issues/33079

#### Exports
- node specific stuff
  - includes everything
  - has both cjs and es6 modules
  - has its own types to ensure everything
  - targets node 10
  - Use wrapper method to avoid [dual package hazard](https://nodejs.org/docs/latest-v15.x/api/packages.html#packages_approach_1_use_an_es_module_wrapper)
- fallback (modern browser)
  - targets stage 4 at time of publishing (https://webpack.js.org/guides/package-exports/#target-environment)
  - only exports stuff that might be used outside of node run tests (like MockResolver)
  - types only include the exports from here
- Provide package.json for various tools that like reading it

#### main/module/types (legacy definitions)
Generally try to keep in line to existing behavior:
- Include everything (no way to tell if we're in node)
- es6 modules target the default BROWSERSLIST_ENV
- cjs targets node 10

Technically legacy cjs could still be for another environment (super legacy tools?) but node 10 isn't super new so probably about same? Don't even know what tools might do this, but might revisit if it is a problem.

### Tradeoffs - lack of node 13.0-13.6 support

Early versions of node 13 supported exports field, but had differing behavior of how to handle it.

What we get
- exports definition the way it's "meant" going to the future
- Lots of wins for both current tooling like webpack as well as future unanticipated tooling that exploits es6

What we lose
- non-LTS release versions that are very old

Most users that don't upgrade for a while but still actively use a project are likely to at least do a minor version update. Furthermore, if they haven't yet it should not be too burdensome to upgrade (and they should anyway for security reasons). Because of this, it is considered an acceptable sacrificed.


packages/test/package.json
```diff
@@ -3,24 +3,40 @@
   "version": "3.0.1",
   "description": "Testing utilities for Rest Hooks",
   "sideEffects": false,
-  "main": "dist/index.cjs.js",
-  "module": "lib/index.js",
+  "main": "dist/index.cjs",
```
ljharb, line 6: for actual back compat support, you should NOT use `type: module`, and CJS files should be `.js`, and ESM files should only be `.mjs`.

(Since your engines is `>= 10`, you're still supporting many pre-exports, pre-ESM node versions)

packages/test/package.json
```diff
@@ -3,24 +3,40 @@
...
+  "type": "module",
+  "exports": {
+    "node": {
+      "module": "./dist/index.js",
+      "require": "./dist/index.cjs",
+      "types": "./lib/index.d.ts"
+    },
+    "types": "./lib/browser.d.ts",
+    "default": "./lib/browser.js"
+  },
```
ljharb, line 18:
```suggestion
  "exports": {
    ".": [
      {
        "node": {
          "module": "./dist/index.js",
          "require": "./dist/index.cjs",
          "types": "./lib/index.d.ts"
        },
        "types": "./lib/browser.d.ts",
        "default": "./lib/browser.js"
      },
      "./lib/browser.js",
    ]
  }
```
the "default" must be CJS; you need the array-to-string fallback for node v13.0-v13.6.

i have no idea what your intention is here tho - are the entry points "node" and "types" and the main? i'm not familiar with putting a "types" field anywhere inside "exports".

ljharb (changes requested):
Note that adding "exports" is always a breaking change.

packages/test/package.json
ntucker: I didn't seem to see a lot of adoption of mjs extension? Is that what node cares about? Does this include the files in the 'exports' part?

ntucker: Correct me if I'm wrong, but older versions of node don't support 'exports' at all - and thus will read from 'main'?

ntucker: As I understand webpack 4 uses module as it does not support exports. So removing it would change its behavior. Why would removing it help anything?

packages/test/package.json
ntucker: These are conditions - using https://webpack.js.org/guides/package-exports/#reference-syntax as my guide. Is there a node reference for 'exports'? It's hard to find anything on google since exports is also the name of normal es6 module syntax.

According to this: https://webpack.js.org/guides/package-exports/#abbreviation the ".": part isn't necessary

"the "default" must be CJS; you need the array-to-string fallback for node v13.0-v13.6." - i'm really confused by this sentence. is there a doc that explains this fallback? this seems like nonsequitor. also, why would default need to be CJS? do older versions of node secretly support 'exports'?

packages/test/package.json
ljharb: I'm not saying to remove the `module` field, i'm saying to remove `"type": "module"` - which has only one purpose, to change how `.js` is parsed. By default, and properly, it's CJS, but with type module it's treated as ESM.

node always considers `.mjs` to be ESM, because that's the proper extension for it.

packages/test/package.json
ljharb: the `.` part is necessary if you want to add additional entry points, and you need to add `package.json` in here explicitly or some tooling will break.

`default` should match `main`, so that `require()`  of your package in node v12.16 is the same thing as in v12.17.

packages/test/package.json
ntucker: ah, I get it now. I got confused because of which line this message was bound to. Will remove type: module - i didn't fully understand its purpose before

packages/test/package.json
ntucker: ahh, good point about keeping package.json.
does node 12.17 support 'exports' field? I thought this was relatively new so it required 14 or something?

I guess an open question from me - is there a version of node that supports exports, but not conditionals inside? Do you know where docs are for how node processes exports and what conditionals it might look at?

packages/test/package.json
ntucker: I found https://nodejs.org/docs/latest-v15.x/api/packages.html#packages_conditional_exports and it seems to suggest that "type": "module" works well with import having a esm

ntucker:
gonna change node 'import' to avoid https://nodejs.org/docs/latest-v15.x/api/packages.html#packages_approach_1_use_an_es_module_wrapper (even tho in practice there probably won't be issues since there is no context created, we can't assume anything about how users use the code)

ntucker:
"type" - The package type determining whether to load .js files as CommonJS or ES modules. from https://nodejs.org/docs/latest-v15.x/api/packages.html#packages_node_js_package_json_field_definitions

We're using .cjs for commonjs files so everything else is es6.

ljharb:
That isn't correct, though. CJS isn't legacy, and it isn't ever going anywhere. The future is "ESM and CJS, forever" not "ESM".

Either way, if you're using `.cjs` you aren't truly compatible with node >= 10, as the engines field claims.

packages/test/package.json
ljharb: v12.17+, v13.7+, and v14+ support everything.

node v13.0 and v13.1 support "exports" but only the string form; v13.2-v13.6 supports the object form but only the "default" condition.

There are no docs I'm aware of that describe multiple node versions.

ntucker:
According to the node docs here: https://nodejs.org/docs/latest-v15.x/api/packages.html module: types is only for how .js files are interpreted and when using exports the "require" conditional is used to find commonjs.

Can you tell me the scenario where this breaks? I'm not following the claim as I only have the references above for my knowledge on the matter.

ljharb:
The node docs are only for a single version of node. Unfortunately that makes them unhelpful for package authors wishing to support multiple versions of node.

Using the object form only for an entry point (`.` or any other) will cause that entry point to be impossible to require or import in node v13.0 and v13.1. Using the object form without a "default" key *that is CJS* will cause that entry point to be impossible to require in node v13.2-v13.6.

ntucker:
Is there a 13.7? Seems like the early versions of 13 they were figuring out how they wanted it to work. I'd rather just not support non-LTS non-updated versions that do weird stuff and mark it in engines.

ljharb:
That's certainly an option you could choose.

packages/test/package.json
```diff
@@ -3,24 +3,46 @@
...
+        "types": "./lib/browser.d.ts",
+        "default": "./lib/browser.js"
+      },
+      "./lib/browser.js"
+    ],
+    "./package.json": "./package.json"
+  },
+  "type": "module",
+  "engines": {
+    "node": ">=10 <13 || ^13.7 || >=14"
```
ljharb, line 26: if you want to restrict to only node versions that use "exports", you'd want:
```suggestion
    "node": "^12.17 || ^13.7 || >=14"
```

docs/upgrade/upgrading-test-to-four.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Upgrading @rest-hooks/test to 4
+---
+
+## Breaking Changes
+
+- Requires node 10+
```
ljharb, line 7: this isn't accurate anymore i think

packages/test/package.json
ntucker: I still have "main" so I intend to support v10

ljharb: do you intend to have other entry points besides the package itself?

ntucker: does 12.17 do the new behavior and not the 13.1 behavior?

ljharb: yes, v12.17 has the full "exports" field behavior. i'm asking because the choice to use `.cjs` will cause problems if you want to support node older than v12.17, for non-main entrypoints.

ntucker: Not currently - but I'll keep that in mind with other package designs. Thanks for all the knowledge about node version history!

ljharb (approved):
i still think using `.cjs` and `"type": "module"` is a very bad idea.

ntucker:
Would simply removing the "type": "module" entry fix it for v13? We still want to export es6 as the 'default' fallback option

ljharb:
no - a properly backwards compatible package has its "main" and "." `default` set to a CJS `.js` file, and `import` is the only one that's ESM.

ljharb:
For someone that understands ESM, there's no need to ever fall back from "import".

PR #622, opened 2021-03-04
fix(pkg): Fix build scripts on windows

piotr-cz:
### Motivation

Build scripts break on Windows because of two issues:

1.  The `mkdir` command on Windows has different signature than on unix based operating systems (used in `build:copy:ambient`) 

    ```
    $ yarn run build
    yarn run v1.22.5
    $ yarn build:types && lerna run build --stream
    $ yarn build:copy:ambient && ttsc --build
    $ mkdir -p ./packages/normalizr/lib && cp ./packages/normalizr/src/schema.d.ts ./packages/normalizr/lib/schema.d.ts && 
    mkdir -p ./packages/endpoint/lib && cp ./packages/endpoint/src/endpoint.d.ts ./packages/endpoint/lib/endpoint.d.ts
    The syntax of the command is incorrect.
    error Command failed with exit code 1.
    ```

1.  Build script for _@rest-hooks/normalizr_ package breaks on Windows because there is different syntax to set environment variables than on unix based systems

    ```
    $ yarn run build
    yarn run v1.22.5
    ...
    @rest-hooks/normalizr: $ BROWSERSLIST_ENV=legacy NODE_ENV=development rollup -c
    @rest-hooks/normalizr: $ BROWSERSLIST_ENV=legacy NODE_ENV=production rollup -c
    @rest-hooks/normalizr: 'BROWSERSLIST_ENV' is not recognized as an internal or external command,
    @rest-hooks/normalizr: operable program or batch file.
    @rest-hooks/normalizr: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    @rest-hooks/normalizr: error Command failed with exit code 1.
    @rest-hooks/normalizr: 'BROWSERSLIST_ENV' is not recognized as an internal or external command,
    @rest-hooks/normalizr: operable program or batch file.
    @rest-hooks/normalizr: error Command failed with exit code 1.
    @rest-hooks/normalizr: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    @rest-hooks/normalizr: ERROR: "build:js:development" exited with 1.
    @rest-hooks/normalizr: error Command failed with exit code 1.
    @rest-hooks/normalizr: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    @rest-hooks/normalizr: error Command failed with exit code 1.
    @rest-hooks/normalizr: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    lerna ERR! yarn run build exited 1 in '@rest-hooks/normalizr'
    lerna WARN complete Waiting for 6 child processes to exit. CTRL-C to exit immediately.
    error Command failed with exit code 1.
    ````

### Solution

1.  To fix this issue I've prepended [shx](https://www.npmjs.com/package/shx) with `mkdir` and `cp` commands.
    > shx is a wrapper around ShellJS Unix commands, providing an easy solution for simple Unix-like, cross-platform commands in npm package scripts

1.  To fix this issue I've used `cross-env` just as it's used in other scripts.
    Bug was introduced in https://github.com/coinbase/rest-hooks/commit/681299017a5b9cc983a2fccf335b649faf404942#diff-9cfadc28fc53057750390fc236c785551f4829940457a506e1308952584b0a73

### Open questions
none


ntucker:
@ljharb how do you feel about the security implications of `shx`?

ljharb:
I’ve never used it - usually cross-env, mkdirp, and rimraf are sufficient for my cross-platform scripts usage.

ntucker:
@piotr-cz is there cross platform troubles with `cp`? Otherwise, I'd suggest using cross-env, mkdirp, and rimraf  - as I have also used those extensively

piotr-cz:
@ntucker There is no `cp` on Windows, but `copy`

piotr-cz:
For cross-platform alternative to `cp` we'd need another package as `cpy-cli` or `copyfiles` (reference: [Stack Overflow](https://stackoverflow.com/a/38031210/1012616)).

ljharb:
i'd go with https://www.npmjs.com/package/copyfiles, more usage.

piotr-cz:
@ljharb 
but if we decide to add new package, why not `shx`?

ljharb:
¯\\\_(ツ)_/¯ i'm mildly concerned that `shx` is too broad an API surface, but that would work too.

piotr-cz:
Here is a list of commands it provides: https://github.com/shelljs/shelljs/tree/master/src

ntucker:
@piotr-cz how do you feel about using copyfiles? It's past version 1 and well used so I'm less worried about hidden security vulns

piotr-cz:
ok, done

ntucker (approved):

ntucker:
Thanks for the work here!

PR #627, opened 2021-03-06
docs: Improve @rest-hooks/hooks README

ntucker:
- Fix title
- Add explanations

### DIFF
--- a/packages/hooks/README.md
+++ b/packages/hooks/README.md
@@ -16,8 +16,7 @@ Composable hooks for networking data
 
 ### useCancelling()
 
-[Aborts](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) inflight request is the parameters
-change.
+[Aborts](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) inflight request if the parameters change.
 
 ```typescript
 const data = useResource(useCancelling(MyEndpoint, { filter }), { filter });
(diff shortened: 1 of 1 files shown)


packages/hooks/README.md
```diff
@@ -16,19 +16,27 @@ Composable hooks for networking data
 
 ### useCancelling()
 
+[Aborts](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) inflight request is the parameters
+change.
```
ljharb, line 20: nitL: hard wrapping seems weird and increases diff churn
```suggestion
[Aborts](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) inflight request is the parameters change.
```

ljharb (approved):

packages/hooks/README.md
ntucker: What is hard wrapping?

ljharb: adding a newline in the middle of a sentence/paragraph - usually to ensure line lengths are under an arbitrary threshold. Instead, it's best if a sentence/paragraph is as long as it is, allowing the display side to do the line breaking.

ntucker: Oh I should probably change my editor settings for .md files, I usually randomly put endlines cuz it doesn't wordwrap.

PR #628, opened 2021-03-08
fix: Explicit extensions for ESM

ntucker:
### Motivation

Type module expects explicit file extensions when importing.

### Solution

- Remap all internal imports to relative + add .js extension
- Use package prefix for imports like in other packages (i.e., `@rest-hooks/test`)

### DIFF
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -4,7 +4,7 @@
   "description": "Testing utilities for Rest Hooks",
   "sideEffects": false,
   "main": "dist/index.cjs",
-  "module": "legacy/index.js",
+  "module": "legacy/browser.js",
   "types": "lib/index.d.ts",
   "exports": {
     ".": [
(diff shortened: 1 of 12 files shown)


packages/test/package.json
```diff
@@ -21,6 +21,9 @@
     ],
     "./package.json": "./package.json"
   },
+  "imports": {
+    "#internal/*": "./lib/*.js"
+  },
```
ljharb, line 26: this feature requires node v12.19+ and v14.6+ - do the build scripts handle it for the general case?

ntucker: All node uses commonjs bundle. Imports are only for other tools like webpack

packages/test/rollup.config.js
```diff
@@ -11,15 +11,10 @@ const dependencies = Object.keys(pkg.dependencies)
 const extensions = ['.js', '.ts', '.tsx', '.mjs', '.json', '.node'];
 process.env.NODE_ENV = 'production';
 process.env.BROWSERSLIST_ENV = 'node10';
+process.env.RESOLVER_ALIAS = '{"@rest-hooks/test":"./src"}';
```
ntucker, line 14: see - no #internal here

packages/test/package.json
ntucker: See line 12

ljharb: I’m confused; what would other tools use this for?

Marking a thing with “imports” doesn’t make it private; “not putting it in exports” does that.

ntucker: Putting it in imports makes it importable by another module. Right now we pack complains when the exported module tries to import another file

ljharb: hmm, that’s confusing. You’re saying this is how webpack wants you to mark “externals” that are relative files?

ntucker:
```js
// good
import { MockResolver } from '@rest-hooks/test';

// bad
import MockResolver from '@rest-hooks/test/MockResolver';
```

I don't want any exports besides the top level. However, I also don't want everything rolled up in one file or webpack won't be able to tree shake. Imports are needed so the export file can import another module inside - without exposing it to top level exports.

ljharb: Right - but that last part isn’t how node works. You can import anything inside the package you want without putting it in exports or imports if it’s a relative path. It’s only if you’re using the package’s own name (the self-reference feature) that “exports” restricts internal imports and you’d need “imports”. Is there a reason the bundle isn’t using relative paths?

ntucker: It is using relative paths. I guess webpack diverges from nodes spec then?

```export { default as MockResolver } from './MockResolver';```

Webpack doesn't like this tho. Maybe you should file an issue with them?

Comment (ntucker) [reply packages/test/package.json:26]:
```
[webpack-cli] ModuleNotFoundError: Module not found: Error: Can't resolve './MockProvider' in '/home/ntucker/src/anansi/node_modules/@rest-hooks/test/lib'
anansi-example-typescript: Did you mean 'MockProvider.js'?
anansi-example-typescript: BREAKING CHANGE: The request './MockProvider' failed to resolve only because it was resolved as fully specified
anansi-example-typescript: (probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
anansi-example-typescript: The extension in the request is mandatory for it to be fully specified.**
```


oh wait does this just mean i need to include the .js at the end?

ljharb: Yes, ESM doesn't resolve extensions.

ntucker: What's a good tool to add js for js files? I can't just add it manually as they start off as .ts files

ljharb: If babel is doing the typescript transpilation (as it should be) then you should be able to use https://www.npmjs.com/package/babel-plugin-module-resolver to transpile and add extensions.

Your current approach should work as well.

ntucker: Is there a project already using that package? I was having trouble writing the alias regex

ljharb: I typically only see it in applications; I haven't used it in projects myself. I'd expect using https://github.com/tleunen/babel-plugin-module-resolver/blob/HEAD/DOCS.md#regular-expressions would work with something like `"^(@rest\/hooks\/test.+)(\.[jt]sx?)?$": "\\1.js"`?

ljharb (approved):

PR #629, opened 2021-03-08
fix: Allow legacy Resource to use getFetchInit()

ntucker:
### Motivation

This was provided but never actually used. This eases migration to [@rest-hooks/rest](https://www.npmjs.com/package/@rest-hooks/rest) allowing users to incrementally adopt the new method before switching.

### Solution

Allows use of `getFetchOptions()` or `getFetchInit()`


ntucker:
These are not equivalent - there was no getFetchInit() equivalent in 4

PR #647, opened 2021-03-10
docs(fix): Fix hooks package link

ntucker:
Existing link 404


ljharb (approved):
the old URL should probably have a redirect, since [cool URLs don't change](https://www.w3.org/Provider/Style/URI)

ntucker:
I'm not sure there ever was that old url - it's not in any of the other docs versions. I think it was just a typo maybe

PR #655, opened 2021-03-14
fix: non-JSON payloads on legacy Resource

ntucker:
### Motivation

In #552 we made sure to only set the header with the correct payloads. However, legacy Resource still will always JSON encode body even if FormData is sent.

### Solution

Backport check from @rest-hooks/rest

### DIFF
--- a/packages/rest-hooks/src/resource/Resource.ts
+++ b/packages/rest-hooks/src/resource/Resource.ts
@@ -38,7 +38,7 @@ export default abstract class Resource extends SimpleResource {
       method: method.toUpperCase(),
     };
     options = this.getFetchInit(options);
-    if (body) options.body = JSON.stringify(body);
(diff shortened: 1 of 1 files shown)


packages/rest-hooks/src/resource/Resource.ts
```diff
@@ -76,3 +76,13 @@ export default abstract class Resource extends SimpleResource {
...
 }
+
+const proto = Object.prototype;
+const gpo = Object.getPrototypeOf;
+
+function isPojo(obj: unknown): obj is Record<string, any> {
+  if (obj === null || typeof obj !== 'object') {
+    return false;
+  }
+  return gpo(obj) === proto;
```
ljharb, line 87: this will break if the object is from another realm (iframe, or node’s vm, etc) but there’s no way to solve that for this kind of check.

ljharb (approved):

PR #688, opened 2021-03-27
enhance: Throw error when urlRoot is not defined

ntucker:
### Motivation

Descriptive aid in development

### Solution

By throwing in key we have the context that it's an improperly defined Resource, so instructions for correction can be more specific.

Since this is a static definition problem we only provide it during development to improve runtime.


ljharb (approved):
Seems like it should still warn in production.

PR #706, opened 2021-04-01
internal: Benchmark CI

ntucker:
### Motivation

Track performance improvements and regressions

### Solution

https://github.com/rhysd/github-action-benchmark

### A note

As far as watching the benchmark results of examples in this repository, the amplitude of the benchmarks is about +- 10~20%. If your benchmarks use some resources such as networks or file I/O, the amplitude might be bigger.

If the amplitude is not acceptable, please prepare a stable environment to run benchmarks. GitHub action supports self-hosted runners.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,55 @@
+name: Benchmark
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+jobs:
+  benchmark:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2-beta
+      with:
+        fetch-depth: 1
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '14'
+    - name: Run benchmark
+      run: yarn install && yarn bench | tee output.txt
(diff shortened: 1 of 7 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,55 @@
+name: Benchmark
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
ljharb, line 7: why not on every branch? there's no allotment on actions usage.

ljharb (approved):
i have never seen a benchmarking approach that provides a consistent baseline; if this does, i'd love to figure out how you did it and use it elsewhere :-)

.github/workflows/benchmark.yml
ntucker: only interesting in tracking things against master - as both the comparison point and points on a graph.

PRs will still measure but not log the data

ntucker:
Not sure what you mean about consistent baseline. Synthetic benches are easier to control; the challenge here is simply hardware variance - hence the note about only really to +/- 10% accuracy.

PR #714, opened 2021-04-03
internal(pkg): Remove husky

ntucker:
We don't use it anywhere


ljharb (approved):
sweet, not a good tool anyways

PR #716, opened 2021-04-04
enhance: All schema members are optional by default

### LINKED ISSUE #637: Denormalization fails if nested entity lists are intended to be optional
**React version**
17.0.1

**Concurrent mode** no

rest-hooks: 5.0.9
@rest-hooks/rest: 1.0.4

**Describe the bug**
I'm trying to set up a resource with a nested list of entities that is intended to be optional.
```ts
export class SomeResource extends Resource {
  id: string = "";
  things: Nested[] = [];

  pk() {
    return this.id;
  }

  static schema = {
    things: [Nested]
  };

  static urlRoot = "/some";
}
```

Responses where `things` is not defined are normalized as expected, when inspecting the state with redux devtools everything is as expected. The default value is set:

Response:
```json
[
  {
    "id": "1",
    "name": "fails"
  },
  {
    "id": "2",
    "name": "works",
    "things": [
      {
        "id": "1",
        "name": "first"
      },
      {
        "id": "2",
        "name": "second"
      }
    ]
  }
]
```

State: 
```ts
{
  entities: {
    Nested: {
      '1': {
        id: '1',
        name: 'first'
      },
      '2': {
        id: '2',
        name: 'second'
      }
    },
    '/some': {
      '1': {
        id: '1',
        name: 'fails',
        things: []
      },
      '2': {
        id: '2',
        name: 'works',
        things: [
          1,
          2
        ]
      }
    }
  },
  results: {
    'GET /some': [
      '1',
      '2'
    ]
  }
}
```

When denormalized the first entity is missing: 

```ts

const list = useResource(SomeResource.list(), {});
// [{ id: '2', name: 'works', things: [...] }]

const fails = useCache(SomeResource.detail(), { id: '1' });
// undefined

```

**To Reproduce**
https://codesandbox.io/s/nostalgic-jepsen-17i64?file=/src/Resource.tsx

**Expected behavior**
Resource should be denormalized with the default value:

```ts

const list = useResource(SomeResource.list(), {});
// [{ id: '1', name: 'fails', things: [] }, { id: '2', name: 'works', things: [...] }]

const fails = useCache(SomeResource.detail(), { id: '1' });
// { id: '1', name: 'fails', things: [] }

```

**Additional context**
No automatic schema validation warnings are triggered. I would expect denormalization to work even if it did or if it was silenced.

ntucker:
Previously failing or deleting results will
now work:
- Arrays no longer delete items where members are missing
- Results missing members will return what they have

Also:
- Results in cache are always 'ready', and thus ignore found
- We only infer results, and not entities, thus entities' nested members are always 'found'

Fixes #637.

### Motivation

Requiring explicit marking of optional is cumbersome.

Principal: easy adoption, incremental safety

### Solution

Previously hard coding required into these paths was thought to be necessary so inferred results wouldn't prematurely return results. However, it is easy to detect when this is the case - enabling us to be more lax when actual results exist. Furthermore, with deletes marked explicitly we don't expect any internal state problems.

Changes:
- Arrays filter only based on deleted, or if the value is undefined
- Full results from useCache and useResource check if inferred to determine whether they care about 'found'
- Entities' nested found is always true

Testing:
- Just commented out tests for 'required' fields, as these will be useful to modify when we have opt-in required
- Infinite loop case with synthetic errors is no longer triggerable due to results cache ignoring 'found' result.
- New test module based on https://github.com/coinbase/rest-hooks/issues/637 example with a few additions

### Next Steps

Opt-in to stricter requirements

```typescript
export class SomeResource extends Resource {
  id: string = "";
  things: Nested[] = [];
  requiredThings: Nested[] = [];
  requiredPrimitive: number = 0;

  pk() {
    return this.id;
  }

  static schema = {
    things: [Nested],
    // [] is just shorthand for Array(), but using the longer version lets us define additional options
    requiredThings: new schema.Array(Nested, { required: true }),
    // introduce a new schema that *only* is used for validation (no data transforms)
    requiredPrimitive: new schema.Required(0),
  };

  static urlRoot = "/some";
}
```

### Breaking change? Semver?

This is technically a breaking behavior. However,

- It is *relaxing* failures. Therefore it's hard to imagine existing working cases failing.
  - *not* changing behavior from one case to another
  - *not* restricting powers, making new cases fail
- Is a hurdle for those upgrading to 5.

Therefore, we are pragmatically not doing a semver bump.


ljharb (approved):
LGTM but seems like it should have some tests

ntucker:
haha, meant to be a draft, still haven't updated existing tests or created the new ones :P

PR #723, opened 2021-04-08
fix: schema.Values singleSchema denormalization

### LINKED ISSUE #722: schema.Values not denormalized

**Concurrent mode** yes/no

**Describe the bug**
I'm attempting to model an endpoint that returns a dictionary of resources. I've defined a custom endpoint on my resource where I override schema with `new schema.Values(this)`.

The normalization works as expected, but the output from `useResource` is not denormalized.

**To Reproduce**
https://codesandbox.io/s/peaceful-smoke-0ges9?file=/src/App.tsx

**Expected behavior**
I would expect denormalization of Values schema to work.

**Additional context**

ntucker:
Fixes #722 .

### Motivation

`schemaHasEntity()` assumed schema.schema would always be pojo. This is not true for Polymorphic types.

### Solution

Use `typeof nestedSchema === 'function'` to check if the nested schema is another schema, rather than pojo - and in this case, simply recurse.

Added tests to both normalizr (validating not a problem there), as well as full integration with `useResource()`. Added for union as well to be comprehensive.

### Open questions

This is tightly coupled to internal implementations of schemas, rather than public APIs. We should establish a pattern to allow extensibility and reduce breakages.

### DIFF
--- a/packages/core/src/state/selectors/useDenormalized.ts
+++ b/packages/core/src/state/selectors/useDenormalized.ts
@@ -144,6 +144,9 @@ function schemaHasEntity(schema: Schema): boolean {
   if (schema && (typeof schema === 'object' || typeof schema === 'function')) {
     const nestedSchema =
       'schema' in schema ? (schema.schema as Record<string, Schema>) : schema;
+    if (typeof nestedSchema === 'function') {
+      return schemaHasEntity(nestedSchema);
+    }
     return Object.values(nestedSchema).reduce(
       (prev, cur) => prev || schemaHasEntity(cur),
       false,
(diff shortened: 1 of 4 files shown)


packages/core/src/state/selectors/useDenormalized.ts
```diff
@@ -144,6 +144,9 @@ function schemaHasEntity(schema: Schema): boolean {
   if (schema && (typeof schema === 'object' || typeof schema === 'function')) {
     const nestedSchema =
       'schema' in schema ? (schema.schema as Record<string, Schema>) : schema;
+    if (typeof nestedSchema === 'function') {
```
ntucker, line 147: only code change. this is mostly tests PR

ljharb (approved):

PR #726, opened 2021-04-09
enhance: Support TypeScript 3, while using TypeScript 4 features

ntucker:
### Motivation

Many type improvements come from latest TypeScript versions. Especially 4. However, we want to continue supporting older versions for backwards compatibility - even if their types are not as strong.

### Solution

- [typesVersions](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions) allows publishing distinct ambient declaration files for different version ranges.
- Use downlevel-dts to automatically get rid of some modern syntax
- For non-convertable features, copy from legacy specific ambient declaration


package.json
```diff
@@ -85,5 +85,8 @@
   "resolutions": {
     "@lerna/conventional-commits": "https://github.com/ntucker/lerna-conventional-commits.git"
   },
-  "version": "0.0.0"
+  "version": "0.0.0",
+  "dependencies": {
+    "downlevel-dts": "^0.7.0"
```
ljharb, line 90: why is this a runtime dep, that non-ts users would need to download too?

ntucker: in this case it doesn't matter because this is the top level package - which is just for workspaces (doesn't get published); however, i'll move it for bookkeeping

ljharb (approved):

PR #730, opened 2021-04-09
internal(fix): Add branch name to both

ntucker:
Wish there was a way to test this more easily


ljharb (approved):
There is; run it on every push on every branch, and then just push up to a branch until it works.

PR #735, opened 2021-04-12
fix: Denormalizing polymoprhic with undefined no longer crashes

ntucker:
### Motivation

In case undefined shows up in a place a Union is expected

### DIFF
--- a/packages/normalizr/src/schemas/Polymorphic.js
+++ b/packages/normalizr/src/schemas/Polymorphic.js
@@ -56,6 +56,9 @@ export default class PolymorphicSchema {
   }
 
   denormalizeValue(value, unvisit) {
+    if (value === undefined) {
+      return [value, false, false];
+    }
     const schemaKey = isImmutable(value) ? value.get('schema') : value.schema;
     if (!this.isSingleSchema && !schemaKey) {
       return [value, true, true];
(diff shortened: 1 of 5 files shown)


packages/normalizr/src/schemas/Polymorphic.js
```diff
@@ -56,6 +56,9 @@ export default class PolymorphicSchema {
   }
 
   denormalizeValue(value, unvisit) {
+    if (value === undefined) {
+      return [value, false, false];
+    }
```
ljharb, line 61: wouldn't this still crash if value is `null`?
```suggestion
    if (value == null) {
      return [value, false, false];
    }
```

ljharb (approved):

packages/normalizr/src/schemas/Polymorphic.js
ntucker: null is dealt with on line 101 of denormalize. added null to test for sanity

ljharb: this is a separate function that could be called independently, with `null`. in that case, it would crash.

ntucker: okay updating types so they can't be passed null

PR #745, opened 2021-04-12
internal(ci): Only run codeql when packages change

ntucker:


ljharb (approved):

ntucker:
@ljharb is "build Expected — Waiting for status to be reported " mean this is configure to require the action i just made only run sometimes?

ljharb:
You renamed the action, and the "required" status is tied to the action name.

ljharb:
`compressed-size-build` can't be required if conditional, so i can just uncheck "build"?

ntucker:
I renamed it because I thought maybe the name confused someone - it should never have been required.

PR #750, opened 2021-04-13
enhance: Remove extraneous outdated hack

ntucker:
### Motivation

A year and a half ago this hack was introduced to allow for measuring rest-hooks with 0 length cache policy for A/B performance testing. This was introduced to prevent infinite loop problems because the code was not as robust back then. Since then, suspense logic was reworked to allow suspending anywhere with useInvalidator, and deletes. The hack has not been needed for some time. However, it remained in place in case there were any dependencies on it.

- The mechanisms were never documented or exposed
- The Coinbase experiment has long since been removed
- I can't imagine any use of building something relying on this - other than to get the same mechanisms that will continue to work without it.

### Solution

Delete hack


ljharb (approved):
¯\\\_(ツ)\_/¯

PR #762, opened 2021-04-20
feat: Endpoint.bind()

ntucker:
### Motivation

Using .bind() but still have endpoint capabilities - like calling .key() without needing to send params.

### Solution

Exploit the already existing `.extend()` to do most of the heavy lifting.

### DIFF
--- a/packages/endpoint/src-legacy-types/endpoint.d.ts
+++ b/packages/endpoint/src-legacy-types/endpoint.d.ts
@@ -89,7 +89,7 @@ export interface EndpointInstance<
   >(
     this: E,
     thisArg: ThisParameterType<E>,
-    ...args: readonly [...P]
+    ...args: P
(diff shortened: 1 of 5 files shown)


packages/endpoint/src-legacy-types/endpoint.d.ts
```diff
@@ -0,0 +1,174 @@
...
+  S extends Schema | undefined = Schema | undefined,
+  M extends true | undefined = true | undefined
+> extends EndpointOptions<K, S, M> {
+  fetch?: FetchFunction;
+}
+
+export type ParamFromFetch<F> = F extends (
+  params: infer P,
+  body?: any,
+) => Promise<any>
```
ljharb, line 32: should this have a `T` so the promise that comes back has the expected type?

ljharb (approved):

PR #770, opened 2021-04-25
fix: Delete should only be triggered on finding DELETE symbol

ntucker:
### Motivation

Deletes have special behavior and should never trigger when something else is awry.

### DIFF
--- a/packages/normalizr/src/schemas/__tests__/__snapshots__/Array.test.js.snap
+++ b/packages/normalizr/src/schemas/__tests__/__snapshots__/Array.test.js.snap
@@ -971,6 +971,31 @@ Array [
   ],
   Array [
     Object {
+      "id": "789",
+      "name": "fido",
+    },
+    Array [
+      Object {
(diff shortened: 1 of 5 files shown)


ljharb (approved):
are there any tests that aren't just snapshots? the tests in this PR don't make it clear in any way what the problem was

ntucker:
Adding some console warnings and a test to capture this

packages/normalizr/src/schemas/__tests__/__snapshots__/Array.test.js.snap
```diff
@@ -969,6 +969,31 @@ Array [
     ],
     undefined,
   ],
+  Array [
```
ntucker, line 972: not the best test....as the snapshot needs to be updated based on the extra calls to this.getSchemaAttribute() to format the console warning

ntucker:
@christiancho you might like this one ;)

PR #776, opened 2021-04-26
enhance: Flexible url parameters for RestEndpoint

ntucker:
### Motivation

Flexible parameters can even be useful for typical rest patterns. For instance, not requiring anything for a list, or simply having the id itself as a single parameter for a detail endpoint.

### DIFF
--- a/packages/endpoint/src/endpoint.d.ts
+++ b/packages/endpoint/src/endpoint.d.ts
@@ -143,7 +143,7 @@ export interface EndpointInstance<
     : IfAny<M, any, IfTypeScriptLooseNull<'read', 'mutate'>>;
 
   /** @deprecated */
-  getFetchKey(params: Parameters<F>[0]): string;
+  getFetchKey(params: Parameters<F>): string;
   /** @deprecated */
   options?: EndpointExtraOptions<F>;
 }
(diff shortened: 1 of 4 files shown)


packages/endpoint/src/endpoint.d.ts
```diff
@@ -143,7 +143,7 @@ export interface EndpointInstance<
     : IfAny<M, any, IfTypeScriptLooseNull<'read', 'mutate'>>;
 
   /** @deprecated */
-  getFetchKey(params: Parameters<F>[0]): string;
+  getFetchKey(params: Parameters<F>): string;
```
ljharb, line 146: is this the right type for this? it kind of seems like you want this to take varargs, but this means instead of taking a single thing, it now takes an array of things?

ntucker: good call

ntucker: added the same test that key() had to getFetchKey()

ljharb (approved):

ntucker:
Backed out of the getFetchKey change....this would potentially break existing pieces. Best to leave legacy as is and only use new things in new contexts.

Issue #760, opened 2021-04-17
RFC: useNewFetcher()


ntucker:
Would love feedback @dair-targ, @smith-it @Jamakase @Secretmapper @Jsarihan @yunyu @praxxis @zacharyfmarion @kmartinezmedia

## Motivation

Fixes #463, #458, #428, #316, #358

Lots of improvements can be made to useFetcher() design - both based on understanding of use cases, as well as evolution of interfaces like [Endpoint](https://resthooks.io/docs/api/Endpoint).

Providing a new hook allows for incremental adoption of a complete redesign of many of the concepts incorporating 2 years of learnings and experience across many companies.

There are two major areas of change here
- The hook and fetch itself
- Programmable side-effects

References:
- [Endpoint](https://resthooks.io/docs/api/Endpoint) (these replace FetchShape in [5.0](https://resthooks.io/docs/upgrade/upgrading-to-5))
- [UpdateParams](https://resthooks.io/docs/api/useFetcher#updateparams-destendpoint-destparams-updatefunction)

<details><summary><b>class User</b></summary>

```typescript
class User extends Entity {
  readonly username: string = '';
  readonly firstName: string = '';
  readonly lastName: string = '';
  readonly isAdmin: boolean = false;
  pk() { retulrn this.username }

  get getterValue() { return this.username.toUpperCase(); }
  get fullName() { return `${this.firstName} ${this.lastName}`; }
}
```

</details>

Summary:

- One hook, many endpoints
- Completely flexible, variable arguments
- Resolve/return value
- Endpoint definition - new updater

### One hook, many endpoints

The rules of hooks are very restrictive, so the less hooks you have to call, the more flexible. This also benefits render performance. In many cases you might want to fetch many different endpoints. What's worse is if you don't know which endpoints you might want to fetch upfront. With old design you'd have to hook up every *possible* one. This really destroys fetch-as-render pattern, as you want to be able to prefetch based on possible routes.

#### Before

```tsx
const createUser = useFetcher(User.create());
const refreshUsers = useFetcher(User.list());


return (
<form onSubmit={() => createUser({}, userPayload)}>
  <button onClick={() => refreshUsers({})}>Refresh list</button>
</form>
);
```

### After

```tsx
const fetch = useNewFetcher();

return (
<form onSubmit={() => fetch(User.create(), {}, userPayload)}>
  <button onClick={() => fetch(User.list(), {})}>Refresh list</button>
</form>
);
```

### Completely flexible, variable arguments

[...]

PR #781, opened 2021-04-27
feat: @rest-hooks/experimental

ntucker:
### Motivation

Allow quick breaking iteration on future plans to fully flesh them out before committing to them.

### Solution

We start with the first pieces of the new `useFetcher()` implementation described in #760.

This version is not complete, but this PR is more about adding the new package.

Differences:
- Uses EndpointInterface
- Works with variable number of parameters


__tests__/new.ts
```diff
@@ -212,6 +212,16 @@ export class TypedArticleResource extends CoolerArticleResource {
 }
 
 export class FutureArticleResource extends CoolerArticleResource {
+  static url(id: any): string {
+    if (this.pk({ id }) !== undefined) {
+      if (this.urlRoot.endsWith('/')) {
+        return `${this.urlRoot}${this.pk({ id })}`;
+      }
+      return `${this.urlRoot}/${this.pk({ id })}`;
```
ljharb, line 220: nit:
```suggestion
      return `${this.urlRoot.replace(/\/$/, '/')}/${this.pk({ id })}`;
```

ljharb (approved):

__tests__/new.ts
ntucker: It was suppsoed to be `${this.urlRoot.replace(/\/$/, '')}/${this.pk({ id })}`; right? (replace / with empty string)

ljharb: oh whoops, yes :-)

PR #805, opened 2021-05-09
enhance: Remove dependency on fromJS() from denormalize

ntucker:
### Motivation

Reduces interface dependency on denormalize from Entity by making Entity aware of interface requirements for denormalize.

Previously fromJS was called from denormalize when constructing entities, so the local cache could be set. This created a dependency not only on the particular construction method, but also the concept around mutability. Having denormalize mutate a seemingly immutable SimpleRecord is like giving it private access.

### Solution

Instead, we flip the script, providing `setLocal()` callback that requires Entity implementers to call appropriately. Since copying and setting the localcache before is only required to prevent infinite loops in nesting, it seems like this knowledge is already somewhat coupled to the implementation. For instance, there is no need for FlatEntity to use this callback as it does not recurse

### Potential for breaking
Custom denormalize implementations will still work as normal - only not updating cache before nesting. The only case where they could break is where an implementation does nesting. However, this is unlikely to be the case while also not calling super. Also note this has no impact on legacy Resource from `rest-hooks` as that extends from FlatEntity.

### DIFF
--- a/packages/normalizr/src/entities/Entity.ts
+++ b/packages/normalizr/src/entities/Entity.ts
@@ -274,13 +274,18 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
     // note: iteration order must be stable
     Object.keys(this.schema).forEach(key => {
       const schema = this.schema[key];
-      const nextInput = key in input ? input[key] : undefined;
+      const nextInput = Object.hasOwnProperty.call(input, key)
+        ? input[key]
+        : undefined;
       const [value, , deletedItem] = unvisit(nextInput, schema);
 
-      if (deletedItem && !(key in instance && !instance[key])) {
+      if (
+        deletedItem &&
+        !(Object.hasOwnProperty.call(input, key) && !instance[key])
+      ) {
         deleted = true;
       }
-      if (key in input && input[key] !== value) {
+      if (Object.hasOwnProperty.call(input, key) && input[key] !== value) {
         entityCopy[key] = value;
       }
     });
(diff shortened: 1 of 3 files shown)


packages/normalizr/src/entities/Entity.ts
```diff
@@ -258,25 +270,22 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
...
     let deleted = false;
-    const denormEntity = input;
 
     // note: iteration order must be stable
     Object.keys(this.schema).forEach(key => {
       const schema = this.schema[key];
-      const nextInput = this.hasDefined(input, key as any)
-        ? input[key]
-        : undefined;
+      const nextInput = key in input ? input[key] : undefined;
```
ljharb, line 277: `in` will also trigger for Object.prototype keys - the previous logic used `hasDefined` which implies it's checking hasOwnProperty.

packages/normalizr/src/entities/Entity.ts
```diff
@@ -258,25 +270,22 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
...
       const [value, , deletedItem] = unvisit(nextInput, schema);
 
       if (deletedItem && !(key in instance && !instance[key])) {
         deleted = true;
       }
-      if (this.hasDefined(input, key as any) && denormEntity[key] !== value) {
-        denormEntity[key] = value;
+      if (key in input && input[key] !== value) {
```
ljharb, line 283: also here

ljharb (approved):

PR #820, opened 2021-05-16
enhance: Simplify normalize addEntity function

ntucker:
BREAKING CHANGE: normalize arg addEntity is now addEntity(schema,
processed, id)

### Motivation

- minimize interfaces
- slight performance win

### Solution

The only reason all props are sent are to call pk(), which already happens - so just pass that on instead.

This will only affect those who override the entity normalize method

### DIFF
--- a/packages/normalizr/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/normalizr/src/__tests__/__snapshots__/index.test.js.snap
@@ -543,16 +543,5 @@ Array [
     },
     undefined,
   ],
-  Array [
-    Recommendations {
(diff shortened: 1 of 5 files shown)


packages/normalizr/src/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -543,16 +543,5 @@ Array [
     },
     undefined,
   ],
-  Array [
```
ntucker, line 546: this is just calls count - deletion here only means we are doing less work!

ljharb (approved):

PR #840, opened 2021-05-23
pkg: jsdom - no more 'request' dep

ntucker:
### Motivation

'request' is deprecated, which adds annoying warnings


ljharb (approved):
not a real security problem, but sure

PR #841, opened 2021-05-24
feat: Add Resource that extends from new Entity

ntucker:
### Motivation

Until entity is updated

### Solution

Also changes inheritance so fetch is included, and endpoints are not defined in BaseResource

### DIFF
--- a/packages/experimental/src/rest/BaseResource.ts
+++ b/packages/experimental/src/rest/BaseResource.ts
@@ -1,0 +1,248 @@
+import { Endpoint } from '@rest-hooks/endpoint';
+
+import Entity from '../entity/EntityRecord';
+import paramsToString from './paramsToString';
+import { RestEndpoint } from './types';
+
+import type { EndpointExtraOptions } from '@rest-hooks/endpoint';
+import type { Schema } from '@rest-hooks/normalizr';
+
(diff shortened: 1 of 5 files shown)


packages/experimental/src/rest/BaseResource.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /** Field where endpoint cache is stored */
+  protected static readonly cacheSymbol = Symbol('memo');
+
+  /** Used to memoize endpoint methods
+   *
+   * Relies on existance of runInit() member.
+   */
+  protected static memo<T>(name: string, construct: () => T): T {
+    if (!Object.hasOwnProperty.call(this, this.cacheSymbol)) {
```
ljharb, line 152:
```suggestion
    if (!Object.prototype.hasOwnProperty.call(this, this.cacheSymbol)) {
```
(altho this will work because Object inherits from Function.prototype which inherits from Object.prototype)

ljharb (approved):

packages/experimental/src/rest/BaseResource.ts
ntucker: this is copy/paste from @rest-hooks/rest. i'll keep this in mind for future updates tho.

PR #843, opened 2021-05-24
feat: Endpoint.update programmable sideeffects

ntucker:
### Motivation

Iterating toward https://github.com/coinbase/rest-hooks/issues/760

- Define programmable update in endpoint
- Improve function signature

### Solution

We maintain working on raw reducer state, however we introduce improvements on the update method itself being much more straightforward and easy to define correctly.

- two degree function
- map key -> update function

```typescript
const createUser = new Endpoint(postToUserFunction, {
  schema: User,
  update: (newUser: Denormalize<S>)  => {
    const updates = {
      [userList.key()]: (users = []) => [newUser, ...users],
      [userList.key({ sortBy: 'createdAt' })]: (users = [], { sortBy }) => {
        const ret = [createdUser, ...users];
        ret.sortBy(sortBy);
        return ret;
      },
    ];
    if (newUser.isAdmin) {
      updates[userList.key({ admin: true })] = (users = []) => [newUser, ...users];
    }
    return updates;
  },
});
```

#### This PR

- Maintain previous updaters in reducer
- add update method to actions
- introduce usage in useFetcher found in experimental

#### Stage 2

- Make previous useFetcher() method transform into new update meta

#### Stage 3

- Stop supporting old update meta by removal from reducer

### DIFF
--- a/packages/core/src/state/reducer.ts
+++ b/packages/core/src/state/reducer.ts
@@ -71,9 +71,9 @@ export default function reducer(
         results = applyUpdatersToResults(results, result, action.meta.updaters);
         if (action.meta.update) {
           const updaters = action.meta.update(result);
-          Object.keys(updaters).forEach(
-            key => { results[key] = updaters[key](results[key]); },
-          );
+          Object.keys(updaters).forEach(key => {
+            results[key] = updaters[key](results[key]);
+          });
         }
         return {
           entities: mergeDeepCopy(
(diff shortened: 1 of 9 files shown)


packages/core/src/state/reducer.ts
```diff
@@ -69,6 +69,12 @@ export default function reducer(
           [action.meta.key]: result,
         };
         results = applyUpdatersToResults(results, result, action.meta.updaters);
+        if (action.meta.update) {
+          const updaters = action.meta.update(result);
+          Object.keys(updaters).forEach(
+            key => (results[key] = updaters[key](results[key])),
```
ljharb, line 75: nit:
```suggestion
            key => { results[key] = updaters[key](results[key]); },
```
seems kind of weird to use the parens rather than a block body

ljharb (approved):

PR #855, opened 2021-05-24
fix: Correct peerDeps

ntucker:
### Motivation

Reliance on new core stuff. Forgot to bump in previous commit


ljharb (approved):
this is a breaking change, but it's probably fine since it's in experimental

PR #866, opened 2021-05-26
pkg: react-hooks-testing-library v7

ntucker:
BREAKING CHANGES:
- requires node 12
- 'suppressErrorOutput will now work when explicitly called, even if the
RHTL_DISABLE_ERROR_FILTERING env variable has been set' (from
react-hooks-testing-library)


packages/test/package.json
```diff
@@ -33,7 +33,7 @@
   },
   "type": "module",
   "engines": {
-    "node": ">=10 <13 || ^13.7 || >=14"
+    "node": "^13.7 || >=14"
```
ljharb, line 36: the OP says "requires node 12" but this drops it. do you mean `>= 12 || ^13.7 || >= 14`?

packages/test/package.json
```diff
@@ -89,10 +89,10 @@
     "url": "https://github.com/coinbase/rest-hooks/issues"
   },
   "dependencies": {
-    "@testing-library/react-hooks": "~5.0.3"
+    "@testing-library/react-hooks": "~7.0.0"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.0.0-beta.1",
+    "@rest-hooks/core": "^1.0.0-beta.1 || ^1.0.0",
```
ljharb, line 95:
```suggestion
    "@rest-hooks/core": "^1.0.0-0",
```

ntucker: what does -0 mean? are there docs on this?

ljharb: It includes all 1.0.0 prereleases and also 1.0.0 itself.

you can check on semver.npmjs.com if a range works for you.

ljharb (approved):

PR #874, opened 2021-05-30
internal: Major version bump

ntucker:
BREAKING CHANGE:
- requires node 12
- 'suppressErrorOutput will now work when explicitly called, even if the
RHTL_DISABLE_ERROR_FILTERING env variable has been set' (from
react-hooks-testing-library)

It wasn't picking up on the version change, so trying to trigger it here.


ljharb (approved):
"fix" seems a weird commit/PR prefix?

PR #889, opened 2021-06-02
feat: Export types for Resource

ntucker:
### Motivation

Just as the types from @rest-hooks/rest

### DIFF
--- a/packages/experimental/src/rest/types.ts
+++ b/packages/experimental/src/rest/types.ts
@@ -37,4 +37,18 @@ export interface RestEndpoint<
   ) => any;
   method: string;
   signal: AbortSignal | undefined;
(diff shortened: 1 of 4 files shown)


packages/experimental/src/rest/types.ts
```diff
@@ -37,4 +37,18 @@ export interface RestEndpoint<
   ) => any;
   method: string;
   signal: AbortSignal | undefined;
+  paginated?: (...args: any) => any;
```
ljharb, line 40: it really doesn't matter what return type this function has??

ljharb (approved):

packages/experimental/src/rest/types.ts
ntucker: made it more specialized

PR #900, opened 2021-06-07
feat: Add inferResults()

ntucker:
### Motivation

Extracting `buildInferredResults()` into normalizr

- Simplify logic
- Customizable inference behavior including disabling
- Start supporting variable arg length

### Solution

Add `infer()` to schemas. Somewhat similar to normalize, tho only tries to build results piece - no entity collection.

### Next

Utilize this in @rest-hooks/core. We split this up so we can distinctly mark a dependency on this feature in the other package.

### DIFF
--- a/packages/experimental/src/entity/Entity.ts
+++ b/packages/experimental/src/entity/Entity.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 import { schema } from '@rest-hooks/endpoint';
 import type { AbstractInstanceType, Schema } from '@rest-hooks/endpoint';
+import { NormalizedIndex } from '@rest-hooks/normalizr';
 
 import { isImmutable, denormalizeImmutable } from './ImmutableUtils';
 
@@ -279,6 +280,24 @@ First three members: ${JSON.stringify(processedEntity.slice(0, 3), null, 2)}`;
     return id;
   }
 
+  static infer(args: any[], indexes: NormalizedIndex, recurse: any): any {
+    if (!args[0]) return undefined;
+    const id = this.pk(args[0], undefined, '');
+    // Was able to infer the entity's primary key from params
+    if (id !== undefined && id !== '') return id;
+    // now attempt lookup in indexes
+    const indexName = indexFromParams(args[0], this.indexes);
+    if (indexName && indexes[this.key]) {
+      // 'as Record<string, any>': indexName can only be found if params is a string key'd object
+      const id =
+        indexes[this.key][indexName][
+          (args[0] as Record<string, any>)[indexName]
+        ];
+      return id;
+    }
+    return undefined;
+  }
+
   static denormalize<T extends typeof Entity>(
     this: T,
     input: Readonly<Partial<AbstractInstanceType<T>>>,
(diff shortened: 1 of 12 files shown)


packages/experimental/src/entity/Entity.ts
```diff
@@ -358,3 +377,14 @@ if (process.env.NODE_ENV !== 'production') {
 export function isEntity(schema: Schema): schema is typeof Entity {
   return schema !== null && (schema as any).pk !== undefined;
 }
+
+function indexFromParams<I extends string>(
+  params: Readonly<object>,
+  indexes?: Readonly<I[]>,
+) {
+  if (!indexes) return undefined;
+  for (const index of indexes) {
```
ljharb, line 386: it’d transpile a lot smaller to use `return indexes.find((index) => index in params)` (but also, shouldn’t this use hasOwnProperty?)

packages/normalizr/src/entities/Entity.ts
```diff
@@ -307,3 +323,14 @@ if (process.env.NODE_ENV !== 'production') {
...
   return schema !== null && (schema as any).pk !== undefined;
 }
+
+function indexFromParams<I extends string>(
+  params: Readonly<object>,
+  indexes?: Readonly<I[]>,
+) {
+  if (!indexes) return undefined;
+  for (const index of indexes) {
+    if (index in params) return index;
```
ljharb, line 333: same comment here

packages/normalizr/src/inferResults.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (typeof schema === 'object' && schema) {
+    const method = Array.isArray(schema) ? arrayInfer : objectInfer;
+    return method(schema, args, indexes, inferResults);
+  }
+
+  return schema as any;
+}
+
+function canInfer(schema: Schema): schema is Pick<SchemaSimple, 'infer'> {
+  return !!schema && typeof (schema as any).infer === 'function';
```
ljharb, line 28: why is the cast needed? does Schema not have an optional infer property?

packages/normalizr/src/schemas/Object.js
```diff
@@ -52,6 +52,14 @@ export const denormalize = (schema, input, unvisit) => {
   return [object, found, deleted];
 };
 
+export function infer(schema, args, indexes, recurse) {
+  let resultObject = {};
+  for (const k in schema) {
```
ljharb, line 57: presumably you want to iterate Object.keys(schema) here, rather that using for-in?

packages/normalizr/src/schemas/Values.js
```diff
@@ -46,4 +46,8 @@ export default class ValuesSchema extends PolymorphicSchema {
       deleted,
     ];
   }
+
+  infer(args, indexes, recurse) {
+    return undefined;
```
ljharb, line 51: if infer isn’t required, then why define it when it returns undefined?

packages/normalizr/src/inferResults.ts
ntucker: Schema can be a whole lot of things including undefined, pojos, array.

ntucker: the class based minimal interfaces are typed as SchemaSimple

packages/normalizr/src/schemas/Values.js
ntucker: it is required for all class cases. every algorithm has special code paths for the non class case but no other type discrimination otherwise.

packages/normalizr/src/inferResults.ts
ljharb: why if the type here was `Schema & { infer?: unknown => unknown }` or something?

ljharb (approved):
last comment is a nonblocker

packages/normalizr/src/inferResults.ts
ntucker: not sure I understand what you're saying here. are you suggesting that i use that type for the inferred piece?

ljharb: I’m suggesting using that for the argument, so you don’t need the cast.

PR #901, opened 2021-06-07
enhance: Use inferResults() from normalizr

ntucker:
Builds on https://github.com/coinbase/rest-hooks/pull/900

### Motivation

Extracting `buildInferredResults()` into normalizr

- Simplify logic
- Customizable inference behavior including disabling
- Start supporting variable arg length

### Solution

Add `infer()` to schemas. Somewhat similar to normalize, tho only tries to build results piece - no entity collection.

### DIFF
--- a/packages/rest-hooks/src/internal.ts
+++ b/packages/rest-hooks/src/internal.ts
@@ -5,5 +5,7 @@ export {
   DispatchContext,
   hasUsableData,
 } from '@rest-hooks/core';
+// TODO: Delete next line for v6
 export const buildInferredResults = __INTERNAL__.buildInferredResults;
+export const inferResults = __INTERNAL__.inferResults;
 export const RIC = __INTERNAL__.RIC;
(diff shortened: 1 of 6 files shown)


packages/rest-hooks/src/internal.ts
```diff
@@ -5,5 +5,5 @@ export {
   DispatchContext,
   hasUsableData,
 } from '@rest-hooks/core';
-export const buildInferredResults = __INTERNAL__.buildInferredResults;
```
ljharb, line 8: doesn’t this mean it’s a breaking change?

ntucker: __INTERNAL__ is sort of like 'unstable'. current policy (up for debate) is to make breaking change on those bump minor versions only.

ntucker: Keeping old export and just adding the new one

ljharb (approved):

PR #907, opened 2021-06-10
feat: Support React 18

ntucker:
### Motivation

https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html

### Solution

- Add to peerDeps
- 18 is used on checkout
- test matrix to validate both versions in CI


ntucker:
@ljharb what would you say is the semver bump for adding support for another lib to peerdep? minor or patch?

ljharb (commented):
Always minor, because it’s adding something.

Shouldn’t this be adding to the test matrix somehow, so tests are ran on both react 17 and 18?

ntucker:
Test matrix would be good but I'm unsure of the best approach for that. I have jest setup to test with react native and Dom but it doesn't install different versions of libraries.  Is there an example you know of for this?

ljharb:
enzyme and react-dates, for example, do this by defining a matrix of env vars, and manually installing the right things in the CI config. I don’t know how circle works well enough to configure that tho.

ntucker:
Sweet, I think this is what I need: https://circleci.com/blog/circleci-matrix-jobs/

ntucker:
@ljharb looks like it worked but now the 'required' marks don't apply correctly because there are distinct names for all parameter applications

ljharb (approved):
i've updated the branch protections, so this should be good to go

PR #919, opened 2021-06-13
feat: Support deletes with responses

ntucker:
### Motivation

Previously we used params to essentially infer the pk, since it's very common for delete to return nothing. However, we can easily support this case, as well as the case with something by adding conditional check

### DIFF
--- a/packages/experimental/src/rest/Resource.ts
+++ b/packages/experimental/src/rest/Resource.ts
@@ -126,7 +126,9 @@ export default abstract class Resource extends BaseResource {
     return this.memo('#delete', () =>
       endpoint.extend({
         fetch(this: RestEndpoint, params: any) {
-          return endpoint.fetch.call(this, params).then(() => params);
+          return endpoint.fetch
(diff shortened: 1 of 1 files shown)


packages/experimental/src/rest/Resource.ts
```diff
@@ -126,7 +126,9 @@ export default abstract class Resource extends BaseResource {
     return this.memo('#delete', () =>
       endpoint.extend({
         fetch(this: RestEndpoint, params: any) {
-          return endpoint.fetch.call(this, params).then(() => params);
+          return endpoint.fetch
+            .call(this, params)
+            .then(res => (typeof res === 'object' ? res : params));
```
ljharb, line 131: this allows for `null` as well, is that intended?

ljharb (approved):

packages/experimental/src/rest/Resource.ts
ntucker: ah good catch

PR #920, opened 2021-06-14
feat: Add Entity.expiresAt() - entity TTL configuration

### LINKED ISSUE #876: List vs Detail Resource Issue (V5 upgrade)
**React version** 16.13.1

**Concurrent mode** no

**Rest Hooks version** 5.1.1
**@rest-hooks/rest** 2.1.2,

**Describe the bug**
Most rest APIs have the strategy of showing limited attributes per item for List page while showing detailed attributes with associations for the Details page, e.g

/posts
```js
[
    { id: 1, title: 'Title 1' },
    { id: 2, title: 'Title 2' },
    ...
]
```

/posts/1
```js
{
    id: 1, title: 'Title 1', body: 'Body', user: {}, comments: [], ...
}
```
So, a separate request must be sent when viewing the detail page. This was the case for v4 but since the v5 upgrade, detail request is not being sent.

**To Reproduce**
https://codesandbox.io/s/reverent-mclaren-7r9ys

If you click 'View' for each post, a separate request for /posts/:id is not sent.

ntucker:
Fixes #876 .

### Motivation

Configurable/Programmable TTL for entities

### Solution

Entities+Resources get `expiresAt()`

```ts
class Entity {
  static expiresAt(
    { expiresAt }: { expiresAt: number; date: number },
    input: any,
  ): number {
    return expiresAt;
  }
}
```

By default as above, just uses the fetch's expiry time.

However, this can also check for complete members, to not count toward an entity's expiry time.

#### Example

```ts
class MyResource extends Resource {
  readonly id: string = '';
  readonly basic: string = '';
  readonly extendedContent?: { stuff: string };

  pk() { return this.id }
  static urlRoot = '/myresource/';

  static expiresAt(
    { expiresAt }: { expiresAt: number; date: number },
    input: any,
  ): number {
     // extendedContent only exists when returning full results, so it's safe to infer this entity
    return input.extendedContent ? expiresAt : 0;
  }
}
```

This would allow a partial result returned in a `list()`, and then when visiting `detail()` since the entity will be considered expired, it will stale-while-revalidate. Whereas now it may avoid additional fetch within a time period, as the entity would have specified its expiry as sometime in the future.

### DIFF
--- a/packages/experimental/src/entity/Entity.ts
+++ b/packages/experimental/src/entity/Entity.ts
@@ -298,6 +298,13 @@ First three members: ${JSON.stringify(processedEntity.slice(0, 3), null, 2)}`;
     return undefined;
   }
 
+  static expiresAt(
+    { expiresAt }: { expiresAt: number; date: number },
+    input: any,
+  ): number {
+    return expiresAt;
+  }
+
   static denormalize<T extends typeof Entity>(
     this: T,
     input: Readonly<Partial<AbstractInstanceType<T>>>,
--- a/packages/normalizr/src/entities/Entity.ts
+++ b/packages/normalizr/src/entities/Entity.ts
@@ -239,6 +239,13 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
     return id;
   }
 
+  static expiresAt(
+    { expiresAt }: { expiresAt: number; date: number },
+    input: any,
+  ): number {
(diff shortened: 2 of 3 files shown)


ljharb (approved):
no tests?

ntucker:
Test in next pr https://github.com/coinbase/rest-hooks/pull/921

PR #924, opened 2021-06-15
feat: Add Resource+Entity to legacy

ntucker:
### Motivation

A place for old behavior to exist so code can move on.

### Solution

Move Entity+Resource chain to legacy. rest-hooks simply re-exports which doesn't change its behavior.
In a future release we can remove the export and leave legacy. This makes it easy to incrementally migrate.

Note: All files are simply moved and some book-keeping with package.json and tsconfig are the only actual file changes.


packages/rest-hooks/src/resource/index.ts
```diff
@@ -1,6 +0,0 @@
-import Resource from './Resource';
-import SimpleResource from './SimpleResource';
```
ljharb, line 2: could someone be deep-importing this file?

ljharb (commented):
It seems like a lot of stuff is being moved out of the rest-hooks package - can you confirm the rest-hooks package is unaffected beyond the deep imports that are now missing?

packages/rest-hooks/src/resource/index.ts
ntucker: theoretically i guess. i don't see why they would with them being available at the top level tho.

ntucker:
Yup, the key is to look at the index file for rest hooks and see that the only export changes are

```ts
export { Resource, SimpleResource } from '@rest-hooks/legacy';
export type { SchemaDetail, SchemaList, Method } from '@rest-hooks/legacy';
```

ljharb (approved):

packages/rest-hooks/src/resource/index.ts
ntucker: Interesting...I feel like before 'exports' in package.json this was inherently broken as main vs module was never really consistent between environments

ntucker: I'm curious to learn about the tree shaking limitations. Are there blog posts or something documenting this?

ljharb: Not that i know of. It’s just that tree-shaking is always best effort, and bundle sizes are always smaller when tree shaking isn’t needed. Coinbase no longer allows anything but deep imports in most of its internal code for this reason.

ntucker: If this is in web, seems like lack of webpack 5 is the real problem as webpack 5 does dead code elimination as well.

ljharb: All treeshaking, regardless of bundler, can’t ever possibly do as good a job as not needing it in the first place ¯\\\_(ツ)\_/¯ 

manifest exports/barrel exports/whatever they’re called are an antipattern, albeit a common one, and it’s best avoided.

ntucker: You mean like `import * as React from 'react';`? That's very different than `import { useMemo } from 'react';`

ntucker: The first one breaks everything, the second one not only doesn't break tooling, it allows minifiers to do wonders on the imported pieces

ljharb: Right - but unless `react` is designed quite carefully, `import { useMemo }` still needs to *evaluate* all that code because of side effects and whatnot.

PR #940, opened 2021-06-19
enhance: Entities normalize to POJO

ntucker:
BREAKING CHANGE:
- fromJS() -> process() to customize init
- normalize results in POJO rather than instances
- FlatEntity, SimpleRecord removed (use @rest-hooks/legacy)
- peerDep @rest-hooks/endpoint > 2

### Motivation

- Almost doubles speed of normalize and denormalize
- Simplifies serialize/deserialize cache - makes SSR more straightforward
- Make using record/tuples in future easier

### Solution

This moves @rest-hooks/experimental Entity to normalizr

#### @rest-hooks/normalizr

- Entity normalize turns into pojo
- FlatEntity removed

##### normalize

1. process()
2. pk()
3. validate()
4. merge()

##### denormalize

1. fromJS()

<details><summary><b>Entity.ts</b></summary>

```ts
export default abstract class Entity {
    static toJSON(): {
        name: string;
        schema: {
            [k: string]: Schema;
        };
        key: string;
    };
    /** Defines nested entities */
    static schema: {
        [k: string]: Schema;
    };
    /**
     * A unique identifier for each Entity
     *
     * @param [parent] When normalizing, the object which included the entity
     * @param [key] When normalizing, the key where this entity was found
     */
    abstract pk(parent?: any, key?: string): string | undefined;
    /** Returns the globally unique identifier for the static Entity */
    static get key(): string;
    /** Defines indexes to enable lookup by */
    static indexes?: readonly string[];
    /** Control how automatic schema validation is handled
     *
     * `undefined`: Defaults - throw error in worst offense
     * 'warn': only ever warn
     * 'silent': Don't bother with processing at all
     *
     * Note: this only applies to non-nested members.
     */
    protected static automaticValidation?: 'warn' | 'silent';
    /**
     * A unique identifier for each Entity
     *
     * @param [value] POJO of the entity or subset used
     * @param [parent] When normalizing, the object which included the entity
     * @param [key] When normalizing, the key where this entity was found
     */
    static pk<T extends typeof Entity>(this: T, value: Partial<AbstractInstanceType<T>>, parent?: any, key?: string): string | undefined;
    /** Creates new instance copying over defined values of arguments */
    static merge<T extends typeof Entity>(this: T, existing: Partial<AbstractInstanceType<T>>, incoming: Partial<AbstractInstanceType<T>>): Partial<AbstractInstanceType<T>>;
    /** Factory method to convert from Plain JS Objects.
     *
     * @param [props] Plain Object of properties to assign.
     * @param [parent] When normalizing, the object which included the record
     * @param [key] When normalizing, the key where this record was found
     */
    static fromJS<T extends typeof Entity>(this: T, props?: Partial<AbstractInstanceType<T>>): AbstractInstanceType<T>;
    /** Do any transformations when first receiving input */
    static process(input: any, parent: any, key: string | undefined): any;
    static normalize(input: any, parent: any, key: string | undefined, visit: (...args: any) => any, addEntity: (...args: any) => any, visitedEntities: Record<string, any>): any;
    protected static validate(processedEntity: any): void;
    static infer(args: any[], indexes: NormalizedIndex, recurse: any): any;
    static expiresAt(meta: {
        expiresAt: number;
        date: number;
    }, input: any): number;
    static denormalize<T extends typeof Entity>(this: T, input: Readonly<Partial<AbstractInstanceType<T>>>, unvisit: schema.UnvisitFunction): [AbstractInstanceType<T>, boolean, boolean];
    private static __defaults;
    /** All instance defaults set */
    protected static get defaults(): any;
    /** Used by denormalize to set nested members */
    protected static set(entity: any, key: string, value: any): void;
}
```
</details>

#### @rest-hooks/experimental

- peerDep @rest-hooks/endpoint > 2
- Entity is exported from @rest-hooks/endpoint, since it is moved

#### @rest-hooks/rest

- peerDep @rest-hooks/endpoint > 2
- EntityRecord added


packages/normalizr/src/entities/Entity.ts
```diff
@@ -63,6 +66,39 @@ export default abstract class Entity extends SimpleRecord {
...
   }
 
+  /** Creates new instance copying over defined values of arguments */
+  static merge<T extends typeof Entity>(
+    this: T,
+    existing: Partial<AbstractInstanceType<T>>,
+    incoming: Partial<AbstractInstanceType<T>>,
+  ) {
+    return { ...existing, ...incoming };
+  }
```
ljharb, line 76: this seems like an odd thing to add given that it's so simple and idiomatic to do it inline

ntucker: this allows user customization. https://resthooks.io/docs/api/Entity#static-mergeexisting-incoming-mergedvalue

For instance, coinbase's prices were all wrong until we used timestamps to resolve which

ljharb: ah, meaning that the base functionality is just an object spread, but user entities might want to override it. fair enough

ntucker: If this was merely a helper to simplify code it would be private

ljharb (approved):

PR #955, opened 2021-06-22
feat: Use 'exports' package.json member

ntucker:
BREAKING CHANGE: Importing directly from hidden files is no longer supported
Node>=12

### Solution

https://github.com/coinbase/rest-hooks/pull/590 + https://github.com/coinbase/rest-hooks/pull/628 applied to @rest-hooks/endpoint, @rest-hooks/core, @rest-hooks/legacy, @rest-hooks/normalizr, @rest-hooks/rest, rest-hooks

- Relative imports replaced with @rest-hooks/endpoint, @rest-hooks/core, @rest-hooks/legacy, @rest-hooks/normalizr, @rest-hooks/rest, rest-hooks
  - Use RESOLVER_ALIAS for babel; alias in tsconfig already exists
- legacy directory for "module" entry; add to build - target 2018
- lib is for "exports" entry - target 2020
- dist/index.cjs for node - target node12
- There is no browser specific entry for this package, so index is still used for both entries
- "type": "module"
- Drop node 10 support - `"^12.17 || ^13.7 || >=14"`


packages/endpoint/package.json
```diff
@@ -25,23 +25,40 @@
       ]
     }
   },
+  "exports": {
+    ".": [
+      {
+        "node": {
+          "import": "./node.mjs",
+          "require": "./dist/index.cjs"
+        },
```
ljharb, line 34: the "node" nesting isn't really needed; you could just have "import" and "default" and it'd be fine.
```suggestion
        "import": "./node.mjs",
```

packages/endpoint/package.json
```diff
@@ -25,23 +25,40 @@
...
+        "default": "./lib/index.js"
+      },
+      "./lib/index.js"
+    ],
+    "./package.json": "./package.json"
+  },
+  "type": "module",
```
ljharb, line 41: i strongly discourage this one - type module achieves nothing except breaking a bunch of tools, and making `.js` mean ESM - which will mean your "default" entry point is ESM, not CJS, and thus can't be required.

All ESM should only be `.mjs`, and `.js` should continue to only mean CJS. Using `.cjs` is "fine" but it will break in unexpected ways in older nodes, and it's not really worth the hassle.
```suggestion
```

packages/endpoint/package.json
ntucker: because node supports both cjs and modules, it is important to route node to the same way no matter how the import happens as node can potentially import the same package in both ways. there are several solutions to this, but this one does it by routing all node loads to cjs.

ljharb: I'm not really sure what you mean. with "import" and then "default", node will always import the "import" file (there's only one way to import anything). The "node" nesting only matters if there existed some tool that respected something _else_ - which does not exist.

packages/endpoint/package.json
ntucker: that reminds me to update the engine entry

packages/endpoint/package.json
ntucker: From the previous PR:

References for behavior in node, webpack and typescript (not supported yet)
- https://webpack.js.org/guides/package-exports
- https://nodejs.org/docs/latest-v15.x/api/packages.html
- https://github.com/microsoft/TypeScript/issues/33079

ljharb: I have no idea why these docs authored this, but as a member of the node modules group that defined how it works, I'm telling you this is unnecessary complexity that achieves nothing.

packages/endpoint/package.json
ljharb: That doesn't change that plenty of tools - that are not node - will still break with this. Using it adds nothing except a minor aesthetic change (arguably a bad one).

packages/endpoint/package.json
ntucker: I've been able to reproduce the [dual package hazard](https://nodejs.org/docs/latest-v15.x/api/packages.html#packages_approach_1_use_an_es_module_wrapper) on my own. This is more about how commonjs + es6 modules work together than exports.

packages/endpoint/package.json
ntucker: What tools? The tools this library cares about in order are:

webpack, node, react native, rollup, parcel, snowpack.

This design is optimized for webpack. Though you may disagree with webpack's decisions, that is the primary target. Supporting even tools that aren't primary are also desirable so long as the cost is reasonable.

I'd love to know more about the other tools to evaluate their likelihood of interacting with this library, along with precisely what sorts of things must be considered.

packages/endpoint/package.json
ljharb: This configuration will have no impact on that hazard - it’s exactly the same, but with more steps.

ntucker: I don't follow where you are deriving this conclusion from. Using both es6 module import and commonjs require will route to the same file, which means members like classes will be ===.

ljharb: With the PR (with the "node" object):
A require of the package will hit the ".", then the "node", then the "require", going to `dist/index.cjs`.
An import of the package will hit the ".", then the "node", then the "import", going to `./node.mjs`.

With my suggestion (without the "node" object):
A require of the package will hit the ".", then the "default", going to `dist/index.cjs`.
An import of the package will hit the ".", then the "import", going to `./node.mjs`.

So, regardless of what `node.mjs` does, both approaches are *identical*, but yours has extra steps.

The way to avoid the dual package hazard is to ensure that `node.mjs` imports any state/singletons from the same CJS file that `dist/index.cjs` does - this is orthogonal to either scenario above.

ljharb: With the PR (with the "node" object):
A require of the package will hit the ".", then the "node", then the "require", going to `dist/index.cjs`.
An import of the package will hit the ".", then the "node", then the "import", going to `./node.mjs`.

With my suggestion (without the "node" object):
A require of the package will hit the ".", then the "default", going to `dist/index.cjs`.
An import of the package will hit the ".", then the "import", going to `./node.mjs`.

So, regardless of what `node.mjs` does, both approaches are *identical*, but yours has extra steps.

The way to avoid the dual package hazard is to ensure that `node.mjs` imports any state/singletons from the same CJS file that `dist/index.cjs` does - this is orthogonal to either scenario above.

ntucker: Removing node as you say will still work for node, yes. However, it will serve the cjs bundle to tools like webpack, which is not what we want. CJS is built with babel settings targetting node, vs the lib are targetting browsers. Furthermore, CJS does not support tree shaking, etc, which does work with webpack. (And yes, validated in practice)

ljharb: So you're saying that, despite `. -> import` pointing to the ESM file, webpack needs `. -> node -> import` to be able to find it?

(ftr, CJS absolutely supports treeshaking, precisely as well as ESM; webpack and rollup are just too lazy to implement it)

packages/endpoint/package.json
ntucker: Without knowing specifics - the application to the general is this: .js has existed long before node exits, and still exists outside that realm. Thus, all modern node tools should be aware of type specifier, and thus supporting modern node (legacy node is explicitly marked as not supported) is handled by this specification. This enables maximum compatibility with all tools not related or aware of node - which again is the more primary target of this library.

ntucker: Many tools look for *.js files but not necessarily *.mjs - especially legacy ones.

packages/endpoint/package.json
ntucker: The design and validation is covered in https://github.com/coinbase/rest-hooks/pull/590. This is simply applying the previously validated design to another package.

Here's a summary:

```
  "exports": {
    // this is the export without any path
    ".": [
      {
        // node has its own files for two reasons - they are in commonjs and transpiled to target node 12
        "node": {
          // node world understands these files so this helps maximize compat with this
          // because a given node can pull both the es6 module and commonjs version, this file simply redirects to the cjs file
          "import": "./node.mjs",
           // this is the commonjs version directly
          "require": "./dist/index.cjs"
        },
       // for maximum future or unknown compat, we provide a default case that uses ESM. this targets jan 2020 browsers
        "default": "./lib/index.js"
      },
      // this is fallback in case a tool doesn't understand conditionals
      "./lib/index.js"
    ],
   // tools like importing package.json, but all / imports are explicitly needed
    "./package.json": "./package.json"
  },
```

ntucker: Complexity is set at 4:

- unpkg: I hear this is a thing. Don't use it myself, but I can see its use
- node
- modern module tooling like webpack
- legacy module tooling like webpack 4

ljharb: OK - i now see the `./lib/index.js` default, versus the `./dist/index.cjs` node.require - i'd missed that.

Anything that doesn't understand exports conditions would also not understand ESM, so I'm still confused why this is here. Webpack understands those conditions - so wouldn't it select the ESM version regardless?

ntucker: I'm not sure of specific tools that have this limitation, but it literally just adds another line to the package.json.  We have the same complexity in how many outputs we build. Is it really that problematic to be safe here? In the future this can be removed if it's proved absolutely unnnecessary

packages/endpoint/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "1.2.2",
   "description": "Declarative Network Interface Definitions",
   "sideEffects": false,
-  "main": "dist/index.cjs.js",
-  "module": "lib/index.js",
+  "main": "dist/index.cjs",
```
ljharb, line 6: per discussion, let's add `dist/package.json` with `"type": "commonjs"` and make this a `.js` file.

ljharb (approved):
we've discussed this offline, and it's good to go with this one change.

packages/core/package.json
```diff
@@ -25,28 +25,45 @@
...
+  "exports": {
+    ".": [
+      {
+        "node": {
+          "import": "./node.mjs",
+          "require": "./dist/index.js"
+        },
+        "default": "./lib/index.js"
+      },
+      "./lib/index.js"
```
ljharb, line 37: fwiw, this would only be triggered by versions of node that understand "exports" but not the object form - so it kind of seems like this should either be "whatever's in main", or, you could just omit the array and use the object directly.

ljharb (approved):
still LGTM

packages/core/package.json
ntucker: Not sure why understanding exports would imply something knows cjs and not es6

ljharb: Because there are versions of node that understand exports but not conditions, and require doesn’t understand ESM.

ntucker: Isn't that why the engines mark excludes those versions of node?

ljharb: right - so then what's the array fallback for, since those are the only versions of node that need it?

ntucker: non-node tools? perhaps it should just be removed.

ljharb: That's what I'm suggesting :-) I don't know of any non-node tools that would bother supporting "exports" but not support condition objects, and it seems like a strange thing to "just in case" support.

ntucker: https://github.com/coinbase/rest-hooks/pull/992

PR #959, opened 2021-06-23
feat: Mark compatibility with upcoming versions

ntucker:
### Motivation

Now that the behavior is known for the upcoming versions, let's make sure all the packages work with it.

### Solution

For those where reliance are not changing we simply add the new versions to the list of support.

### DIFF
--- a/packages/experimental/package.json
+++ b/packages/experimental/package.json
@@ -79,7 +79,7 @@
     "@babel/runtime": "^7.7.2"
   },
   "peerDependencies": {
-    "@rest-hooks/core": "^1.2.1",
+    "@rest-hooks/core": "^1.2.1 || ^2.0.0-0",
     "@rest-hooks/endpoint": "^2.0.0",
     "@rest-hooks/rest": "^3.0.0",
     "@types/react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
--- a/packages/hooks/package.json
+++ b/packages/hooks/package.json
@@ -73,7 +73,7 @@
     "@babel/runtime": "^7.7.2"
   },
   "peerDependencies": {
-    "@rest-hooks/endpoint": "^0.6.1 || ^1.0.0-0",
+    "@rest-hooks/endpoint": "^0.6.1 || ^1.0.0-0 || ^2.0.0-0",
     "@types/react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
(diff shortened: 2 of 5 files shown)


ljharb (approved):
would be good to ensure all peerDeps have identical ranges in either devDeps or deps

ntucker:
not sure i follow

ljharb:
a peer dep generally should also be a dep, or a dev dep. if it's a dev dep, then it's installed for tests. if it's a dep, then it's auto-installed for npm < 7 users and also yarn users, but the peer dep requirement is still hoisted up.

ntucker:
Hmm, didn't realize marking both dependencies and peerDependency was valid. This ensures that if someone does install it above it uses that? Will it still complain if they don't install it?

devDeps don't really apply here since this is a monorepo.

ljharb:
> devDeps don't really apply here since this is a monorepo.

i suppose if all of the packages declare zero dev deps, then that makes sense; that's not the best way to run a monorepo but it's certainly a choice.

> This ensures that if someone does install it above it uses that? Will it still complain if they don't install it? Is there a blog or docs for this behavior?

If it's both a dep and a peer dep, then they can't NOT install it. `npm ls` will complain, however, if they have a version already installed that conflicts/doesn't deduplicate with the version range you've declared as a peer dep.

I'm not aware of any blog or docs, it's just what falls out of the way npm's always worked.

Note that npm 7 automatically installs peer deps, so putting it in deps doesn't add anything (nor hurt) for them, it's only yarn and npm < 7 users that would get the benefit.

ntucker:
What happens if peerDep range is ^1 || ^2 || ^3, and dep is ^3; but user has ^1 installed? Will there only be ^1 or will there be two?

ljharb:
In that case, you'd end up with duplicates, which is why the dep/devDep range must always identically match the peerDep range. `||` is valid everywhere; all semver ranges are valid in all places a semver range is expected.

ntucker:
Case study:

- peerDep range is ^1 || ^2 || ^3
- dep is ^1 || ^2 || ^3.1
- user already has 3.0.1 installed

Does this update the existing 3.0.1 to 3.1?

ljharb:
That's a good question. It clearly *should*, but I'm not sure how NP-smart npm can possibly be here to figure that out.

Thus, I think that would still be a breaking change for a v3.0 user, and that the only thing that makes sense is if the dep range never deviates from the peer dep range.

PR #970, opened 2021-06-27
enhance: Remove extraneous hack for expiresAt

ntucker:
BREAKING CHANGE: Rest Hooks 4 invalidIfStale behavior completely removed

### Motivation

- This was a workaround for particular behavior in Rest Hooks 4, before the modern invalidation schema
- This is no longer triggerable

### Solution

Delete code


ljharb (approved):
If it’s no longer triggerable, is it actually a breaking change?

ntucker:
If you dive into the code and alter stuff this could be problematic. Kinda a moot point since there are already breaking changes lined up

PR #973, opened 2021-06-28
internal: Fix some test async behavior

ntucker:
### Motivation

- Some typos around async behavior found when doing https://github.com/coinbase/rest-hooks/pull/553

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,6 +10,7 @@ const baseConfig = {
     'packages/experimental',
     'packages/legacy/src/resource',
   ],
+  testEnvironment: 'jsdom',
   testURL: 'http://localhost',
 };
 
--- a/packages/hooks/src/__tests__/useCancelling.ts
+++ b/packages/hooks/src/__tests__/useCancelling.ts
@@ -34,6 +34,10 @@ describe('useCancelling()', () => {
       .delay(2000)
       .reply(200, payload2);
   });
+  afterAll(() => {
+    jest.useRealTimers();
+    nock.cleanAll();
+  });
 
   it('should abort when props change and resolve when kept the same', async () => {
     const { result, rerender } = renderHook(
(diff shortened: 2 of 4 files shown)


jest.config.js
```diff
@@ -10,6 +10,7 @@ const baseConfig = {
     'packages/experimental',
     'packages/legacy/src/resource',
   ],
+  testEnvironment: 'jsdom',
```
ljharb, line 13: this means you won't be running tests "like it's node", but more "like it's a browser".

i've found it works well on other projects to run tests twice, once with each testEnvironment. not a blocker ofc.

packages/hooks/src/__tests__/useCancelling.ts
```diff
@@ -34,6 +34,10 @@ describe('useCancelling()', () => {
       .delay(2000)
       .reply(200, payload2);
   });
+  afterAll(() => {
+    jest.useRealTimers();
+    nock.cleanAll();
+  });
```
ljharb, line 40: should this be after _all_, or after _each_?

ljharb (approved):

jest.config.js
ntucker: Jest config returns array. Second one is node like you said.

packages/hooks/src/__tests__/useCancelling.ts
ntucker: It was before all for the setup. Does jest interleave tests?

jest.config.js
ljharb: ah oops, didn't look at the whole file. nice :-)

packages/hooks/src/__tests__/useCancelling.ts
ljharb: given that timers and nock are per-process, using "all" means that tests in the same process (which might or might not be as granular as "same file") could be depending on the timer/nock behavior from other tests to pass. Using "each" ensures every test runs in isolation from the others in this respect.

PR #994, opened 2021-06-30
fix: module build import path

ntucker:
### Motivation

`Could not find module in path: '@rest-hooks/core/legacy/state/actions.js' relative to '/node_modules/@rest-hooks/core/legacy/state/NetworkManager.js'`

The renaming to .js doesn't work with directories

### Solution

Just refer directly to the index file for now until tooling is bettter

Previously `export { useDenormalized } from '@rest-hooks/core/state/selectors';` -> `export { useDenormalized } from './state/selectors.js'`

Now `export { useDenormalized } from '@rest-hooks/core/state/selectors/index';` -> `export { useDenormalized } from './state/selectors/index.js'`

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -4,7 +4,7 @@ export {
   default as reducer,
   initialState,
 } from '@rest-hooks/core/state/reducer';
-export { useDenormalized } from '@rest-hooks/core/state/selectors';
+export { useDenormalized } from '@rest-hooks/core/state/selectors/index';
 export {
   useCache,
   useFetcher,
@@ -19,15 +19,15 @@ export {
   useInvalidateDispatcher,
   useResetter,
   hasUsableData,
-} from '@rest-hooks/core/react-integration';
-export type { ErrorTypes } from '@rest-hooks/core/react-integration';
+} from '@rest-hooks/core/react-integration/index';
+export type { ErrorTypes } from '@rest-hooks/core/react-integration/index';
 export {
   StateContext,
   DispatchContext,
   DenormalizeCacheContext,
 } from '@rest-hooks/core/react-integration/context';
 
-export * from '@rest-hooks/core/state/actions';
+export * from '@rest-hooks/core/state/actions/index';
 export * as actionTypes from '@rest-hooks/core/actionTypes';
 export * from '@rest-hooks/use-enhanced-reducer';
 export * from '@rest-hooks/endpoint';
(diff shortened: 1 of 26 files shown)


ntucker:
@ljharb how do directories work when importing relative in 'exports' members? I know we added .js to make relative file imports work, but what about index.js?

ntucker:
I wonder if there's a way to test against the actual build - the benchmark tests the node path so maybe a pupeteer thing?

ljharb:
I’m not sure what you mean with your question - exports aren’t considered within the package, and if the field is present, standard node resolution does not apply from without.

packages/core/src/index.ts
```diff
@@ -4,7 +4,7 @@ export {
   default as reducer,
   initialState,
 } from '@rest-hooks/core/state/reducer';
-export { useDenormalized } from '@rest-hooks/core/state/selectors';
+export { useDenormalized } from '@rest-hooks/core/state/selectors/index';
```
ljharb, line 7: for this to work with or without the index, you’d need to add this exact specifier to the core package’s exports field

ntucker: The tooling remaps this to a relative import (as it has before). This is how we make typescript happy and allow babel to build various targets

ntucker: this turns into `from './state/selectors/index.js'`

before it was `from './state/selectors.js'` - but selectors is a directory so it didn't work

ntucker:
@ljharb I was able to validate the /lib folder by adding an example that builds with webpack. There's a problem with react-refresh so that PR is not ready, but building for prod works. I also confirmed if I regress the solution here, it prevents that from building.

The current beta release is broke and I would like to fix with this PR

packages/core/src/index.ts
ljharb: ahhhh ok, i misunderstood. when using the package name, in node itself, it goes through "exports" (it's https://www.npmjs.com/package/has-package-self-reference) so this pattern looks like it's supposed to be a specifier in "exports".

ljharb (approved):
as long as consumers won't have to put "/index" then this is just a dev-time wart :-)

ntucker:
currently still don't support subdirectory imports....this is just code in rest hooks.

Was thinking of adding /schema to normalizr tho as it still uses the inherited export * as schema pattern (this was a fork of https://github.com/paularmstrong/normalizr at some point)

PR #996, opened 2021-06-30
feat: Add todo example

ntucker:
### Motivation

Validate exports module build with webpack 5
Provide continually up to date example

### Solution

add `validate-esmodule-browser-build` CI step that builds production but skips minification (to complete faster - we're just checking if it builds)

original: https://github.com/Rest-Hooks/todo-example


.circleci/config.yml
```diff
@@ -71,6 +71,9 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/
+      - run:
+          command: |
+            yarn add -W --dev react@^18.0.0-0 react-dom@^18.0.0-0 react-test-renderer@^18.0.0-0
```
ljharb, line 76: doesn't this mean nothing's being tested on react 17 anymore?

ntucker: This is only the code coverage task. Normal unit tests are run separately. I guess I could consolidate by doing a conditional check on the params to see if it should be reported...just don't want both tests to report themselves

ntucker: unit_tests are above. this is test_coverage

ljharb (approved):

PR #1002, opened 2021-07-02
feat: Allow normalizing entities that are just ids

ntucker:
### Motivation

Use case is to allow for flexible fetch combinations that denormalize with entities retrieved elsewise.

### Solution

New policy is all validation/checking is done on denorm side. Normalize should allow as much as possible. This was originally allowed by the normalizr library.

### DIFF
--- a/packages/normalizr/src/normalize.ts
+++ b/packages/normalizr/src/normalize.ts
@@ -16,18 +16,13 @@ const visit = (
   addEntity: any,
   visitedEntities: any,
 ) => {
-  if (!value || !schema || !['function', 'object'].includes(typeof schema)) {
+  if (!value || !schema) {
     return value;
   }
 
-  if (!schema.normalize || typeof schema.normalize !== 'function') {
-    // serializable
-    if (typeof schema === 'function') {
-      return new schema(value);
-    }
-    const method = Array.isArray(schema) ? arrayNormalize : objectNormalize;
-    return method(
-      schema,
+  if (schema.normalize && typeof schema.normalize === 'function') {
+    if (typeof value !== 'object') return value;
+    return schema.normalize(
       value,
       parent,
       key,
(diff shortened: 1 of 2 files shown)


packages/normalizr/src/normalize.ts
```diff
@@ -16,15 +16,18 @@ const visit = (
...
 ) => {
-  if (!value || !schema || !['function', 'object'].includes(typeof schema)) {
+  if (!value || !schema) {
     return value;
   }
 
+  // serializable
+  if (typeof schema === 'function' && !schema.normalize) {
+    return new schema(value);
+  }
```
ljharb, line 26: you may want to check for `schema.prototype` as well, unless the intention is for a non-constructible schema to throw

ljharb (approved):

packages/normalizr/src/normalize.ts
ntucker: what would be a valid check that something is nonconstructable? BigInt.prototype i'm getting {}

ljharb: The prototype check covers concise methods and arrow functions. There’s no better way to weed out things like BigInt and Symbol, though.

ntucker: Any insight into why BigInt is designed like that? lol. makes things hard.

Anyway, landed this for now since those aren't technical supported so they should throw errors; an improvement would be replacing those errors with a more helpful message, but that is not a regression so can be done for a future time.

ntucker: Really i'd like to just call the function if it doesn't want new, but it doesn't seem like there's a way to really make that work.

ljharb: tbh the most straightforward pattern is probably just "call a function", and they can do `() => new x()` if that's what they want.

PR #1038, opened 2021-07-15
fix: Don't inline normalizr into cjs

ntucker:
BREAKING CHANGE: Requires node 12

### Motivation

- Keep the same entities
- class should inherit from class or new breaks

### Solution

- rollup is configured to keep external the packages in peerdeps, but since normalizr is transitive dep it was being inlined. it's really best to only import from things explicitly stated, so we move it to only import from endpoint
- target node 12


ljharb (approved):
the diff seems fine but i have no idea how this newly requires node 12, nor how this is a breaking change ¯\\\_(ツ)\_/¯

ntucker:
oops, commited but didn't push the node12 stuff

ljharb:
cool, still LGTM but now it all makes sense

PR #1046, opened 2021-07-17
enhance: Experimental fetcher resolves before react render

ntucker:
### Motivation

Part of https://github.com/coinbase/rest-hooks/issues/760

This makes little difference in React 18 since renders are batched; however in React < 18, this means that code after promise resolution will be executed before react renders - allowing actions that need to take place as a result of successful fetch. For example navigating off a deleted page after delete.

### Solution

Using promise chaining allows us to order microtasks so the resolution of useFetcher()'s promise happens before dispatch of receive.

This has many consequences for existing behaviors so we are opting in via the new fetch mechanism.

It's now recommended to wrap all fetches in act when testing like so:

```ts
    await act(async () => {
      await result.current.fetch(ComplexResource.detail(), {
        id: '5',
      });
    });
```

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -99,8 +99,28 @@ export default class NetworkManager implements Manager {
   protected handleFetch(action: FetchAction, dispatch: Dispatch<any>) {
     const fetch = action.payload;
     const { key, throttle, resolve, reject } = action.meta;
-    const deferedFetch = () =>
-      fetch()
+
+    const deferedFetch = () => {
+      let promise = fetch();
+      const resolvePromise = (
+        promise: Promise<string | number | object | null>,
+      ) =>
+        promise
+          .then(data => {
+            resolve(data);
+            return data;
+          })
+          .catch(error => {
+            reject(error);
+            throw error;
+          });
+      // schedule non-throttled resolutions in a microtask before receive
+      // this enables users awaiting their fetch to trigger any react updates needed to deal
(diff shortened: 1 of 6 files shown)


ntucker:
@sweetro

packages/core/src/state/NetworkManager.ts
```diff
@@ -123,14 +143,20 @@ export default class NetworkManager implements Manager {
...
+    };
+
     if (throttle) {
-      promise = this.throttle(key, deferedFetch);
+      return this.throttle(key, deferedFetch)
+        .then(data => resolve(data))
+        .catch(error => reject(error));
     } else {
-      promise = deferedFetch();
+      return deferedFetch().catch(() => {});
```
ljharb, line 158:
```suggestion
      return deferedFetch().catch();
```

packages/core/src/state/__tests__/networkManager.ts
```diff
@@ -92,6 +92,7 @@ describe('NetworkManager', () => {
       body: { id: 5, title: 'hi' },
       throttle: false,
     });
+    (fetchRejectAction.meta.promise as any).catch(e => {});
```
ljharb, line 95:
```suggestion
    (fetchRejectAction.meta.promise as any).catch();
```

ljharb (approved):

packages/core/src/state/NetworkManager.ts
ntucker: Removing the function makes the exception fallthrough and the tests fail.

PR #1048, opened 2021-07-18
feat: useController()

ntucker:
### Motivation

Consolidate, simplify hooks

Consistent interface between managers and hooks

Global referential equality available everywhere (managers and updaters)

Simplify and consolidate TTL and error concepts

Less code in hooks = less work on rendering leaf nodes

Icing on cake: ez migration to EndpointInterface and flexible args support for hooks

Future breaking changes can allow ez migration with version strings sent to `useController('v2')`

### Solution

We're introducing fetch, invalidate, resetEntireStore today into experimental.

Full future plan:

```ts
const controller = useController();

// actions
controller.fetch(UserResource.detail(), { id }) // sideEffects means no throttle, otherwise throttle
controller.receive(payload, UserResource.detail(), { id })
controller.invalidate(UserResource.detail(), { id })
controller.resetEntireStore()
controller.subscribe(UserResource.detail(), { id })
controller.unsubscribe(UserResource.detail(), { id })
// posisble new
controller.abort(UserResource.detail(), { id }) // only aborts if in flight
// note: to force fetch of sideEffect: undefined - call abort first
// this should enable good offline/online managers

// retrieval
const state = useContext(StateContext);
const [value, expiresAt] = controller.getResponse(state, UserResource.detail(), { id })
const error = controller.getError(state, UserResource.detail(), { id })
```

### DIFF
--- a/packages/experimental/src/useController.ts
+++ b/packages/experimental/src/useController.ts
@@ -1,0 +1,70 @@
+import { EndpointInterface } from '@rest-hooks/endpoint';
+import { DispatchContext } from '@rest-hooks/core';
+import { useContext, useCallback } from 'react';
+import { actionTypes } from '@rest-hooks/core';
+
+import createFetch from './createFetch';
+import { UpdateFunction } from './types';
+
+/**
+ * Imperative control of Rest Hooks store
+ * @see https://resthooks.io/docs/api/useController
+ */
+export default function useController(throttle = false): Controller {
+  const dispatch = useContext(DispatchContext);
+
+  // TODO: elevate into CacheProvider and simply useContext
+  const fetch = useCallback(
+    <E extends EndpointInterface>(
+      endpoint: E,
+      ...args: readonly [...Parameters<E>]
+    ) => {
+      const action = createFetch(endpoint, {
+        args,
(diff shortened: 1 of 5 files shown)


packages/experimental/src/useController.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    [dispatch, throttle],
+  );
+
+  const invalidate = useCallback(
+    <E extends EndpointInterface>(
+      endpoint: E,
+      ...args: readonly [...Parameters<E>] | readonly [null]
+    ) =>
+      args?.[0] !== null
```
ljharb, line 38:
```suggestion
      args[0] !== null
```
it's impossible that `args` is not an array

ljharb (approved):

packages/experimental/src/useController.ts
ntucker: Are there no out of bounds checks in Javascript?

ljharb: Of course not, arrays are just objects. Either way, `?.` does nothing different unless the LHS is null or undefined.

PR #1083, opened 2021-07-31
enhance: Hide throttled fetches from devtools

### LINKED ISSUE #1071: Redux-devtools: Don't show throttled fetches
These can end up being kinda spammy

ntucker:
Fixes #1071 .

### Motivation

For nested members, TTL eliminates fetches. However, with siblings fetches must be throttled. If this happens, devtools fills up with duplicate fetches that have no actual impact on the state.

### Solution

- NetworkManager passes through all non-throttled fetch actions
- Place devtools after networkmanager
- Update reducer tracking to ensure NM is still installed

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -43,13 +43,16 @@ export default class NetworkManager implements Manager {
           switch (action.type) {
             case FETCH_TYPE:
               this.handleFetch(action, dispatch);
-              // This is the only case that causes any state change
-              // It's important to intercept other fetches as we don't want to trigger reducers during
-              // render - so we need to stop 'readonly' fetches which can be triggered in render
-              if (action.meta.optimisticResponse !== undefined) {
-                return next(action);
+              // Eliminate throttled fetches from future middlewares + reducer
+              // TODO: Maybe make throttling its own middleware?
+              if (action.meta.throttle && action.key in this.fetched) {
+                return Promise.resolve();
               }
-              return Promise.resolve();
+              // helps detect missing NetworkManager when using redux
+              if (process.env.NODE_ENV !== 'production') {
+                action.meta.nm = true;
+              }
+              return next(action);
             case RECEIVE_TYPE:
               // only receive after new state is computed
               return next(action).then(() => {
--- a/packages/rest-hooks/src/react-integration/provider/index.ts
+++ b/packages/rest-hooks/src/react-integration/provider/index.ts
@@ -17,8 +17,16 @@ CacheProvider.defaultProps = {
   ],
 };
 /* istanbul ignore next */
-if (process.env.NODE_ENV !== 'production')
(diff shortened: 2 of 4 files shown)


packages/core/src/state/NetworkManager.ts
```diff
@@ -43,13 +43,16 @@ export default class NetworkManager implements Manager {
...
               this.handleFetch(action, dispatch);
-              // This is the only case that causes any state change
-              // It's important to intercept other fetches as we don't want to trigger reducers during
-              // render - so we need to stop 'readonly' fetches which can be triggered in render
-              if (action.meta.optimisticResponse !== undefined) {
-                return next(action);
+              // Eliminate throttled fetches from future middlewares + reducer
+              // TODO: Maybe make throttling its own middleware?
+              if (action.meta.throttle && action.key in this.fetched) {
+                return Promise.resolve();
```
ljharb, line 49: why not make the function `async` so you don't need the explicit promise stuff?

packages/rest-hooks/src/react-integration/provider/index.ts
```diff
@@ -17,8 +17,16 @@ CacheProvider.defaultProps = {
...
 /* istanbul ignore next */
-if (process.env.NODE_ENV !== 'production')
+if (process.env.NODE_ENV !== 'production') {
   CacheProvider.defaultProps.managers.unshift(new DevToolsManager());
+  if (CacheProvider.defaultProps.managers.length > 1) {
+    // swap so devtools is right after networkmanager
+    [
+      CacheProvider.defaultProps.managers[1],
+      CacheProvider.defaultProps.managers[0],
+    ] = CacheProvider.defaultProps.managers.slice(0, 2);
```
ljharb, line 27:
```suggestion
    ([
      CacheProvider.defaultProps.managers[1],
      CacheProvider.defaultProps.managers[0],
    ] = CacheProvider.defaultProps.managers.slice(0, 2));
```
are the parens only required for setting variables, not object properties? this is a super niche/rare way to use destructuring, so i'm not certain :-)

ljharb (approved):

packages/rest-hooks/src/react-integration/provider/index.ts
ntucker: sans parens is what prettier output. i think object properties is needed for the parse to distinguish from scope blocks. but with [] this is not necessary.

packages/core/src/state/NetworkManager.ts
ntucker: Targetting old versions using async/await transpilation added a ton of code, so I tried avoiding it. Even now it's kinda meh in this case since it would just be 'return'

ljharb: Yeah the generator transform of async await sucks; a promises one would be much better. Fair enough.

PR #1084, opened 2021-08-01
enhance: Improve reducer error visibility

### LINKED ISSUE #1073: Errors thrown during user-run update should be surfaced
Currently not rejecting the main promise, or showing up in console logs. Only shows itself after inspecting the state

ntucker:
Fixes #1073 .

### Motivation

Users are liable to write functions that throw errors. In these cases

- be graceful
- be loud

### Solution

- all errors in reducer's receive are console.error in devmode
- errors during Endpoint.update() are always console.error instead of reducing.
  - the state will still update with whatever it had

### DIFF
--- a/packages/core/src/state/reducer.ts
+++ b/packages/core/src/state/reducer.ts
@@ -1,5 +1,10 @@
 import { normalize } from '@rest-hooks/normalizr';
-import { ActionTypes, State, ReceiveAction } from '@rest-hooks/core/types';
+import {
+  ActionTypes,
+  State,
+  ReceiveAction,
+  ReceiveMeta,
+} from '@rest-hooks/core/types';
 import { createReceive } from '@rest-hooks/core/state/actions/index';
 import {
   RECEIVE_TYPE,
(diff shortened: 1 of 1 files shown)


packages/core/src/state/reducer.ts
```diff
@@ -80,15 +87,28 @@ export default function reducer(
           ...state.results,
           [action.meta.key]: result,
         };
-        results = applyUpdatersToResults(results, result, action.meta.updaters);
-        if (action.meta.update) {
-          const updaters = action.meta.update(
+        try {
```
ntucker, line 90: the diff is weird here, this is just an indentation due to adding `try {`

ljharb: looks better when whitespace changes are hidden, at least

ljharb (approved):

packages/core/src/state/reducer.ts
ntucker: hmm, are you not using the website? is the github desktop or some other method?

ljharb: The website has a “hide white space” node, under the gear icon (since a couple years ago). You can also append ?w=1 to any diff url (since always)

PR #1085, opened 2021-08-01
fix: RESET clears inflight fetches

### LINKED ISSUE #1072: RESET should prevent resolve/rejection of inflight promises
In resolve/reject function, check for existence in inflight requests *or* using abortcontroller.

Note: this means non-throttle requests will need some form of bookkeeping

ntucker:
Fixes #1072 .

### Motivation

[reset](https://resthooks.io/docs/api/useResetter) clears inflight promises, not just state already established.

This is critical when using RESET for auth logout as the subsequent promises will likely reject with auth errors.

#### Visible Behavior

- unmounting CacheProvider with inflight requests will not attempt to dispatch after unmount causing react error
  - promises are neither resolved nor rejected
- upon reset, all inflight requests will not dispatch receives
  - promises still reject so external listeners know (abort signals do this as well)
- useResource(), useRetrive() will re-issue needed fetches upon reset so they never end up loading infinitely
  - this only triggers *after* commit of reset action so users have a chance to unmount those components if they are no longer relevant (like doing a url redirect from an unauthorized page)

#### Implementation details
- Reset action includes date. This is stored in state as 'lastReset'. This is used to trigger refetches in useRetrieve()
  - NetworkManager now relies on getState() - using it to find the lastReset time. This determines whether to dispatch actions after a fetch completes
  - Since reset is a 'singleton action', tracking the last time allows us to simply compare current fetch times with last reset to know whether to process their dispatches.
- No special test code for fetch date - instead remove it from snapshots so tests don't break on every run

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -3,7 +3,12 @@ import {
   Middleware,
   Dispatch,
 } from '@rest-hooks/use-enhanced-reducer';
-import { FetchAction, ReceiveAction, Manager } from '@rest-hooks/core/types';
+import {
+  FetchAction,
+  ReceiveAction,
+  Manager,
+  State,
+} from '@rest-hooks/core/types';
 import {
   RECEIVE_TYPE,
   FETCH_TYPE,
@@ -29,6 +34,7 @@ export default class NetworkManager implements Manager {
   declare readonly dataExpiryLength: number;
   declare readonly errorExpiryLength: number;
   protected declare middleware: Middleware;
+  private lastClear: Date | number = -Infinity;
 
   constructor(dataExpiryLength = 60000, errorExpiryLength = 1000) {
     this.dataExpiryLength = dataExpiryLength;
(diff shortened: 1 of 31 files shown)


ntucker:
Adding tests now

packages/core/src/state/NetworkManager.ts
```diff
@@ -29,6 +34,7 @@ export default class NetworkManager implements Manager {
   declare readonly dataExpiryLength: number;
   declare readonly errorExpiryLength: number;
   protected declare middleware: Middleware;
+  private lastClear = new Date(-1);
```
ljharb, line 37: seems like maybe doing `new Date(NaN)` would be a clearer choice?

ljharb (approved):

packages/core/src/state/NetworkManager.ts
ntucker: new Date(NaN) < new Date(0) === false

Whatever is on left hand side I want to always be less than right hand side.

ljharb: in that case, just `-Infinity`? ie, not a date object at all

ntucker: Oh nice, didn't realize that had such a property 👍

PR #1132, opened 2021-08-21
pkg: Latest react 18 prerelease

ntucker:
### Motivation

Ensure compatibility from the latest

### Solution

They started using dates, but put them after the hashes, which makes sorting not work correctly - so the only way to get latest pre-release is by using the next tag. This sadly means upon full release we'll have to change the specified version.


ntucker:
The first commit is from another pr so just look at the last one

ljharb (approved):
not sure why "next" is needed, `^18.0.0-0` should always point to the latest react 18 prerelease - oh wait, i read your OP. that's really dumb of the react team.

PR #1149, opened 2021-08-25
enhance: Back to avoiding reducer for fetch

ntucker:
### Motivation

Turns out react doesn't always short circuit reducers even with ===. It seemed like there were sometimes jank around loading.
Prior: https://github.com/coinbase/rest-hooks/pull/1083 https://github.com/coinbase/rest-hooks/pull/1130

### Solution

Back to avoiding reducer - but we introduce `skipLogging` which is sent to devtools to know when to skip logging fetches. This adds an API surface area but is ultimately a cleaner code path.

Effectively https://github.com/coinbase/rest-hooks/pull/1083  is undone in favor of this solution

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -8,6 +8,7 @@ import {
   ReceiveAction,
   Manager,
   State,
+  ActionTypes,
 } from '@rest-hooks/core/types';
 import {
   RECEIVE_TYPE,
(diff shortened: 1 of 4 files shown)


packages/core/src/state/NetworkManager.ts
```diff
@@ -101,6 +92,10 @@ export default class NetworkManager implements Manager {
     };
   }
 
+  skipLogging(action: ActionTypes) {
+    return action.type === FETCH_TYPE && action.meta.key in this.fetched;
```
ljharb, line 96: will the key ever possibly be `toString` or something else on Object.prototype?

ljharb (approved):

packages/core/src/state/NetworkManager.ts
ntucker: unlikely but possible since this is user set

ntucker: updated with hasownproperty

PR #1171, opened 2021-08-28
internal: Dependabot - ignore react

ntucker:
### Motivation

Being on next means this gets spammy


ljharb (approved):
this seems like it'll come back to bite you in the future tho :-)

ntucker:
I doubt I'll miss any react upgrades :P I'll get rid of it once 18 is final release

PR #1183, opened 2021-08-31
enhance: validate during denormalization

ntucker:
### Motivation

Entities can come from different endpoints than they are retrieved from.

This enables cases where you might want to send only partial fields for a short list, and then send the rest in longer. In this case you don't want to use partial entities.

```ts
class ArticlePreview extends Entity {
  readonly id: string = '';
  readonly title: string = '';

  pk() { return this.id; }
  static key() { return 'Article'; }
}
const articleList = new Endpoint(fetchList, { schema: [ArticlePreview] });

class ArticleFull extends ArticlePreview {
  readonly createdAt: Date = new Date(0);

  static schema = {
    createdAt: Date,
  }

  static validate(processedEntity) {
    if (!Object.hasOwn(processedEntity, 'content')) return 'Missing content';
  }
}
const articleDetail = new Endpoint(fetchDetail, { schema: ArticleFull });
```

### Solution

Thus, we can specify which fields are needed in `validate()` and if a validation error occurs, we mark that result as needing to do a full fetch (aka deleted).

This changes validate() from throwing errors to returning errormessages. This is a breaking change for this method, however validate is currently unreleased, so it is not necessary to document this in the release notes.

Infinite loops are not possible as the moment validation fails in normalization, it is cut short due throw errors instead of suspending.

### DIFF
--- a/packages/normalizr/src/entities/Entity.ts
+++ b/packages/normalizr/src/entities/Entity.ts
@@ -183,7 +183,12 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
     ) {
       return id;
     }
-    this.validate(processedEntity);
+    const errorMessage = this.validate(processedEntity);
+    if (errorMessage) {
+      const error = new Error(errorMessage);
+      (error as any).status = 400;
+      throw error;
+    }
     visitedEntities[entityType][id].push(input);
 
     Object.keys(this.schema).forEach(key => {
@@ -204,7 +209,7 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
     return id;
   }
 
-  protected static validate(processedEntity: any) {
+  protected static validate(processedEntity: any): string | undefined {
     /* istanbul ignore else */
     if (
       process.env.NODE_ENV !== 'production' &&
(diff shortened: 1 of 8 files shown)


packages/normalizr/src/entities/Entity.ts
```diff
@@ -331,7 +330,8 @@ First three members: ${JSON.stringify(input.slice(0, 3), null, 2)}`;
     this: T,
     input: Readonly<Partial<AbstractInstanceType<T>>>,
     unvisit: schema.UnvisitFunction,
-  ): [AbstractInstanceType<T>, boolean, boolean] {
+  ): [denormalized: AbstractInstanceType<T>, found: boolean, suspend: boolean] {
```
ljharb, line 333:
```suggestion
  ): [denormalized: AbstractInstanceType<T>, found: boolean, suspend: boolean] as const {
```
?

ljharb (approved):

packages/normalizr/src/entities/Entity.ts
ntucker: As const is for values but this is a type Def so it's already a tuple

ljharb: ah ok, TIL

PR #1194, opened 2021-09-01
internal: Add github example app

ntucker:


examples/github-app/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "github-app",
+  "version": "0.1.0",
```
ljharb, line 3: this should probably have `"private": true`, so it's not published by accident.

ljharb (approved):

examples/github-app/package.json
ntucker: Line 21 has that

ljharb: ah ok, buried in there :-)

ntucker: Can't make it easy or it wouldn't be as fun 😆

PR #1215, opened 2021-09-05
feat: Add @rest-hooks/graphql

### LINKED ISSUE #361: Roadmap for GraphQL
I've seen mentions of GraphQL in the issue tracker and some other places throughout this repository.

I'd like to know what would take to get **rest-hooks** to an architecture/API which would support GraphQL as a first-class citizen, e.g.

- Do the internal components (e.g. the caching mechanism) need to change?
- Does the API surface need to change?

Another pertaining question would be if you're planning to implement a client directly or use an existing third-party library. I infer the latter might be possible since I've seen https://github.com/coinbase/rest-hooks/issues/201, or maybe it already is possible currently through some user-land implementation and I'm not aware of it.

Overall, the point of this issue is to

- List what's missing and what is the timeline for it
- Point at some modules of interest in the code base
- What good ideas for an implementation would be
- What the community could contribute to make GraphQL viable

I'm personally interested in using rest-hooks for everything (REST, WebSockets, ...) since I really like the design and how modern it is.

ntucker:
Fixes #361 .

## Motivation

GraphQL made easy with rest hooks.

## Solution

- This is a prerelease so we're not doing the same rigor of testing. However, this was validated in a real world project.
- Error handling not as elegant as @rest-hooks/rest yet.
- First version requires explicit schema passing to use. Future could interpolate from the querystring itself.

### Define GraphQL endpoint

```typescript
const gql = new GQLEndpoint('https://nosy-baritone.glitch.me');
```

### Simple TypeScript definition

```typescript
class User extends GQLEntity {
  readonly name: string = '';
  readonly email: string = '';
}
```

### Write type-safe queries

```typescript
const userList = gql.query(
  `{
    users {
      id
      name
      email
      }
    }`,
  { users: [User] },
);

const userDetail = gql.query<{ name: string }>(
  `query UserDetail($name: String!) {
    user(name: $name) {
      id
      name
      email
    }
  }`,
  { user: User },
);
```

### One line data-hookup

```tsx
const { user } = useResource(userDetail, { name: 'Fong' });
return (
  <>
    <h2>{user.name}</h2>
    <p>{user.email}</p>
  </>
);
```

### Mutations

```ts
const gql = new GQLEndpoint('https://graphql.org/swapi-graphql');

const createReview = gql.mutation<{
  ep: string;
  review: { stars: number; commentary: string };
}>(
  `mutation CreateReviewForEpisode($ep: Episode!, $review: ReviewInput!) {
    createReview(episode: $ep, review: $review) {
      stars
      commentary
    }
  }`,
  { createReview: Review },
);
```

```tsx
const createReview = useFetcher(createReview);
return <ReviewForm onSubmit={data => createReview(data)} />;
```

### DIFF
--- a/packages/graphql/rollup.config.js
+++ b/packages/graphql/rollup.config.js
@@ -1,0 +1,72 @@
+import commonjs from 'rollup-plugin-commonjs';
+import resolve from 'rollup-plugin-node-resolve';
+import babel from 'rollup-plugin-babel';
+import json from 'rollup-plugin-json';
+import { terser } from 'rollup-plugin-terser';
+import filesize from 'rollup-plugin-filesize';
+import replace from 'rollup-plugin-replace';
+
+import pkg from './package.json';
+
+const dependencies = Object.keys(pkg.dependencies)
+  .concat(Object.keys(pkg.peerDependencies))
+  .concat(['@rest-hooks/endpoint', '@rest-hooks/normalizr'])
+  .filter(dep => !['@babel/runtime'].includes(dep));
+
+const extensions = ['.js', '.ts', '.tsx', '.mjs', '.json', '.node'];
+const nativeExtensions = ['.native.ts', ...extensions];
+process.env.NODE_ENV = 'production';
+process.env.RESOLVER_ALIAS = '{"@rest-hooks/graphql":"./src"}';
+
+function isExternal(id) {
+  const ret = dependencies.includes(id);
+  if (!ret) {
+    for (const dep of dependencies) {
+      if (id.startsWith(dep)) return true;
+    }
+  }
+  return ret;
+}
+
+const configs = [];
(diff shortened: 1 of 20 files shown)


packages/graphql/rollup.config.js
```diff
@@ -0,0 +1,72 @@
...
+const nativeExtensions = ['.native.ts', ...extensions];
+process.env.NODE_ENV = 'production';
+process.env.RESOLVER_ALIAS = '{"@rest-hooks/graphql":"./src"}';
+
+function isExternal(id) {
+  const ret = dependencies.includes(id);
+  if (!ret) {
+    for (const dep of dependencies) {
+      if (id.startsWith(dep)) return true;
+    }
```
ljharb, line 26:
```suggestion
    return dependencies.some((dep) => id.startsWith(dep));
```

ljharb (approved):

PR #1271, opened 2021-09-19
feat: Send controller to middlewares

ntucker:
- applyManager()
- controller in middlewares

### Motivation

Controller provides a cleaner and more consistent interface to the store. This also makes the interaction points more discoverable as any TypeScript supporting editor will autocomplete (even if the user is using javascript).

### Solution

Added to middleware api

```ts
export interface MiddlewareAPI<
  R extends React.Reducer<any, any> = React.Reducer<any, any>,
> {
  getState: () => React.ReducerState<R>;
  dispatch: Dispatch<R>;
  controller: Controller;
}
```

This is adapted to still be redux-middleware compatible via the `applyManager()` method. This turns a manager into a redux-middleware

Due to the way typescript handles function parameter typing, this will not break compatibility with any existing middlewares since it is only an addition.

Also, by not using controller yet, those who aren't using applyManager() in redux case will not have to worry about it breaking.

```tsx
import type { Manager, Middleware } from '@rest-hooks/core';
import type { EndpointInterface } from '@rest-hooks/endpoint';

export default class StreamManager implements Manager
{
  protected declare middleware: Middleware;
  protected declare websocket: Websocket;
  protected declare endpoints: Record<string, EndpointInterface>;

  constructor(url: string, endpoints: Record<string, EndpointInterface>) {
    this.websocket = new Websocket(url);
    this.endpoints = endpoints;

    this.middleware = ({ controller, getState }) => {
      this.websocket.onmessage = (event) => {
        controller.receive(this.endpoints[event.type], ...event.args, event.data);
      }
      return (next) => async (action) => next(action);
    }
  }

  cleanup() {
    this.websocket.close();
  }

  getMiddleware<T extends StreamManager>(this: T) {
    return this.middleware;
  }
}
```

#### CacheStore

[`<CacheProvider />`](https://resthooks.io/docs/api/CacheProvider) is now split into two components. The outer processes all the input parameters creating reverentially stable props for `<CacheStore />`. `<CacheStore />` only contains the minimal pieces needed to handle store changes as to minimize code run when the store updates, which is frequently. This separation also places a clean boundary to make it slightly easier to read.

### DIFF
--- a/packages/core/src/controller/Controller.ts
+++ b/packages/core/src/controller/Controller.ts
@@ -14,7 +14,7 @@ import type { DenormalizeCache } from '@rest-hooks/normalizr';
 type RHDispatch = (value: ActionTypes) => Promise<void>;
 
 interface ConstructorProps {
-  dispatch: RHDispatch;
+  dispatch?: RHDispatch;
   globalCache?: DenormalizeCache;
 }
 
@@ -27,12 +27,12 @@ export default class Controller {
   declare readonly globalCache: DenormalizeCache;
 
   constructor({
-    dispatch,
+    dispatch = () => Promise.reject('dispatch not set'),
     globalCache = {
       entities: {},
       results: {},
     },
-  }: ConstructorProps) {
+  }: ConstructorProps = {}) {
     this.dispatch = dispatch;
     this.globalCache = globalCache;
   }
--- a/packages/test/src/providers.tsx
+++ b/packages/test/src/providers.tsx
@@ -1,4 +1,10 @@
-import { State, reducer, CacheProvider, Manager } from '@rest-hooks/core';
+import {
+  State,
+  reducer,
+  CacheProvider,
+  Manager,
+  Controller,
+} from '@rest-hooks/core';
 import {
   ExternalCacheProvider,
   PromiseifyMiddleware,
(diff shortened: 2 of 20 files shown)


packages/core/src/controller/Controller.ts
```diff
@@ -32,8 +32,8 @@ export default class Controller {
       entities: {},
       results: {},
     },
-  }: ConstructorProps) {
-    this.dispatch = dispatch;
+  }: ConstructorProps = {}) {
+    this.dispatch = dispatch ?? (() => Promise.reject('dispatch not set'));
```
ljharb, line 36:
```suggestion
    this.dispatch = dispatch ?? () => Promise.reject('dispatch not set');
```
altho it seems like you could do `dispatch = () => Promise.reject('dispatch not set')` in the constructor args, unless you care about supporting `null`

ljharb (approved):

packages/core/src/controller/Controller.ts
ntucker: is this just removing the parens? This was required by the lint parser

ljharb: yes, and that's very weird but fine i guess.

the real comment was about using defaulting vs `??`, unless the type allows for `null`

ntucker: ?? works on undefined right? the parameter is optional

ljharb: sure, but it also works on null.

if you don't care about null, it seems a bit cleaner to do it declaratively in the constructor params as a default instead of imperatively in the constructor body with `??`.

packages/test/src/providers.tsx
```diff
@@ -23,18 +29,21 @@ let makeExternalCacheProvider: (
 try {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const { createStore, applyMiddleware, combineReducers } = require('redux');
+  const applyManager =
```
ntucker, line 32: To maintain compatibility

packages/core/src/controller/Controller.ts
ntucker: Hmm, in that case you're constructing the function every call no matter what. That's probably pretty minor, but I also don't see a huge gain from using default.

ljharb: i'm pretty sure it's only evaluated when the default is triggered:
```js
function f(x = console.log('*')) {} [f(), f(null), f({})]
```
only logs once.

ntucker: ahh, that actually makes sense now that i think about it.

PR #1272, opened 2021-09-19
internal: Don't fail coverage if coveralls is down

ntucker:
### Motivation

Coveralls has been down for days blocking progress. Sometimes their services has issues as well - these shouldn't block progress.

### Solution

We will fail on test:coverage, but the actual sending to coveralls we don't fail


ljharb (approved):
Seems fine, but I’d suggest codecov

PR #1278, opened 2021-09-20
feat: Maintain referential stability when deps are used

ntucker:
### Motivation

When using deps list, we want to inline the function so it won't maintain stability. However, without deps list it should be based on the function.


notwillk (approved):
Yay, more tests!

PR #1290, opened 2021-09-23
feat: Add controller.getResponse()

ntucker:
### Motivation

Continuing https://github.com/coinbase/rest-hooks/pull/1239

This extracts the core logic into a more accessible piece.
- Easier implementations for various frameworks
- Easier implementations for different hookups like context selectors
- Programmatic access to same values from hooks, which can be used for procedural pieces

### Solution

```tsx
const { data, suspend, expiresAt } = controller.getResponse(UserResource.detail(), { id }, state)
const error = controller.getError(UserResource.detail(), { id }, state)
```

useDenormalized logic moved into controller.getResponse(). Except it's based on [endpoints](https://resthooks.io/docs/getting-started/endpoint) instead of shapes. Because of this useDenormalized first constructs an endpoint with the adapter `shapeToEndpoint()`

### DIFF
--- a/packages/core/src/controller/Controller.ts
+++ b/packages/core/src/controller/Controller.ts
@@ -1,15 +1,25 @@
-import type { ActionTypes } from '@rest-hooks/core/types';
+import type { ActionTypes, State } from '@rest-hooks/core/types';
 import type { EndpointInterface, ResolveType } from '@rest-hooks/endpoint';
 import createInvalidate from '@rest-hooks/core/controller/createInvalidate';
 import createFetch from '@rest-hooks/core/controller/createFetch';
 import createReset from '@rest-hooks/core/controller/createReset';
+import { selectMeta } from '@rest-hooks/core/state/selectors/index';
 import createReceive from '@rest-hooks/core/controller/createReceive';
+import { NetworkError, UnknownError } from '@rest-hooks/core/types';
 import {
   createUnsubscription,
   createSubscription,
 } from '@rest-hooks/core/controller/createSubscription';
 import type { EndpointUpdateFunction } from '@rest-hooks/core/controller/types';
-import type { DenormalizeCache } from '@rest-hooks/normalizr';
+import {
+  denormalize,
+  DenormalizeCache,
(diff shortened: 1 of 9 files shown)


packages/core/src/controller/Controller.ts
```diff
@@ -159,16 +169,158 @@ export default class Controller {
...
+    let expiresAt = selectMeta(state, key)?.expiresAt;
+    // fallback to entity expiry time
+    if (!expiresAt) {
+      // expiresAt existance is equivalent to cacheResults
+      if (found) {
+        // oldest entity dictates age
+        expiresAt = Infinity;
+        // using Object.keys ensures we don't hit `toString` type members
+        Object.keys(resolvedEntities).forEach(key =>
+          Object.keys(resolvedEntities[key]).forEach(pk => {
```
ljharb, line 273:
```suggestion
        Object.entries(resolvedEntities).forEach(([key, entities]) =>
          entities.forEach(pk => {
```

packages/core/src/controller/Controller.ts
```diff
@@ -159,16 +169,158 @@ export default class Controller {
...
+  if (schema && (typeof schema === 'object' || typeof schema === 'function')) {
+    const nestedSchema =
+      'schema' in schema ? (schema.schema as Record<string, Schema>) : schema;
+    if (typeof nestedSchema === 'function') {
+      return schemaHasEntity(nestedSchema);
+    }
+    return Object.values(nestedSchema).reduce(
+      (prev, cur) => prev || schemaHasEntity(cur),
+      false,
+    );
```
ljharb, line 321:
```suggestion
    return Object.values(nestedSchema).some(x => schemaHasEntity(x));
```

ljharb (approved):

packages/core/src/controller/Controller.ts
ntucker: object doesn't seem to have a builtin foreach...

ljharb: oh duh, my bad on that line then

PR #1308, opened 2021-09-29
enhance: Improve compatibility with SSR and React Native

ntucker:
### Motivation

CacheProvider is needed in all installs, so ensure it works in all compatible cases.

### Solution

- .native extension allows for RN specific version that uses <Text>
- Don't use <Suspense/> in React 16, 17 when doing SSR - as it will show fallback which will likely block entire app

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -112,14 +112,10 @@
   },
   "peerDependencies": {
     "@types/react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
-    "react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
-    "react-native": ">=0.59"
+    "react": "^16.8.4 || ^17.0.0 || ^18.0.0-0"
   },
   "peerDependenciesMeta": {
     "@types/react": {
-      "optional": true
-    },
-    "react-native": {
       "optional": true
     }
   }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -52,8 +52,8 @@
     "./rest_hooks_logo_and_text.svg"
   ],
   "scripts": {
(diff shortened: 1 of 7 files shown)


packages/core/package.json
```diff
@@ -112,11 +112,15 @@
   },
   "peerDependencies": {
     "@types/react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
-    "react": "^16.8.4 || ^17.0.0 || ^18.0.0-0"
+    "react": "^16.8.4 || ^17.0.0 || ^18.0.0-0",
+    "react-native": ">=0.59"
```
ljharb, line 116: This is a breaking change for any package manager that doesn’t understand peerDependenciesMeta, and will force those react web consumers to install react-native. I very much doubt this is a desired approach.

I’d suggest a separate package for RN users, so non-RN users don’t have to interact with RN requirements.

packages/core/rollup.config.js
```diff
@@ -15,7 +15,7 @@ const dependencies = Object.keys(pkg.dependencies)
       !['@rest-hooks/use-enhanced-reducer', '@babel/runtime'].includes(dep),
   );
 
-const extensions = ['.js', '.ts', '.tsx', '.mjs', '.json', '.node'];
+const extensions = ['.js', '.ts', '.tsx', '.mjs', '.json', '.node', 'jsx'];
```
ljharb, line 18: why does jsx omit the dot and the others include it?

packages/core/package.json
ntucker: "force those react web consumers to install react-native" - isn't the only version of npm that does this npm 7+, which does understand optional?

packages/core/rollup.config.js
ntucker: ack, good catch

packages/core/package.json
ljharb: By force, i mean that `npm ls` will exit nonzero, which is true in every version of npm. Making someone’s dep graph invalid is a problem even if it’s not exposed by default on npm 3-6.

ljharb: Is there a reason it can't be compatible with RN without adding the peer dep?

ntucker: I guess the node module resolution algorithm will still work even without marked peerdeps.

ntucker: Only way .native will work is through metro so it's be really weird if RN wasn't found.

ljharb (approved):

PR #1313, opened 2021-09-29
enhance: Hoist conditional for Boundary SSR

ntucker:
### Motivation

Only if check once

### DIFF
--- a/packages/core/src/react-integration/provider/BackupBoundary.tsx
+++ b/packages/core/src/react-integration/provider/BackupBoundary.tsx
@@ -1,18 +1,15 @@
 import React, { Suspense, memo, useMemo, version } from 'react';
 
-function BackupBoundary({
-  children,
-}: {
-  children: React.ReactNode;
-}): JSX.Element {
-  // since Suspense does not introduce DOM elements, this should not affect rehydration.
-  if (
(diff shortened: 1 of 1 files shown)


packages/core/src/react-integration/provider/BackupBoundary.tsx
```diff
@@ -1,18 +1,15 @@
...
-  if (
-    (version.startsWith('16') || version.startsWith('17')) &&
-    typeof window === 'undefined'
-  )
-    return children as JSX.Element;
-  return <Suspense fallback={<Loading />}>{children}</Suspense>;
-}
+const NoSuspense =
+  (version.startsWith('16') || version.startsWith('17')) &&
+  typeof window === 'undefined';
```
ljharb, line 5: so react 16 or 17 lacks suspense on the server, but both have it on the client, and react 18 will have it on the server?

ljharb (approved):

packages/core/src/react-integration/provider/BackupBoundary.tsx
ntucker: exactly. Though more specifically in react 17 it will render the fallback on server. But this is undesirable since this suspense is supposed to be a catch-all and would effectively disable all SSR rather than letter users control.

PR #1314, opened 2021-09-29
enhance: Detect multiple normalizr copies installed

### LINKED ISSUE #1312: Delete return symbol on useRessource
**React version** 16.13.0

**Rest Hooks version** (e.g., 4.5.9) 6.1.1

**Describe the bug**

When you delete an entity on a resource that is actually used on a useResource list hooks, the value of the entity in useResource array is `Symbol(ENTITY WAS DELETED)` instead of being juste delete from array.

This lead to a component crash because component is waiting for a EntityType[] (like send by useResource hook) and it receive another type. There is no typescript error that can help to see bug before because the useResource hook register an EntityType[]

**To Reproduce**

1) Have a useResource var in a component like `const list = useResource(Entity.list(),{}) `

2) list is annouced by typescript to be Entity[]

3) Delete an entity with fetch on Entity.delete()

4) App is crashing because for exemple list[0].name is not available with "TypeError: Cannot read property 'name' of undefined" because in list array there is `Symbol(ENTITY WAS DELETED)` on list[0]

**Expected behavior**

Delete of the entity on the list array or param to get only valid Entity on useResource return

Like in the doc https://resthooks.io/docs/api/Delete#impact-on-useresource

ntucker:
Fixes #1312 (assuming this is the cause)

### Motivation

If someone accidentally installs multiple versions of normalizr, identity of Symbols won't work, but it won't manifest in an obvious way.

### Solution

Outright throw an error during denormalization. We only do this during dev mode as packages are not expected to change from dev to production and this would incur a cost for a very extreme edge case.

### DIFF
--- a/packages/normalizr/src/denormalize.ts
+++ b/packages/normalizr/src/denormalize.ts
@@ -37,6 +37,16 @@ const unvisitEntity = (
   if (entity === DELETED) {
     return [undefined, true, true];
   }
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    typeof entity === 'symbol' &&
+    (entity as symbol).toString().includes('DELETED')
+  ) {
+    throw new Error(
+      `Unrecognized symbol detected.
+Make sure you do not have multiple versions of @rest-hooks/normalizr installed.`,
+    );
+  }
   if (typeof entity !== 'object' || entity === null) {
     return [entity, false, false];
   }
--- a/packages/normalizr/src/__tests__/index.test.js
+++ b/packages/normalizr/src/__tests__/index.test.js
@@ -407,6 +407,19 @@ describe('denormalize', () => {
     ).toMatchSnapshot();
   });
 
+  test('denormalizes throws when unknown symbol is found as entity', () => {
+    const entities = {
+      Tacos: {
+        1: Symbol('ENTITY WAS DELETED'),
+      },
+    };
+    expect(() => denormalize(['1', '2'], [Tacos], entities))
+      .toThrowErrorMatchingInlineSnapshot(`
+      "Unrecognized symbol detected.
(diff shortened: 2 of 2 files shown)


packages/normalizr/src/denormalize.ts
```diff
@@ -37,6 +37,16 @@ const unvisitEntity = (
...
     return [undefined, true, true];
   }
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    typeof entity === 'symbol' &&
+    (entity as symbol).toString().includes('DELETED')
+  ) {
+    throw new Error(
+      `Unrecognized symbol detected.
+Make sure you do not have multiple versions of @rest-hooks/normalizr installed.`,
```
ljharb, line 47: the solution here is that normalizr should be a peer dep everywhere it's used - this kind of check is the same kind of runtime protection that React does (because React should also be a peer dep).

Another option is that you could use `Symbol.for`, and then there's no state that needs to be a singleton.

ljharb (approved):

packages/normalizr/src/denormalize.ts
ntucker: Symbol.for seems a bit scary without changing the naming. That would be a breaking change so maybe something to do in the future. Honestly I'm not sure why people get multiple normalizrs, but multiple reacts come easily when using npm since it has a tendency to just keep spawning new installs even when ranges match existing. (even for peerdeps)

ljharb: you'd definitely have to change the naming for that to work.

i've never seen any of those kinds of issues when it's consistently a peer dep, and npm 7 will fail the entire install if there's conflicting reacts ¯\\\_(ツ)\_/¯

PR #1329, opened 2021-10-02
docs: Better guide to avoid npm upgrade issue

ntucker:
Played around til I could get decent instructions here. If they upgrade in one command it avoids duplication so frontload those instructions then change the npm mitigation to be a backup plan.

### DIFF
--- a/docs/upgrade/upgrading-to-6.md
+++ b/docs/upgrade/upgrading-to-6.md
@@ -2,6 +2,9 @@
 title: Upgrading from 5 to 6
 ---
 import BeforeAfterTabs from '@site/src/components/BeforeAfterTabs';
+import PkgTabs from '@site/src/components/PkgTabs';
+
+<PkgTabs pkgs="rest-hooks@6 @rest-hooks/rest@3" upgrade />
 
 ## Highlights
 
@@ -22,13 +25,12 @@ older versions, so this will have to be upgraded in unison. All breaking changes
 
 :::caution
 
-npm 7 is buggy and installs multiple versions when peerDeps are specified.
-
-Due to this it is recommended to either use [yarn](https://yarnpkg.com/) or try the following:
+Due to a bug in npm 7, it might install multiple peerDeps (@rest-hooks/normalizr). If
+this happens, you can fix by completely uninstalling rest hooks packages and then reininstalling:
 
 ```bash
 npm uninstall rest-hooks @rest-hooks/rest
-npm install --save rest-hooks @rest-hooks/rest
+npm install --save rest-hooks@6 @rest-hooks/rest@3
 ```
 
 :::


docs/upgrade/upgrading-to-6.md
```diff
@@ -22,13 +25,12 @@ older versions, so this will have to be upgraded in unison. All breaking changes
 
 :::caution
 
-npm 7 is buggy and installs multiple versions when peerDeps are specified.
-
-Due to this it is recommended to either use [yarn](https://yarnpkg.com/) or try the following:
+Due to a bug in npm 7, it might install multiple peerDeps (@rest-hooks/normalizr). If
```
ljharb, line 28: i'm still not clear on what this bug is, and i don't recall any issues filed about it on npm's repo through npm 7's entire lifecycle. Can you elaborate?

ntucker: peerdeps should never add packages that are already matched by another dep. peerdeps are supposed to mean it's a library and thus it wants to use whatever version the user has for whatever other reason.

ntucker: the whole point of peerdeps existing is so that invariant can hold true. without which what is the point? just use deps

ljharb: I'm confused. The point of peer deps is to hoist a requirement to the root. That means if it's already there and is compatible, use it; if it's there and not compatible, the dep graph is invalid; if it's not there, it should be added.

ljharb: In other words, since forever, when you add something as a peer dep, you ALWAYS are saying "the user MUST explicitly add this into their package.json, or their dep graph is invalid". That's always what it's meant since its inception.

ntucker: "That means if it's already there and is compatible, use it; if it's there and not compatible, the dep graph is invalid" - that is not true with npm 7.

ntucker: it happily installs new versions

ntucker: and then once you get in that state, it won't clean itself up

ljharb: So you're saying you can make a repro repo where it's already present in the dep graph, meets the requirement, and it's creating a duplicate?

If so, can you please file that on npm/cli? I'm pretty confident nobody at npm has ever heard of this.

ntucker: I can do that but users still need to get this working with existing versions

ljharb: Sure, that’s fine, I’m just mystified that you’re the only person that’s seemed to experience this, and I’d love to see it fixed.

ljharb (approved):

PR #1373, opened 2021-10-11
internal: Skip tagging example projects

ntucker:
Lerna's docs are hard but validated `"private": false,` makes this work


ntucker:
Hmm, apparently this doesn't work either https://github.com/yarnpkg/yarn/issues/6371

I really dont know why lerna is creating tags for things that aren't published.

PR #1397, opened 2021-10-17
feat: Add useSuspense()

ntucker:
BREAKING CHANGE: Require @rest-hooks/core@3

### Motivation

Use EndpointInterface instead of ancient FetchShape.
Improve naming.

### Solution

After many discussions `useSuspense()` was determined to be most clear naming.


ntucker:
It doesn't *need* tests, but I will add some here.

PR #1426, opened 2021-10-20
fix: passthrough error that are not NetworkError on NetworkErrorBoundary

### LINKED ISSUE #1425: NetworkErrorBoundary catch all error
**React version** (e.g., 16.8.5) : 16.13.0

**Concurrent mode** no

**Rest Hooks version** 6.1.2

**Describe the bug**

The component NetworkErrorBoundary check if error is a network error before setting error to true in the boundary.

If error is not a network error, we render again the children (Crash again, handle again by NetworkErrorBoundary, render children again, crash again, infinite).

**Expected behavior**

NetworkErrorBoundary has to passtrough errors that are not network error

rtorrente:
Fixes #1425 

### Motivation

NetworkErrorBoundary don't passthrough error that are not NetworkError. It causes infinite crash loop.

### Solution

If error isn't a network error, throw the error in the NetworkErrorBoundary component cause error passthrough to upper ErrorBoundary

### DIFF
--- a/docs/api/Endpoint.md
+++ b/docs/api/Endpoint.md
@@ -106,7 +106,7 @@ endpoint an unpredictable number of times.
 
 Declarative definition of how to [process responses](./schema)
 
-- [where](./schema) to expect [Entities](./Entity)
+- [where](./schema) to expect [Entities](./Entity.md)
 - Classes to [deserialize fields](../guides/network-transform#deserializing-fields)
 
 Not providing this option means no entities will be extracted.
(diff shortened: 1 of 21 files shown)


ntucker:
Hmm if the problem is it attempts to re-render before throwing above....We can introduce a mount count to track this. I'll take a stab at a test

ntucker:
Hmm I can't figure out how to suggest changes, but changing the test to 

```tsx
  it('should fallthrough if status is not set', () => {
    const originalError = console.error;
    console.error = jest.fn();
    let renderCount = 0;
    function Throw(): ReactElement {
      renderCount++;
      throw new Error('you failed');
    }
    const tree = (
      <NetworkErrorBoundary>
        <Throw />
        <div>hi</div>
      </NetworkErrorBoundary>
    );
    expect(() => render(tree)).toThrow();
    console.error = originalError;
    expect(renderCount).toBeLessThan(5);
  });
```

worked. without your change it renders 5 times vs 3 with the change.

PR #1455, opened 2021-10-29
enhance: Dispatch new receive action if new fetch triggered it

ntucker:
### Motivation

Access to endpoint in action is useful

### Solution

In case of users extending NetworkManager without TypeScript we conditionally check controller existance for legacy protection

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -177,23 +177,24 @@ export default class NetworkManager implements Manager {
 
           // don't update state with promises started before last clear
           if (createdAt >= lastReset) {
-            /*if (action.endpoint) {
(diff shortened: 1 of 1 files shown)


ljharb (approved):
any test?

notwillk (approved):

ntucker:
Should be covered by existing tests as I have both paths (legacy, modern) fetch covered

ljharb:
That doesn't make sense to me - it's calling a function now it wasn't calling before, so there can't be any test coverage for the new function call or it'd have been failing before.

ntucker:
The function called here is covered demonstrated in https://app.codecov.io/gh/coinbase/rest-hooks/blob/nm-controller/packages/core/src/controller/Controller.ts

Why would it fail when the function being called doesn't change the outcome of the test, and isn't intended to.

PR #1615, opened 2022-01-12
docs(fix): Update tree-view import to match latest version

ntucker:
Breaking change from https://github.com/coinbase/rest-hooks/pull/1608


ljharb (approved):
seems like there should be some tests that would have prevented #1608 from landing as-is?

ntucker:
Yeah currently there are 0 tests for the website part.

Ideally we should
- validate that build runs
- on changes to /docs or /website - automatically deploy new site

PR #1616, opened 2022-01-12
feat: Add endpoint.optimisticUpdater

### LINKED ISSUE #1515: RFC: optimisticUpdater access to store
# **Is your feature request related to a problem? Please describe.**
Using existing state to determine optimistic response is valuable in some cases to resolve race conditions

### Example - toggle

```ts
const toggle = new Endpoint(fetch('toggle', { method: 'POST' }).then(res => res.json()), {
  optimisticUpdater(snap, id) {
    return snap.getResponse(this, id) ? false : true;
  }
});
```

### Example - PATCH

/updateVisSettings

```
{
  id: 5,
  visType: 'graph',
  numCols: 5,
}
```

```ts
class VisSettings extends Entity {
  pk() { return this.id; }
}

const updateVisSettings= new Endpoint(
  (id, body) => fetch(`/updateVisSettings/${id}`, { method: 'PATCH', body }).then(res => res.json()),
  {
    optimisticUpdater(snap, id, body) {
      return { ...snap.getResponse(getVisSettings, id), ...body };
    },
    schema: VisSettings,
  }
);
```

```ts
getVisSettings(5);
/*
{
  id: 5,
  visType: 'graph',
  numCols: 5,
}
*/
updateVisSettings(5, { visType: 'chart' });
/*
{
  id: 5,
  visType: 'chart',
  numCols: 5,
}
*/
updateVisSettings(5, { numCols: 3 });
/*
{
  id: 5,
  visType: 'chart',
  numCols: 3,
}
*/
/*
without using previous state it would instead be incorrectly
{
  id: 5,
  visType: 'graph',
  numCols: 3,
}
*/
```

# **Describe the solution you'd like**

### Store

For operations during reducer, access to dispatch is banned. Thus we instantiate with current state building `Snapshot` from Controller and State

```ts
class Snapshot {
  private state: State;
  private controller: Controller;
  /*************** Data Access ***************/
  getResponse(endpoint, ...args)​ => { data, expiryStatus, expiresAt };
  getError(endpoint, ...args)​ => ErrorTypes | undefined;
}
```

### Endpoint.optimisticUpdater

`optimisticUpdate` supplemented by `optimisticUpdater` - the former to be deprecated.

```ts
export interface EndpointExtraOptions<F extends FetchFunction = FetchFunction> {
  /** Enables optimistic updates for this request - uses return value as assumed network response
  * to be deprecated
  */
  readonly optimisticUpdate?: (...args: Parameters<F>) => ResolveType<F>;

  /** Enables optimistic updates for this request - uses return value as assumed network response */
  readonly optimisticUpdater?: (snap: Snapshot, ...args: Parameters<F>) => ResolveType<F>;
}
```

This runs during reducer instead of action dispatcher.

### reducer

- close reducer around controller

```ts
case FETCH_TYPE: {
  let optimisticResponse = action.meta.optimisticResponse;
  if (action.endpoint.optimisticUpdater) {
    optimisticResponse = action.meta.optimisticUpdater.call(action.endpoint, controller.snapshot(state), ...action.meta.args);
  }
}
```

## Alternatives

Instead of snapshot, we could send state and controller. This would introduce less concepts, but sacrifice cleanliness of API. Furthermore, all dispatch functions must be banned anyway, so the conception introduce seems like a moot point.

```ts
export interface EndpointExtraOptions<F extends FetchFunction = FetchFunction> {
  /** Enables optimistic updates for this request - uses return value as assumed network response
  * to be deprecated
  */
  readonly optimisticUpdate?: (...args: Parameters<F>) => ResolveType<F>;

  /** Enables optimistic updates for this request - uses return value as assumed network response */
  readonly optimisticUpdater?: (controller: Controller, state: State, ...args: Parameters<F>) => ResolveType<F>;
}
```

### Future work

Incorporate snapshot concept into all side effect methods like 'update'

ntucker:
Fixes #1515 .

### Motivation

Using existing state to determine optimistic response is valuable in some cases to resolve race conditions

### Example - toggle

```ts
const toggle = new Endpoint(
  (): Promise<boolean> =>
    fetch(`/toggle/`, { method: 'POST' }).then(res => res.json()),
  {
    name: 'toggle',
    optimisticUpdater(snap) {
      const { data } = snap.getResponse(getbool);
      if (!data) throw new NoOptimistic();
      return data ? false : true;
    },
  },
);
```

### Example - PATCH

/updateVisSettings

```
{
  id: 5,
  visType: 'graph',
  numCols: 5,
  modifiedAt: {client:1642805150493, server: 1642805150493 }, // vector clock
}
```

```ts
class VisSettings extends Entity {
  pk() { return this.id; }

  static merge(existing, incoming) {
    if (existing.modifiedAt.client < incoming.modifiedAt.client
      || existing.modifiedAt.server< incoming.modifiedAt.server) {
       return {....existing, ...incoming}
    }
    return existing;
  }
}

const updateVisSettings= new Endpoint(
  (id, body) => fetch(`/updateVisSettings/${id}`, { method: 'PATCH', body: { ...body, modifiedAt: Date.now() }).then(res => res.json()),
  {
    optimisticUpdater(snap, id, body) {
      const { data } = snap.getResponse(getVisSettings, id);
      return { id, ...body, modifiedAt: {client: body.modifiedAt, server: data.modifiedAt.server } };
    },
    schema: VisSettings,
  }
);
```

```ts
getVisSettings(5);
/*
{
  id: 5,
  visType: 'graph',
  numCols: 5,
}
*/
updateVisSettings(5, { visType: 'chart' });
/*
{
  id: 5,
  visType: 'chart',
  numCols: 5,
}
*/
updateVisSettings(5, { numCols: 3 });
/*
{
  id: 5,
  visType: 'chart',
  numCols: 3,
}
*/
/*
without using previous state it would instead be incorrectly
{
  id: 5,
  visType: 'graph',
  numCols: 3,
}
*/
```

### Store

For operations during reducer, access to dispatch is banned. Thus we instantiate with current state building `Snapshot` from Controller and State

```ts
class Snapshot {
  private state: State;
  private controller: Controller;
  /*************** Data Access ***************/
  getResponse(endpoint, ...args)​ => { data, expiryStatus, expiresAt };
  getError(endpoint, ...args)​ => ErrorTypes | undefined;
  fetchStart: number;
}
```

### Endpoint.optimisticUpdater

`optimisticUpdate` supplemented by `optimisticUpdater` - the former to be deprecated.

```ts
export interface EndpointExtraOptions<F extends FetchFunction = FetchFunction> {
  /** Enables optimistic updates for this request - uses return value as assumed network response
  * to be deprecated
  */
  readonly optimisticUpdate?: (...args: Parameters<F>) => ResolveType<F>;

  /** Enables optimistic updates for this request - uses return value as assumed network response */
  readonly optimisticUpdater?: (snap: Snapshot, ...args: Parameters<F>) => ResolveType<F>;
}
```

This runs during reducer instead of action dispatcher.

### reducer -> createReducer

- close reducer around controller

```ts
case FETCH_TYPE: {
  let optimisticResponse = action.meta.optimisticResponse;
  if (action.endpoint.optimisticUpdater) {
    optimisticResponse = action.meta.optimisticUpdater.call(action.endpoint, controller.snapshot(state), ...action.meta.args);
  }
}
```

- test lib has a check to determine which to use
- will continue exporting reducer closed around a global controller to keep compatibility. people using this method should not have issues as they won't rely on that controller. however, it is best to take control of the lifetime of the global cache, and thus it is not recommended for use otherwise. will mark it deprecated.

### OptimisticAction

Since optimistic actions compute their payload on receiving, we don't send a payload, therefore we want a new action to reflect that. This also triggers codepath to use endpoints' optimicUpdater function.

### filterOptimistic() changes

Previously optimistic actions would be filtered if they occurred before a response. This was done to prevent optimistic responses queued up from overriding fetches that are resolved after.

While this solves one specific race condition, it does not allow for handling others. The first is that it only handles one race condition; but does not allow for handling cases like toggles or if earlier fetches take longer to even reach the server.

#### Now

For `optimisticUpdater` triggered updates, fetch resolution will only remove its corresponding optimistic action.

The previously handled race condition can be done more robustly as well every other case, by using Entity.merge logic instead.

For example, even without changing an actual server API, artificially adding the time stamp of fetch initiation allows replicating previous behavior:

```ts
class OpEntity extends Entity {
  
  static merge(existing, incoming, existingMeta, incomingMeta) {
    if (existingMeta.date<= incomingMeta.date) {
      return {...existing, ...incoming};
    }
    return existing;
  }
}

const updateOptimistically = new Endpoint(
  async (id, body) => fetch(`/updateVisSettings/${id}`, { method: 'PATCH', body ).then(res => res.json()),
  {
    optimisticUpdater(snap, id, body) {
      const { data } = snap.getResponse(getVisSettings, id);
      return { id, ...data, ...body };
    },
    schema: OpEntity ,
  }
);
```

### Deprecation summary

- reducer -> createReducer
- Endpoint.optimisticUpdate -> Endpoint.optimisticUpdater
- Entity.merge(existing, incoming) -> Entity.merge(existing, incoming, existingMeta, incomingMeta)

### Review notes

- I highly recommend hiding whitespace when viewing diff
- packages/core/src/react-integration/__tests__/integration-optimistic-legacy-endpoint.web.tsx: this is copy/paste of existing tests (packages/core/src/react-integration/tests/integration-optimistic-endpoint.web.tsx). to validate existing endpoints. The above file is then adapted for new methods

### Future work

Incorporate snapshot concept into all side effect methods like 'update'


packages/core/src/state/reducer.ts
```diff
@@ -20,155 +22,168 @@ export const initialState: State<unknown> = {
   lastReset: -Infinity,
 };
 
-export default function reducer(
```
ntucker, line 23: this diff view is very misleading since we added an indentation. the only actual change is the addition of these lines:

```ts
        if (action.endpoint?.optimisticUpdater) {
          optimisticResponse = action.endpoint?.optimisticUpdater.call(
            action.endpoint,
            new Snapshot(state, controller),
            // if endpoint exists, so must args; TODO: fix typing
            ...(action.meta.args as any[]),
          );
        }
```

packages/core/src/state/reducer.ts
```diff
@@ -20,162 +28,192 @@ export const initialState: State<unknown> = {
...
-        /* istanbul ignore next */
-        if (process.env.NODE_ENV !== 'production') {
-          console.warn(
-            'Fetch appears unhandled - you are likely missing the NetworkManager middleware',
-          );
-          console.warn(
-            'See https://resthooks.io/docs/guides/redux#indextsx for hooking up redux',
-          );
-        }
+export default function createReducer(controller: Controller) {
```
ntucker, line 31: I highly recommend hiding whitespace when viewing diff for this file. (gear icon above)

packages/core/src/react-integration/__tests__/integration-optimistic-legacy-endpoint.web.tsx
```diff
@@ -0,0 +1,358 @@
+import {
```
ntucker, line 1: this is copy/paste of existing tests (packages/core/src/react-integration/__tests__/integration-optimistic-endpoint.web.tsx). to validate existing endpoints. The above file is then adapted for new methods

jest.config.js
```diff
@@ -38,6 +38,7 @@ const projects = [
     displayName: 'ReactDOM',
     setupFiles: ['<rootDir>/scripts/testSetup.js'],
     testEnvironment: 'jsdom',
+    transformIgnorePatterns: ['/node_modules/(?!@babel/runtime)'],
```
ljharb, line 41: is this to transform just babel-runtime, or only that?

packages/core/src/controller/createOptimistic.ts
```diff
@@ -0,0 +1,53 @@
...
+  {
+    args,
+  }: {
+    args: readonly [...Parameters<E>];
+  },
+): OptimisticAction {
+  const expiryLength: number = endpoint.dataExpiryLength ?? 1000;
+  /* istanbul ignore next */
+  if (process.env.NODE_ENV === 'development' && expiryLength < 0) {
+    throw new Error('Negative expiry length are not allowed.');
```
ljharb, line 22: so what happens in production if you pass a negative length? there’s nothing that stops someone from relying on that - sometimes people have NODE_ENV always set to production, even in development.

packages/core/src/state/__tests__/applyUpdatersToResults.ts
```diff
@@ -7,12 +7,6 @@ describe('applyUpdatersToResults', () => {
     [ArticleResourceWithOtherListUrl.listShape().getFetchKey({})]: ['1', '2'],
   };
 
-  it('returns results if updaters is not defined', () => {
```
ljharb, line 10: does this indicate a breaking change?

packages/core/src/state/createReducer.ts
```diff
@@ -0,0 +1,264 @@
...
+        let receiveAction: ReceiveAction | OptimisticAction;
+
+        if (optimisticUpdater && action.endpoint) {
+          receiveAction = createOptimistic(action.endpoint, {
+            args: action.meta.args as readonly any[],
+          });
+        } else if (optimisticResponse) {
+          receiveAction = legacyCreateReceive(optimisticResponse, {
+            ...action.meta,
+            dataExpiryLength: 9999999999999,
```
ljharb, line 62:
```suggestion
            dataExpiryLength: Infinity,
```

packages/core/src/types.ts
```diff
@@ -19,25 +19,18 @@ import {
...
+import type { ErrorTypes } from '@rest-hooks/endpoint';
+import type { EndpointUpdateFunction } from '@rest-hooks/core/controller/types';
 
 export type { AbstractInstanceType, UpdateFunction };
 
 export type ReceiveTypes = typeof RECEIVE_TYPE;
 
 export type PK = string;
 
-export interface NetworkError extends Error {
```
ljharb, line 31: These seem like breaking changes as well, since they’re not re-exported

jest.config.js
ntucker: This just transforms babel runtime. With default all of node modules was ignored

packages/core/src/state/__tests__/applyUpdatersToResults.ts
ntucker: No this checked was hoisted to the reducer. This function is not exported from module

packages/core/src/types.ts
ntucker: These were moved to endpoint. All endpoint exports are reexportted. Since this has a normal dep of endpoint it should include that addition

packages/core/src/controller/createOptimistic.ts
ntucker: The dev only checks are all to help discover poorly written static code - as a supplement to typescript as there are things it cannot check. Running local dev with prod means you're opting out of that, just like someone can opt out of typescript.

ljharb: I guess I’m wondering why provide less protection at runtime, rather than always throwing this error?

ntucker: Smaller runtime, faster execution time.

ljharb: Seems far less important than correctness to me, but ok

ljharb (approved):

packages/core/src/controller/createOptimistic.ts
ntucker: The constraint problem is defined as:

- maximize performance and minimize client payload
- ensure catching all static definition errors as early and easy as possible
- guarantee no catastrophic failures in production via isolation

Even with a bad definition here, the result is not catastrophic failure

PR #1653, opened 2022-01-31
enhance: Fetch resolution only removes the optimistic update corresponding to that fetch

ntucker:
### Motivation

We need to ensure only one optimistic response is removed from queue per fetch. This requires exact timestamp matching.

Previously, we did not copy the timestamp but create it twice - this means time between each resulted in them not precisely equally.

### Solution

- Add `Controller.resolve(endpoint, { args, response, fetchedAt, error })`
  - This is similar to receive, except it triggers resolution of an inflight fetch.
- Make docs site use relative package paths to use the latest packages for rest hooks
- Track fetch createdAt with timestamp as it has milisecond resolution. This is used in the receive action as 'fetchedAt' to find the specific corresponding optimistic update to filter away.
- Remove vector clocks example for now as this is only requires for edits on other clients.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -68,6 +68,7 @@ export interface ReceiveMeta<S extends Schema | undefined> {
   args?: readonly any[];
   updaters?: Record<string, UpdateFunction<S, any>>;
   update?: (result: any, ...args: any) => Record<string, (...args: any) => any>;
+  fetchedAt?: number;
   date: number;
   expiresAt: number;
   errorPolicy?: (error: any) => 'soft' | undefined;
@@ -102,6 +103,7 @@ export type OptimisticAction<
       result: any,
       ...args: any
     ) => Record<string, (...args: any) => any>;
+    fetchedAt: number;
     date: number;
     expiresAt: number;
     errorPolicy?: (error: any) => 'soft' | undefined;
@@ -112,7 +114,7 @@ export type OptimisticAction<
 
 export interface ResetAction {
   type: typeof RESET_TYPE;
-  date: Date;
+  date: number | Date;
 }
 
 interface FetchMeta<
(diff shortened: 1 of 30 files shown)


packages/graphql/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@rest-hooks/graphql",
   "version": "0.1.1",
-  "description": "Endpoints for GraphQL APIs",
+  "description": "Endpoints for GraphQL APIs sdfsdfs",
```
ljharb, line 4: ?

packages/endpoint/src/SnapshotInterface.ts
```diff
@@ -20,5 +20,5 @@ export default interface SnapshotInterface {
     ...args: readonly [...Parameters<E['key']>] | readonly [null]
   ) => ErrorTypes | undefined;
 
-  readonly fetchStart: number;
+  readonly fetchedAt: number;
```
ljharb, line 23: does this make it a breaking change?

packages/core/src/types.ts
```diff
@@ -134,7 +136,7 @@ interface FetchMeta<
   resolve: (value?: any | PromiseLike<any>) => void;
   reject: (reason?: any) => void;
   promise: PromiseLike<any>;
-  createdAt: Date;
+  createdAt: number;
```
ljharb, line 139: also maybe this?

ljharb (approved):

packages/endpoint/src/SnapshotInterface.ts
ntucker: this was only released in the previous beta. we don't promise api stability of new features in betas.

packages/graphql/package.json
ntucker: oops!

packages/core/src/types.ts
ntucker: made it number | Date, added back compat handling

PR #1705, opened 2022-02-14
fix: Only import from explicit dependencies

ntucker:
### Motivation

Newer tools are stricter about this sort of thing.


notwillk (approved):
Best PR ever!

PR #1725, opened 2022-02-20
feat: Add component prop to Img

### LINKED ISSUE #1723: @rest-hooks/img custom element
Right now we're stuck with anchor, but what if we want a custom element?

ntucker:
Fixes #1723 .

### Motivation

When using design systems sometimes you want to use a specific element to render your media.

### Solution

Using the same signature as `React.ComponentProps` generic allows us to ensure prop enforcement for both components and 'string' components.

### DIFF
--- a/packages/img/src/Img.tsx
+++ b/packages/img/src/Img.tsx
@@ -1,12 +1,23 @@
-import React, { ImgHTMLAttributes } from 'react';
+import React from 'react';
 import { useResource } from '@rest-hooks/core';
 
 import getImage from './getImage';
 
-export default function Img(
-  props: ImgHTMLAttributes<HTMLImageElement>,
-): React.ReactElement {
-  const { src, alt } = props;
-  useResource(getImage, src ? { src } : null);
-  return <img alt={alt} {...props} />;
+type Props<
+  C extends
+    | keyof JSX.IntrinsicElements
+    | React.JSXElementConstructor<{ src: string }> = 'img',
+> = React.ComponentProps<C> & { component: C };
+
+export default function Img<
+  C extends
+    | keyof JSX.IntrinsicElements
+    | React.JSXElementConstructor<any> = 'img',
+>(props: Props<C>): React.ReactElement {
+  const { component, ...rest } = props;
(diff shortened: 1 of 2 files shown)


packages/img/src/Img.tsx
```diff
@@ -1,12 +1,23 @@
...
+> = React.ComponentProps<C> & { component: C };
+
+export default function Img<
+  C extends
+    | keyof JSX.IntrinsicElements
+    | React.JSXElementConstructor<any> = 'img',
+>(props: Props<C>): React.ReactElement {
+  const { component, ...rest } = props;
+  useResource(getImage, rest.src ? { src: rest.src } : null);
+  return React.createElement(component, rest);
```
ljharb, line 19: it seems really weird to call a component "Img" when it doesn't actually guarantee an image is rendered.

Why is this easier than the caller just using `useResource` (or another hook) directly?

ntucker: Users will have to create their own component like this anyway because of how suspense works. you don't want the data binding above where it is needed.

ntucker: Even in the dom -  isn't the only element that renders images.

ntucker: Img is not for  - it is an abbreviation of Image

ljharb: i get that - but it still seems like it'd make more sense as a custom hook rather than a full component that dynamically takes a component.

ntucker: There already is a hook - it's called useSuspense. The endpoint is provided by this library. However, for 90% of cases, you want to immediately render in its own component. In this case a user would write all of the above, except with statically chosen component.

```tsx
useSuspense(getImage, { src: 'mine.svg' });
<Suspense>
  <Avatar src="mine.svg" />
</Suspense>
```


The above does not work because the suspense hook *must* be below the Suspense boundary.

ljharb: ah, gotcha. this still seems weird to me, but alrighty

ljharb (approved):

PR #1737, opened 2022-02-25
fix: Resource.url properties are enumerable when provided

### LINKED ISSUE #1727: Resource with 'url' property returns '__url' when using spread syntax
**React version** 7.0.2

**Package manager version** yarn 1.22.17
**Node version** 17.14.0

**Rest Hooks version** 6.1.7

**Describe the bug**
Using the spread syntax `...` on a resource object returns `__url` instead of the `url` that I expected. 

**Minimum reproducable example or test case**
Here's a minimal example.  

`test.tsx`
 ```typescript class 
Test extends Resource {
  readonly id: string;
  readonly url: string;
          
  pk() {
    return this.id;
  }
          
  static urlRoot = `http://localhost:7000/test`;  // returns {id: '1', url: 'www.test-url.com}
}
```

`App.tsx`
```typescript
function App() {
    const test = useResource(Test.detail(), {});
    console.log({...test});
    return (
        <AsyncBoundary>
            <div className="App"/>
        </AsyncBoundary>
    );
}

interface Props {
    fallback?: React.ReactElement;
    children: React.ReactNode;
}

function AsyncBoundary({children, fallback = <p>'loading'</p>}: Props) {
    return (
        <Suspense fallback={fallback}>
            <NetworkErrorBoundary>{children}</NetworkErrorBoundary>
        </Suspense>
    );
}
```

The console output
```typescript
{id: '1', __url: 'www.test-url.com'}
```

**Expected behavior**
Output I was expecting
```typescript
{id: '1', url: 'www.test-url.com'}
```

ntucker:
Fixes #1727 .

### Motivation

Resources should return the same object initialized with fromJS() when they are enumerated through spread. This matches normally expected behavior.

### Solution

We can still provide a non-enumerable url getter in the prototype by making the setter define url in the instance with `Object.defineProperty`. Previous attempts to do this failed due to not using the special method. We also test hasOwnProperty() behavior

### DIFF
--- a/packages/experimental/src/rest/BaseResource.ts
+++ b/packages/experimental/src/rest/BaseResource.ts
@@ -52,8 +52,6 @@ export default abstract class BaseResource extends EntityRecord {
   /** URL to find this SimpleResource */
   declare readonly url: string;
 
-  private declare __url?: string;
-
   /** Get the url for a SimpleResource
    *
    * Default implementation conforms to common REST patterns
@@ -236,21 +234,24 @@ export default abstract class BaseResource extends EntityRecord {
       }),
     );
   }
-}
 
-// We're only allowing this to get set for descendants but
-// by default we want Typescript to treat it as readonly.
-Object.defineProperty(BaseResource.prototype, 'url', {
-  get(): string {
-    if (this.__url !== undefined) return this.__url;
-    // typescript thinks constructor is just a function
-    const Static = this.constructor as typeof BaseResource;
-    return Static.url(this);
-  },
-  set(url: string) {
-    this.__url = url;
-  },
-});
+  static {
+    Object.defineProperty(this.prototype, 'url', {
+      get(): string {
+        // typescript thinks constructor is just a function
+        const Static = this.constructor as typeof BaseResource;
+        return Static.url(this);
+      },
+      set(v: string) {
+        Object.defineProperty(this, 'url', {
+          value: v,
+          writable: true,
+          enumerable: true,
+        });
+      },
+    });
+  }
+}
 
 const proto = Object.prototype;
 const gpo = Object.getPrototypeOf;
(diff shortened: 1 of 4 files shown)


packages/experimental/src/rest/BaseResource.ts
```diff
@@ -236,21 +234,24 @@ export default abstract class BaseResource extends EntityRecord {
...
-    // typescript thinks constructor is just a function
-    const Static = this.constructor as typeof BaseResource;
-    return Static.url(this);
-  },
-  set(url: string) {
-    this.__url = url;
-  },
-});
+  static {
+    Object.defineProperty(this.prototype, 'url', {
```
ljharb, line 239: I’m confused why this isn’t defined as a normal prototype getter/setter?

packages/rest/src/SimpleResource.ts
```diff
@@ -377,21 +375,24 @@ export default abstract class SimpleResource extends EntityRecord {
...
-    const Static = this.constructor as typeof SimpleResource;
-    return Static.url(this);
-  },
-  set(url: string) {
-    this.__url = url;
-  },
-});
+  static {
+    Object.defineProperty(this.prototype, 'url', {
+      get(): string {
```
ljharb, line 381: Same question here

ljharb (approved):

packages/experimental/src/rest/BaseResource.ts
ntucker: Typescript distinguishes between getters but I don't want the types to be that way

ntucker: Otherwise a user cannot declare a non getter type, which is exactly what is enabled by the setter definition.

ljharb: It seems very unfortunate to write worse actual code just to make TS types happy.

ntucker: Are there any downsides other than just looking uglier?

PR #1738, opened 2022-02-25
fix: Expiry defaults for useSuspense() should match useResource()

ntucker:
### Motivation

Noticing fetching was increasing when moving to useSuspense(). Realized, the defaults in useSuspense() had error and success expiry times reversed from the defaults specified here: https://github.com/coinbase/rest-hooks/blob/master/packages/core/src/state/NetworkManager.ts#L48

### Solution

Now:
1 minute for success
1 second for error


notwillk (approved):
LGTM

PR #1739, opened 2022-02-25
enhance: Format timestamps in redux devtools for ease of reading

ntucker:
### Motivation

Much easier to debug if timestamps are readable

### Solution

Use `Intl` if available.

### DIFF
--- a/packages/rest-hooks/src/manager/DevtoolsManager.ts
+++ b/packages/rest-hooks/src/manager/DevtoolsManager.ts
@@ -24,6 +24,25 @@ export default class DevToolsManager implements Manager {
   constructor(
     config: DevToolsConfig = {
       name: `Rest Hooks: ${globalThis.document?.title}`,
+      serialize: {
+        replacer: (key: string, value: unknown) => {
+          if (
+            typeof value === 'number' &&
+            (key === 'date' || key.endsWith('At'))
+          ) {
+            if (typeof Intl !== 'undefined') {
+              return Intl.DateTimeFormat('en-US', {
+                hour: 'numeric',
+                minute: 'numeric',
+                second: 'numeric',
+                fractionalSecondDigits: 3,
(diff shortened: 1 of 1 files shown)


packages/rest-hooks/src/manager/DevtoolsManager.ts
```diff
@@ -24,6 +24,25 @@ export default class DevToolsManager implements Manager {
   constructor(
     config: DevToolsConfig = {
       name: `Rest Hooks: ${globalThis.document?.title}`,
+      serialize: {
+        replacer: (key: string, value: unknown) => {
```
ljharb, line 28: seems kind of weird that the return type of this function is `string | Date | unknown`. Should the date and/or the value be stringified?

ljharb (approved):

packages/rest-hooks/src/manager/DevtoolsManager.ts
ntucker: in the case it's not what we're processing we just want to pass through. This has to operate on any value that comes in. I could make it a generic but since I don't really care about return value, I don't see why it's helpful.

ntucker: I think devtools handles other serializations itself https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md

ljharb: fair enough, so it's kind of like toJSON

ntucker: Exactly. In fact I believe at some point it uses this API https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter

PR #1743, opened 2022-02-26
enhance: Devtools improvements

ntucker:
Some refinements based on real-world usage. For instance, by specifying replacer we now need to specify serlializer.optionsto get the default serlializer.options otherwise empty one is used which is different semantically.


packages/rest-hooks/src/manager/DevtoolsManager.ts
```diff
@@ -25,20 +27,27 @@ export default class DevToolsManager implements Manager {
...
+          date: true,
+        },
+        /* istanbul ignore next */
+        replacer: (key: string | number | symbol, value: unknown) => {
           if (
+            HASINTL &&
             typeof value === 'number' &&
+            typeof key === 'string' &&
+            value < Infinity &&
+            value > Infinity &&
```
ljharb, line 42: what about `NaN`?
```suggestion
            isFinite(value) &&
```

packages/rest-hooks/src/manager/DevtoolsManager.ts
```diff
@@ -25,20 +27,27 @@ export default class DevToolsManager implements Manager {
...
+        // the default options are only used if serialize isn't specified, so we include the default here
+        options: {
+          circular: '[CIRCULAR]',
+          date: true,
+        },
+        /* istanbul ignore next */
+        replacer: (key: string | number | symbol, value: unknown) => {
           if (
+            HASINTL &&
             typeof value === 'number' &&
```
ljharb, line 39:
```suggestion
            (typeof value === 'number' || typeof value === 'bigint') &&
```

ljharb (approved):

packages/rest-hooks/src/manager/DevtoolsManager.ts
ntucker: I won't ever be using bigint for a timestamp. This is for human readable Date.now()

ljharb: a user wouldn't ever provide one? also the upcoming `Temporal.Now.instant().epochNanoseconds` is a bigint :-)

ntucker: These are not user fields - this is for my meta data

PR #1745, opened 2022-02-28
fix: lastReset should serialize

ntucker:
### Motivation

Store must be 100% serializable to make SSR work.

I ran into this issue when building out React 18 SSR. The fetches would all stall. After investigation I realized -Infinity serializes to null since it isn't part of JSON spec to have infinity.

### Solution

Since the error is so catastrophic but not visible, we do two fold protection:

- Assume nothing about correctness of cache and do comprehensive checks in getLastReset()
- Use 0 as initial value. This should be just as sufficient as -Infinity since time is currently well in the future of 1970.

### DIFF
--- a/packages/core/src/state/NetworkManager.ts
+++ b/packages/core/src/state/NetworkManager.ts
@@ -136,7 +136,8 @@ export default class NetworkManager implements Manager {
 
   protected getLastReset() {
     const lastReset = this.getState().lastReset;
-    if (typeof lastReset !== 'number') return lastReset.getTime();
+    if (lastReset instanceof Date) return lastReset.valueOf();
+    if (typeof lastReset !== 'number') return -Infinity;
     return lastReset;
   }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/state/NetworkManager.ts
```diff
@@ -136,7 +136,12 @@ export default class NetworkManager implements Manager {
 
   protected getLastReset() {
     const lastReset = this.getState().lastReset;
-    if (typeof lastReset !== 'number') return lastReset.getTime();
+    if (
+      typeof lastReset === 'object' &&
+      typeof lastReset.getTime === 'function'
```
ljharb, line 141: this could still throw if `lastReset` is `null`.

ljharb (approved):

packages/core/src/state/NetworkManager.ts
ntucker: ah, null is object. dang it. what's a better check for Date?

ljharb: the only robust check for Date is https://npmjs.com/is-date-object. The closest thing to that is the quite-brittle `instanceof Date`.

packages/core/src/state/NetworkManager.ts
```diff
@@ -136,7 +136,8 @@ export default class NetworkManager implements Manager {
 
   protected getLastReset() {
     const lastReset = this.getState().lastReset;
-    if (typeof lastReset !== 'number') return lastReset.getTime();
+    if (lastReset instanceof Date) return lastReset.valueOf();
```
ljharb, line 139: heads up that this can still return NaN, which isn't serializable.

ntucker: The return value of this isn't serialized. It's the state that must be serialized

PR #1771, opened 2022-03-07
feat: Entity.useIncoming() for race condition handling

ntucker:
### Motivation

https://github.com/coinbase/rest-hooks/pull/1653 enables full race condition handling when server sends its date. However, without such special handling common cases were subject the client race conditions that were previously avoided.

### Solution

Previously we had logic to try to solve the common case; this pulls this out into overridable function that makes it simple. This also has advantage of improved legacy compat as noone will have previously defined this so previous merge overrides still work.


ljharb (approved):
No test?

ljharb (approved):

PR #1778, opened 2022-03-09
pkg: React 18-rc.2

ntucker:
Latest rc. https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html


.circleci/config.yml
```diff
@@ -142,7 +142,7 @@ workflows:
       - unit_tests:
           matrix:
             parameters:
-              react-version: ["^17.0.0", "beta"]
+              react-version: ["^17.0.0", "^18"]
```
ljharb, line 145: this semver range will not include prereleases; you may want `^18.0.0-rc`
```suggestion
              react-version: ["^17.0.0", "^18.0.0-rc"]
```

ljharb (approved):

.circleci/config.yml
ntucker: On line 82 it checks for this special case to do no installation (if [ "<< parameters.react-version >>" != "^18" ]; then). So this is just a descriptive name

ljharb: gotcha, makes sense.

PR #1779, opened 2022-03-10
fix: Union schemas with null args[0] hooks

ntucker:
### Motivation

null params means 'do nothing' for all our hooks. Currently the inferencing breaks assuming args[0] is an object.

### Solution

Build in protection in union, as well as not trying to 'infer' for null params since that makes no sense.


notwillk (approved):
Lgtm

PR #1783, opened 2022-03-10
enhance: Hooks should type return value based on 'null' arg

ntucker:
### Motivation

Precisely handle null case.

Reduce type and runtime complexity by moving 'null' handling where it belongs. `null` arguments are only a concept needed for hooks since hooks cannot be called conditionally. Therefore, that logic should be tied to the hooks, and removed from further callstacks as early as possible

### Solution

 - Hoist null check to as early as possible
 - Simplify snapshot types to alleviate typescript pressure. No reason some should need null params in a getOptimisticResponse
 - We keep null handling in existing controller methods for back compatibility. However, in the future we should remove these as null is a leaky abstraction beyond hooks
 - Ensure new hooks type handle null parameters as 'undefined' data

Note: making the title of this based on the external impact, since that trumps any internal changes


notwillk (approved):
LGTm

PR #1785, opened 2022-03-13
enhance: Only import from packages in package.json

ntucker:
### Motivation

For PNP, yarn 2 compatibility as well as not having issues with other package managers.

### Solution

Only import from packages referenced in package.json

### DIFF
--- a/packages/experimental/package.json
+++ b/packages/experimental/package.json
@@ -83,7 +83,7 @@
   },
   "peerDependencies": {
     "@rest-hooks/core": "^3.0.0-0",
-    "@rest-hooks/endpoint": "^2.0.0",
+    "@rest-hooks/endpoint": "^2.2.0-beta.2",
     "@rest-hooks/rest": "^3.0.0",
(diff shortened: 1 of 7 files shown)


packages/experimental/package.json
```diff
@@ -83,7 +83,7 @@
   },
   "peerDependencies": {
     "@rest-hooks/core": "^3.0.0-0",
-    "@rest-hooks/endpoint": "^2.0.0",
+    "@rest-hooks/endpoint": "^2.2.0-beta.2",
```
ljharb, line 86: this is a breaking change (experimental package or no)

ljharb (approved):

packages/experimental/package.json
ntucker: 1. This is only breaking for beta-to-beta release, which we don't guarantee compatibility.
2. This dependency was already required in a previous breaking pr, but the specification was mistakenly not updated.

So essentially this breaking change is already covered.

PR #1786, opened 2022-03-13
enhance: Loosen SnapshotInterface to allow compatible but distinct versions

ntucker:
### Motivation

With multiple versions of @rest-hooks/endpoint installed, this could result in hooks not matching the types expected.

### Solution

`ExpiryStatusInterface` has identical structural type, but lacks the nominal typing restriction that breaks multi-package compatibility.

### DIFF
--- a/packages/endpoint/src/Expiry.ts
+++ b/packages/endpoint/src/Expiry.ts
@@ -3,3 +3,5 @@ export const enum ExpiryStatus {
   InvalidIfStale,
   Valid,
 }
+// looser version to allow for cross-package version compatibility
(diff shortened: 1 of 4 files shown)


ljharb (approved):
enums are gross

ntucker:
Yeah I kinda have mixed feelings now. It is nice to have the typeahead, but adding effectively nominal typing is a divergence from much of the principals of typescript.

ljharb:
imo nominal typing is wonderful, and peer deps should be preventing any duplication; it just sucks that enums blur the lines between type and value space.

PR #1848, opened 2022-03-30
docs(fix): Restore version routes

ntucker:
- None of the other version routes were working because route priority is only set with path === ''
  - Also get rid of README overrides that conflict with other routes (README is base path for docusaurus)
- Mark v5 as maintained
- Don't show badge for current version
- Only compile the last 4 versions in devmode for perf
- Programmatic version select menu generation
- Upgrade to latest docusaurus (beta.18) https://github.com/facebook/docusaurus/releases/tag/v2.0.0-beta.18


packages/normalizr/README.md
```diff
@@ -25,8 +25,8 @@ Normalizr is a small, but powerful utility for taking JSON with a schema definit
 
 ## Documentation
 
-* [Introduction](./docs/introduction.md)
```
ljharb, line 28: does this mean links to this page will now be broken?

ljharb (approved):

packages/normalizr/README.md
ntucker: Introduction has always been at / so any url links won't change. (Url is set by path). As long as there are no references to the file in the code base there won't be any issues

PR #1852, opened 2022-04-01
fix: Default optimistic race condition assume in-order fetch timings

ntucker:
### Motivation

While adapting API is useful to guarantee correct handling of networking race conditions, the default case should handle the most common assumption. That assumption is that fetches will be resolved in the order they are sent. This makes sense because the only race condition where this is violated is if the server receive time is variable. Whereas there are many parts afterward there could affect response timing.

This assumption makes sense because with an RDS, transactions will lock a given row on first-come basis; and that transaction will only include updates from that request - thus any requests received during processing of that request will *not* be reflected in the response.

#### Events:

Numbered are events; the bullets represent what state is applied

1) fetch1
    - optimistic1
2) fetch2
    - optimistic1
    - optimistic2
3) fetch3
    - optimistic1
    - optimistic2
    - optimistic3
4) resolve1
    - resolve1
    - optimistic2
    - optimistic3
5) resolve2
    - resolve1
    - resolve2
    - optimistic3
6) resolve3
    - resolve1
    - resolve 2
    - resolve 3

In summary, there are three timings for a request - send time, server processing time, and receive time. This default algorithm will handle any problems with the latter 2 - requiring only send time to be in order for data integrity to be maintained.
 
### Solution

Store `fetchedAt` in entityMeta and use it as default useIncoming() implementation.

Additionally:
- Update @rest-hooks/legacy so it includes extra methods
- Update EntityInterface and handle legacy Entities that don't have new methods to support several versions back.

### Is this breaking?

While technically a breaking change in behavior

- The previous behavior was not intended
- Given download tracking it seems unlikely anyone has grown dependent on this as it is only newly introduced in 6.2
- Behavior change only matters in race conditions; and it seems very unlikely people would actually want to rely on the existing behavior since that behavior is problematic.

### DIFF
--- a/packages/core/src/react-integration/__tests__/integration-optimistic-endpoint.web.tsx
+++ b/packages/core/src/react-integration/__tests__/integration-optimistic-endpoint.web.tsx
@@ -299,7 +299,7 @@ describe.each([
 
       // first optimistic
 
-      act(() => {
+      await act(async () => {
         fetches.push(
           result.current.fetch(
             CoolerArticleResource.partialUpdate().extend({
@@ -324,6 +324,7 @@ describe.each([
           content: 'firstoptimistic',
         }),
       );
+      await new Promise(resolve => setTimeout(resolve, 0));
 
       // second optimistic
       act(() => {
@@ -350,6 +351,7 @@ describe.each([
           content: 'firstoptimistic',
         }),
       );
+      await new Promise(resolve => setTimeout(resolve, 0));
 
       // third optimistic
       act(() => {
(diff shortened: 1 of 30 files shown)


packages/core/src/react-integration/__tests__/integration-optimistic-endpoint.web.tsx
```diff
@@ -406,8 +409,167 @@ describe.each([
...
         CoolerArticleResource.fromJS({
           ...payload,
-          title: 'first',
+          title: 'second',
+          tags: ['thirdoptimistic'],
+        }),
+      );
+    });
+
+    it('should clear optimistic when server response resolves in order', async () => {
```
ntucker, line 418: @sweetro This test case covers a server responding to each request in order. As you can see with the expects - each time we stop applying the next optimistic response

notwillk (approved):
LGTM

PR #1853, opened 2022-04-01
fix: Hooks with null param maintain basic schema structure

ntucker:
### Motivation

Maintaining schema structure simplifies cases where entities are nested in a structure even if null parameter is set. A common need for this is in pagination.

```tsx
const { results, nextPage } = useSuspense(
  PaginatedArticleResource.list(),
  should ? {} : null,
);
```

### Solution

Undo part of https://github.com/coinbase/rest-hooks/pull/1783

However, still not accept null params in snapshot and keep some of the type improvements.

Add a test for pagination that validates types and runtime value.

### DIFF
--- a/packages/core/src/controller/Controller.ts
+++ b/packages/core/src/controller/Controller.ts
@@ -26,7 +26,6 @@ import {
 } from '@rest-hooks/normalizr';
 import { inferResults } from '@rest-hooks/normalizr';
 import { unsetDispatch } from '@rest-hooks/use-enhanced-reducer';
-import { nullResponse } from '@rest-hooks/core/react-integration/newhooks/constants';
 
 type RHDispatch = (value: ActionTypes) => Promise<void>;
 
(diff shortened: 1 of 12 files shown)


packages/core/src/controller/Controller.ts
```diff
@@ -244,13 +243,11 @@ export default class Controller {
...
     expiresAt: number;
   } => {
-    if (rest[0] === null) {
-      return nullResponse;
-    }
     const state = rest[rest.length - 1] as State<unknown>;
     const args = rest.slice(0, rest.length - 1) as Parameters<E['key']>;
-    const key = endpoint.key(...args);
-    const cacheResults = state.results[key];
+    const activeArgs = args?.[0] !== null;
```
ljharb, line 248: so this is basically testing for "the first item in `rest` is `null`"? what does "activeArgs" mean?

ljharb (commented):
i can't tell if this is a refactor or a breaking change

packages/core/src/controller/Controller.ts
ntucker: open to naming suggestions - it distinguishes two argument mode:

[null] - meaning 'do nothing' (how to deal with no if/else in hooks)
...args - whatever args are defined by endpoint - so 'activate' the hook

ntucker:
Technically this is undoing an accidental breaking change in https://github.com/coinbase/rest-hooks/pull/1783. That wasn't intended to be breaking as I misunderstood some things

notwillk (commented):
Looks good, but I'll let @ljharb do the final review since he has questions

packages/core/src/controller/Controller.ts
ljharb: ok so `args.length === 1 && args[0] === null` indicates "do nothing"?
```suggestion
    const isActivation = args.length !== 1 || args[0] !== null;
```
(also, `rest.slice(0, -1)` on the line above)

ljharb (approved):
with discussed changes

packages/core/src/controller/Controller.ts
ntucker: existing behavior has only checked first arg for null, so this would technically be a breaking change. What about isActive? activation doesn't seem like it describes how the hook works

ntucker: pushed change with isActive as name (and also updated its uses...since that's pretty important)

PR #1861, opened 2022-04-02
docs: Update yarn.lock for docs site

ntucker:
### Motivation

Since we use relative paths (waiting on yarn 3, which is waiting on renovate), yarn.lock has to be updated every publish to get new versions.


notwillk (approved):
I much prefer the new numbers!

PR #1866, opened 2022-04-03
feat: Add useDLE() to core

### LINKED ISSUE #1467: useStatefulResource throw "no provider error"
**React version** (e.g., 16.8.5) :  16.14.0

**Concurrent mode** no

**Rest Hooks version** 6.1.7

**Describe the bug**

useStatefulResource throw an error "It appears you are trying to use Rest Hooks without a provider.
Follow instructions: https://resthooks.io/docs/getting-started/installation#add-provider-at-top-level-component" but there is already a <CacheProvider> at the top of the app. 

The same request with a normal useResource works without any problem

**Expected behavior**

The request have to works when there is a <CacheProvider/> in the app

_Additionnal information : I'm not sure why this feature is in the legacy package as I am dealing with cases that require its use._

_For example if I'm looking to create a Guard component which try  to verify that a resource exists and that it has a particular property before showing the children. 
If the request fails or the property is not present, I need to call a function and redirect the user. Without useStatefullResource, I would have to create a specific ErrorBoundary, which will unnecessarily complicate the code. Maybe my usage could be done by another use that I did not understand?_

ntucker:
Fixes #1467

### Motivation

- Keep legacy just for back compat support.
- Avoid cross-package peerDeps issues with weird package managers

### Solution

useDLE (data-loading-error) is like useStatefulResource except:

- Accepts [Endpoints](https://resthooks.io/docs/getting-started/endpoint) only
- Inside core package so additional package not needed.
- Doesn't awkwardly rely on denormalize export from normalizr
- Everything else is copy/paste of useStatefulResource()

Note: Keeping useStatefulResource() in legacy for use with FetchShape

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -30,6 +30,7 @@ export {
 export {
   useSuspense,
   useFetch,
+  useDLE,
 } from '@rest-hooks/core/react-integration/newhooks/index';
 export type { ErrorTypes } from '@rest-hooks/core/react-integration/index';
 export {
--- a/packages/core/src/react-integration/newhooks/__tests__/fixtures.ts
+++ b/packages/core/src/react-integration/newhooks/__tests__/fixtures.ts
@@ -1,0 +1,122 @@
+export const payload = {
+  id: 5,
+  title: 'hi ho',
+  content: 'whatever',
+  tags: ['a', 'best', 'react'],
+};
+export const payload2 = {
+  id: 6,
+  title: 'next',
+  content: 'my best content yet',
+  tags: ['b'],
+};
+
+export const createPayload = {
+  id: 1,
+  title: 'hi ho',
+  content: 'whatever',
+  tags: ['a', 'best', 'react'],
+};
+
+export const articlesPages = {
+  prevPage: '23asdl',
+  nextPage: 's3f3',
+  results: [
+    {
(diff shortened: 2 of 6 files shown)


packages/core/src/react-integration/newhooks/useDle.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+  | { data: DenormalizeNullable<S>; loading: true; error: undefined }
+  | { data: DenormalizeNullable<S>; loading: false; error: ErrorTypes }
+>;
+
+/**
+ * Use async date with { data, loading, error } (DLE)
+ * @see https://resthooks.io/docs/guides/no-suspense
+ */
+export default function useDle<
```
notwillk, line 37: How do you feel about `useDLE` instead?

notwillk (approved):

packages/core/src/react-integration/newhooks/useDle.ts
ntucker: I was thinking about that. It does appear more pleasing to the eye and makes it more obvious it's a TLA

ljharb: Initialisms should always be full caps; Java-style camel casing algorithms are horrific

PR #1867, opened 2022-04-03
enhance: Resource extends from Entity rather than EntityRecord

ntucker:
BREAKING CHANGE: Resource loses hasDefined, mergeRecord,
toObjectDefined, keysDefined, set

### Motivation

These weren't super useful in practice

### Solution

No longer rely on rest package which shouldn't have been the case to begin with.


ljharb (approved):

ljharb:
(surprising no tests broke tho)

PR #1869, opened 2022-04-04
feat: Improved Resource

ntucker:
BREAKING CHANGE:
- Removed EntityRecord
- Removed SimpleResource
- New BaseResource
- RestFetch, FetchMutate, FetchGet args change (FetchMutate<Params, Body> -> FetchMutate<[Params, Body]>)
- Remove deprecated 'Shape' methods (pre-endpoints)

### Motivation

https://resthooks.io/blog/2021/08/29/Experimental-useController#new-hiearchy

#### Support variable arg REST Endpoints

```ts
class MyResource {
  static current<T extends typeof Resource>(this: T): RestEndpoint<FetchGet<[]>, T, undefined> {
    return super.endpoint().extend({
      schema: this,
    });
  }
  static delete<T extends typeof Resource>(this: T): RestEndpoint<FetchMutate<[id: number]>, schema.Delete<T>, true> {
    return super.delete().extend({
      getOptimisticResponse: (snap, id) => id,
      schema: new schema.Delete(this),
    });
  }
}
```

#### BaseResource - all the Resource with none of the endpoints. Start your endpoint definitions from scratch.

- Won't come with endpoints that it doesn't have
- Doesn't have to conform to inheritance hierarchy of base endpoints

```ts
// read-only
class MyResource extends BaseResource {
  static detail<T extends typeof Resource>(
    this: T,
  ): RestEndpoint<
    FetchGet<[id:number]>,
    T,
    undefined
  > {
    return this.endpoint().extend({
        schema: this,
      });
  }
}
```

### Solution

Pulled from experimental, but did not include pagination

Split into 4 parts as intended to be landed in 4 commits rather than squashed.


packages/legacy/src/rest-3/errors.ts
```diff
@@ -0,0 +1,7 @@
+export class NotImplementedError extends Error {
+  /* istanbul ignore next */
+  message =
+    process.env.NODE_ENV === 'production'
+      ? /* istanbul ignore next */ 'Not Implemented'
+      : 'Not Implemented. You must provide an override for this method.';
```
ljharb, line 6: is there a point in changing this message in prod? only the site devs are going to be looking at it

packages/legacy/src/rest-3/paramsToString.native.ts
```diff
@@ -0,0 +1,10 @@
+export default function paramsToString(
+  searchParams: Readonly<Record<string, string | number>>,
+) {
+  const params = new URLSearchParams(searchParams as any);
+  try {
+    params.sort();
+    // eslint-disable-next-line no-empty
+  } catch {}
+  return params.toString();
```
ljharb, line 9:
```suggestion
  return String(params ?? '');
```

packages/rest/src/BaseResource.ts
```diff
@@ -0,0 +1,263 @@
...
+        return Static.url(this);
+      },
+      set(v: string) {
+        Object.defineProperty(this, 'url', {
+          value: v,
+          writable: true,
+          enumerable: true,
+        });
+      },
+    });
```
ljharb, line 251: why can't these be normal prototype getter methods again?

packages/rest/src/BaseResource.ts
```diff
@@ -0,0 +1,263 @@
...
+
+const proto = Object.prototype;
+const gpo = Object.getPrototypeOf;
+
+function isPojo(obj: unknown): obj is Record<string, any> {
+  if (obj === null || typeof obj !== 'object') {
+    return false;
+  }
+  return gpo(obj) === proto;
+}
```
ljharb, line 263: `{ __proto__: null }` is a pojo also, but will fail this predicate

packages/legacy/src/rest-3/errors.ts
ntucker: Smaller bundle. React has a better solution where they give an id you can go to a website for when you get errors in prod mode

ljharb: seems like a lot of added complexity to save 46 gzippable characters

packages/rest/src/BaseResource.ts
ntucker: TypeScript differentiates between getters and non getters so it doesn't let people define normal url properties which is what this setter is supposed to allow

ljharb: and there's no way to use TS comments to fix it, instead of writing code that results in a larger bundle?

packages/rest/src/BaseResource.ts
ntucker: how would that happen? seems an odd thing to pass to getFetchInit

packages/legacy/src/rest-3/paramsToString.native.ts
ntucker: how does the existing fail? undefined?

packages/rest/src/BaseResource.ts
ljharb: a null object? it's a normal thing to create if you want to avoid the possibility of key collisions.

packages/legacy/src/rest-3/errors.ts
ntucker: as the branch has no interactions with other state or branches it adds a complexity of size 1.

ljharb: i'm worried about the human complexity, not the algorithmic complexity (which rarely matters in practice)

ntucker: algorithmic complexity is things like runtime and memory, which relate to the cost of running an algorithm in terms of computational resources with relation to the input size. The only control flow that can affect these things are variable allocations, and loops (recursion is equivalent to loops). Adding branches cannot affect these runtime conditions by themselves.

I don't know what you mean by 'human complexity' but I am speaking to the understandability, testability, provability of the code. If one were to prove the behavior of an algorithm, the complexity I am describing would be the number of distinct cases that would have to be enumerated. Additionally even if some cases are technically equivalent, if it is not obvious that they are, that too must be proven resulting in those cases being added to complexity.

How does this manifest? Nesting conditionals where the variables do not have exclusive state is one way; then if those states can affect other parts of the control flow (by function calls) one must multiply all those possibilities. That's where things like two way data binding become intractable because while yes, they often result in infinite runtime loops they also result in infinite cases of control flow.

The affect of this conditional is isolated. Its only side effect is a console log - which happens in both cases. It has no return value. Without reading any code one can see that there is no way for it to affect the behavior of any other piece. Therefore this conditional provably multiplies by nothing.

packages/rest/src/BaseResource.ts
ntucker: what is a TS comment?

packages/rest/src/BaseResource.ts
ntucker: I found this when googling null object: https://stackoverflow.com/questions/801032/why-is-null-an-object-and-whats-the-difference-between-null-and-undefined

Can you find better docs or share example I'm still not understanding. Appreciate you!

packages/rest/src/BaseResource.ts
ntucker: https://github.com/coinbase/rest-hooks/pull/1737#discussion_r815360548 < original conversation

ntucker: Another key distinction is the setter/getter are on the prototype, but the setter sets a property on the actual value itself, which overrides this by avoiding the prototype chain lookup.

packages/legacy/src/rest-3/paramsToString.native.ts
ljharb: both undefined and null, yeah. the try/catch on the sort means you can't be sure it's coercible to an object by this point.

packages/rest/src/BaseResource.ts
ljharb: oh sure, but `set url(v) {  Object.defineProperty(…); }` would achieve the same thing, if not for the TS stuff.

packages/rest/src/BaseResource.ts
ljharb: The term "null object" is just a colloquialism for "an object literal with a null [[Prototype]]", like `Object.create(null)`. The point of it is to remove the chance of collision with things on Object.prototype.

"pojo" checks are generally pretty weird - generally speaking, it's easier to have an argument either be a function, an array, or "we'll pretend it's an options bag".

ntucker: Maybe pojo is the wrong term? Really this is just to exclude other valid body arguments to fetch like strings, blobs, etc. 

Perhaps this is something you can create another PR for? As this code already exists and this PR is intended to just be a moving of code and updating of tests.

ljharb: that's up to you; i was just pointing out the problem.

packages/legacy/src/rest-3/paramsToString.native.ts
ntucker: I guess I'm confused how `new URLSearchParams(searchParams as any);` results in null or undefined

packages/rest/src/BaseResource.ts
ntucker: I do have `obj === null` above...and I get typeof undefined === 'undefined'. So it seems both those cases are handled

packages/legacy/src/rest-3/paramsToString.native.ts
ljharb: oh, you're totally right. in this case it'd just be that if someone monkeyed with URLSearchParams.prototype.toString, the return might not actually be a string, whereas with `String(params)` it always will be.
```suggestion
  return String(params);
```
but i'll resolve this, it's not as important as i thought.

notwillk (approved):
LGTM!

PR #1881, opened 2022-04-08
fix: Legacy exports

ntucker:
### Motivation

Relative imports only work with files not folders


notwillk (approved):

ljharb:
um, they definitely work with folders too? This change isn’t wrong but should not have been needed.

PR #1887, opened 2022-04-08
internal: React types to 18

ntucker:
17->18. They also included a change to how component types work that required update to children props


notwillk (approved):
A description would be nice, but the code looks good

ljharb (approved):

PR #1908, opened 2022-04-13
enhance: Improve robustness when using distinct schemas to normalize/denormalize

ntucker:
### Motivation

New [validation cycle](https://resthooks.io/docs/getting-started/validation#partial-results) means users are encouraged to use multiple distinct schemas for the same entity depending on endpoint.

This breaks previous invariant that denormalize path always applied to same normalize path, which could result in confusing bugs if not defined correctly. For instance, if a nested schema is not defined for a normalization path, but it ends up actually getting that field, then the denormalization will end up getting a value from the response - rather than processed id.

### Solution

"partial denormalization" is already intended for support (as there are tests for it). However, this broke with the assumption id was a string in `unvisitEntity`. This is used for the global referential equality guarantee.

First step is to update types correctly to ensure good processing.

Next we compute `PK` in unvisitEntity in case id is the object itself. This allows us to correctly index the global equality store. In case this fails, we simply short-circuit just like if entity were an unexpected type (`[entity, false, false]`).

The end result is that if normalization wasn't run, it can still denormalize to an expected value. This means consistent schema definitions for Entities are not required.

This bug was hidden in a previous snapshot test that was clearly showing invalid results. So that test is now updated with a good snapshot.

From the tests:

```ts
  test('can denormalize already partially denormalized data', () => {
    const entities = {
      Menu: {
        '1': { id: '1', food: { id: '1' } },
      },
      Food: {
        '1': { id: '1' },
      },
    };

    expect(denormalize('1', Menu, entities)).toMatchSnapshot();
    expect(denormalize('1', Menu, fromJS(entities))).toMatchSnapshot();
  });
```

Here 'food' is set as an object rather than the pk itself. However, since that object can be used to create a PK it will still work.

### DIFF
--- a/packages/normalizr/src/denormalize.ts
+++ b/packages/normalizr/src/denormalize.ts
@@ -7,7 +7,7 @@ import {
   DenormalizeCache,
   UnvisitFunction,
 } from '@rest-hooks/normalizr/types';
-import Entity, { isEntity } from '@rest-hooks/normalizr/entities/Entity';
+import { isEntity } from '@rest-hooks/normalizr/entities/Entity';
 import { DELETED } from '@rest-hooks/normalizr/special';
 import WeakListMap from '@rest-hooks/normalizr/WeakListMap';
 import { denormalize as arrayDenormalize } from '@rest-hooks/normalizr/schemas/Array';
@@ -17,23 +17,19 @@ import { isImmutable } from '@rest-hooks/normalizr/schemas/ImmutableUtils';
 const DRAFT = Symbol('draft');
 
 const unvisitEntity = (
-  id: any,
-  schema: any,
+  entityOrId: Record<string, any> | string,
+  schema: EntityInterface,
   unvisit: UnvisitFunction,
   getEntity: (
     entityOrId: Record<string, any> | string,
-    schema: typeof Entity,
-  ) => EntityInterface | typeof DELETED,
+    schema: EntityInterface,
+  ) => object | typeof DELETED,
   localCache: Record<string, Record<string, any>>,
   entityCache: DenormalizeCache['entities'],
   dependencies: object[],
   cycleIndex: { i: number },
-): [
-  denormalized: EntityInterface | undefined,
-  found: boolean,
-  deleted: boolean,
-] => {
-  const entity = getEntity(id, schema);
+): [denormalized: object | undefined, found: boolean, deleted: boolean] => {
+  const entity = getEntity(entityOrId, schema);
   if (entity === DELETED) {
     return [undefined, true, true];
   }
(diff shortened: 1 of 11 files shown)


packages/normalizr/src/denormalize.ts
```diff
@@ -51,25 +47,33 @@ Make sure you do not have multiple versions of @rest-hooks/normalizr installed.`
...
-  if (!localCache[schema.key][id]) {
+  if (!localCache[schema.key][pk]) {
     const trackingIndex = dependencies.length;
     dependencies.push(entity);
 
     const wrappedUnvisit = withTrackedEntities(unvisit);
     // { [DRAFT] } means we are still processing - which if found indicates a cycle
     wrappedUnvisit.setLocal = entityCopy =>
-      (localCache[schema.key][id] = { [DRAFT]: entityCopy, i: trackingIndex });
+      (localCache[schema.key][pk] = { [DRAFT]: entityCopy, i: trackingIndex });
```
ntucker, line 72: @ljharb since 'pk' here can be derived from networking response, is prototype pollution a potential issue here? The codeql scan just complains about schema.key - but since that's developer set it does not concern me.

notwillk (approved):
Assuming this prototype issue is cool, then I aprove

packages/normalizr/src/denormalize.ts
ljharb: Yes, unless `localCache` is has null objects as its values.

packages/normalizr/src/denormalize.ts
```diff
@@ -51,25 +47,33 @@ Make sure you do not have multiple versions of @rest-hooks/normalizr installed.`
...
+    typeof entityOrId === 'string'
+      ? entityOrId
+      : schema.pk(isImmutable(entity) ? (entity as any).toJS() : entity);
+  if (pk === undefined || pk === '') {
+    return [entity, false, false];
+  }
+
   if (localCache[schema.key] === undefined) {
-    localCache[schema.key] = {};
+    localCache[schema.key] = Object.create(null);
```
ntucker, line 59: @ljharb so this should protect against prototype pollution? (remember schema.key is dev set, but pk is based on network response and we index like so `localCache[schema.key][pk]`)

ljharb: Yep! It’s safe to put any user-supplied keys into a null object.

packages/normalizr/src/denormalize.ts
ntucker: Actually if we're dictating the value set.... the best that they could do is break the prototype since they can't control what it's set to - no?

PR #1917, opened 2022-04-16
fix: urlRoot type inference handles escapes anywhere in string

ntucker:
### Motivation

If \\: shows up after a real group those groups will be missed.


notwillk (approved):
Nice catch!

PR #1963, opened 2022-04-29
fix: Accept undefined or null responses for schema entries

ntucker:
### Motivation

```tsx
class MyItem extends Entity {
  readonly createdAt: Date | undefined;
  static schema = {
    createAt: Date,
  }
}
```

@WayneYe

### Solution

- `null` and `undefined` are always considered valid from a network response
- 'required' fields should not default to either or users should specify a validate() method
- Greatly simplified logic as handling null and undefined in root denormalize
- `undefined` inputs short-circuit 'found' tracking since they should infer something other than undefined if we should recurse. Therefore any `undefined` value is considered not 'found'

### Is this a breaking change?

While this technically changes behavior; We meticulously validated each change to prevent regressions. We are strongly certain that if the new behavior is different it means it is more of what a user wants. I.e., they get a result when it would have errored before.

### DIFF
--- a/packages/normalizr/src/denormalize.ts
+++ b/packages/normalizr/src/denormalize.ts
@@ -126,28 +126,32 @@ const getUnvisit = (
   ): [denormalized: any, found: boolean, deleted: boolean] {
     if (!schema) return [input, true, false];
 
-    if (!schema.denormalize || typeof schema.denormalize !== 'function') {
-      if (typeof schema === 'function') {
-        if (input instanceof schema) return [input, true, false];
-        return [new schema(input), true, false];
-      } else if (typeof schema === 'object') {
-        const method = Array.isArray(schema)
-          ? arrayDenormalize
-          : objectDenormalize;
-        return method(schema, input, unvisit);
-      }
-    }
-
     // null is considered intentional, thus always 'found' as true
     if (input === null) {
       return [input, true, false];
     }
 
+    const hasDenormalize = typeof schema.denormalize === 'function';
+
+    if (!hasDenormalize && typeof schema === 'function') {
+      if (input instanceof schema) return [input, true, false];
+      // field deserialization should never count against 'found' (whether to used inferred results)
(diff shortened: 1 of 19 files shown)


packages/normalizr/src/denormalize.ts
```diff
@@ -126,28 +126,33 @@ const getUnvisit = (
...
-      }
-    }
-
     // null is considered intentional, thus always 'found' as true
     if (input === null) {
       return [input, true, false];
     }
 
+    const hasDenormalize =
+      schema.denormalize && typeof schema.denormalize === 'function';
```
ljharb, line 135:
```suggestion
    const hasDenormalize = typeof schema.denormalize === 'function';
```
since if it's falsy it won't have a typeof of function

packages/normalizr/src/schemas/Array.ts
```diff
@@ -43,20 +43,15 @@ export const denormalize = (
...
 ): [denormalized: any, found: boolean, suspend: boolean] => {
   schema = validateSchema(schema);
-  let deleted = false;
-  let found = true;
-  if (input === undefined && schema) {
-    [, found, deleted] = unvisit(undefined, schema);
-  }
   return [
-    input && input.map
+    input.map
```
ljharb, line 47: are you expecting a non-array here with `map` present? or could this use `Array.isArray`

packages/normalizr/src/schemas/ImmutableUtils.ts
```diff
@@ -12,7 +12,6 @@
  */
 export function isImmutable(object: any) {
   return !!(
-    object &&
     typeof object.hasOwnProperty === 'function' &&
```
ljharb, line 15: removing this line, with a type of any/unknown, will throw when you try to access `object.hasOwnProperty` (because of null or undefined). if that's intentional, then the type should be changed to `{}` since that's the type that matches "object-coercible"

PR #1973, opened 2022-04-30
docs: Provide warning about using jest.mock

### LINKED ISSUE #1689: Imports from @react-hooks/test failing with circular dependency TypeError
17.0.2

**Concurrent mode** yes/no
no

```json
"@rest-hooks/core": "^3.0.2",
"@rest-hooks/endpoint": "^2.1.0",
"@rest-hooks/hooks": "^2.1.0",
"@rest-hooks/rest": "^3.0.1",
"@rest-hooks/test": "^7.1.3",
```

**Describe the bug**
Any and all imports from `@rest-hooks/test` cause the following error: `TypeError: Class extends value undefined is not a constructor or null`. 

I first got this error when using `MockResolver` in my component unit tests. The line that failed was not my usage, but rather my import of it. A fair amount of searching led me to believe this is related to circular dependencies or a failure for a dependency to load. After ensuring the integrity of my own `.test.tsx` and `.tsx` files, I figured I'd try another approach: I tried the deprecated `MockProvider`, as well as using `mockInitialState()`, all of which resulted in the same import error, not usage errors.

**To Reproduce**
1. Follow the instructions here to set up `MockResolver` for a component unit test. (I'm using Jest)
2. Run tests 

**Expected behavior**
The packages should import without error.

**Additional context**
I tried these imports in a separate file, as well, and received the same failure. I followed `react-testing-library`'s approach to creating custom renderers, and created one with a `MockResolver` wrapping the `children`, and the import once again failed.

ntucker:
Fixes #1689 .

### Motivation

Someone tried to do

```tsx
jest.mock("rest-hooks")
```

But did not provide an implementation for every export. Therefore when `@rest-hooks/test` tried to create a new class that extends it, it was extending undefined resulting in `TypeError: Class extends value undefined is not a constructor or null`

### Solution

Warn on the debugging docs that this is a bad idea.


ljharb (approved):
This is kind of universal advice for jest mocking, tho, so imo it's something jest users just need to know (and jest needs to document)

PR #2029, opened 2022-05-30
feat: Add nonce property for security in SSR

ntucker:
### Motivation

When embedding a script; validity of that script can be checked against a nonce. Let's provide an option to provide that.


notwillk (approved):
Lgtm!

PR #2039, opened 2022-06-03
fix: Hydration mismatch in React 17

### LINKED ISSUE #2037: Hydration mismatch when using with Next.js
**React version** (17.0.2)

**Package manager version** (npm 6)

**Node version** (14.19.1)

**Rest Hooks version** (6.3.4)

image

**Describe the bug**
Using with Next.js produces a hydration mismatch error. Simply place `CacheProvider` in `pages/_app.js`:

image

As recommended by Next to globally introduce providers or layouts. Then create a random page with some div:

image

**To Reproduce**
1. Install Next.js 12.1.5
2. Put `CacheProvider` in `pages/_app.js` as in the picture
3. Create any other page with some div ex. `pages/index.js`
4. Visit localhost to see the error (development environment - `npm run dev`). 

**Expected behavior**
No hydration mismatch as in the source code - I looked at the source and it seems like if we use React 16 or 17 and is on the server, the `Suspense` component is not injected - it is only injected via client - the problem seems to be that any page in Next refuses to be rendered as proper HTML on server, whereas it renders correctly as a div after hydration on client - hence the server html lacks a div.

**Additional context**
`Warning: Did not expect server HTML to contain a <div> in <div>. next-dev.js:25:25
    error next-dev.js:25
    React 18
    renderReactElement index.js:526
    doRender index.js:852
    _callee$ index.js:421
    Babel 8
    render index.js:412
    _callee$ index.js:407
    Babel 14
    hydrate index.js:291
    <anonymous> next-dev.js:45
    (Async: promise callback)
    <anonymous> next-dev.js:40
    NextJS 7
`

**Minimum reproducable example or test case**

ntucker:
Fixes #2037

### Motivation

To deal with Legacy react not rendering below suspense boundaries during SSR - we won't render them. However, this creates a hydration mismatch warning, so we need to identify this case client-side and change behavior appropriately.

### Solution

Injecting scripts introduces all sorts of troubles with CSP; so we're just adding an element with a specific ID to detect the path.

I couldn't easily find a solution to actually test hydration - so I just tested the SSR node-side.

### DIFF
--- a/packages/core/src/react-integration/provider/BackupBoundary.tsx
+++ b/packages/core/src/react-integration/provider/BackupBoundary.tsx
@@ -1,15 +1,26 @@
 import React, { Suspense, memo, useMemo, version } from 'react';
 
 /* istanbul ignore next  */
-const NoSuspense =
-  (version.startsWith('16') || version.startsWith('17')) &&
-  typeof window === 'undefined';
+const LegacyReact = version.startsWith('16') || version.startsWith('17');
+/* istanbul ignore next  */
+const SSR = typeof document === 'undefined';
+const SSR_ID = 'rest-hooks-SSR';
+/* istanbul ignore next  */
+const HYDRATED = !SSR && document.getElementById(SSR_ID);
 
 // since Suspense does not introduce DOM elements, this should not affect rehydration.
 const BackupBoundary: React.FunctionComponent<{ children: React.ReactNode }> =
   /* istanbul ignore if */
-  NoSuspense
-    ? /* istanbul ignore next  */ ({ children }) => children as JSX.Element
+  LegacyReact && SSR
+    ? /* istanbul ignore next  */ ({ children }) => (
+        <div id={SSR_ID}>{children}</div>
+      )
+    : /* istanbul ignore if */ HYDRATED
+    ? /* istanbul ignore next  */ ({ children }) => (
+        <Suspense fallback={<div id={SSR_ID}></div>}>
+          <div id={SSR_ID}>{children}</div>
+        </Suspense>
+      )
     : ({ children }) => <Suspense fallback={<Loading />}>{children}</Suspense>;
 
 export default memo(BackupBoundary);
--- a/packages/core/src/react-integration/provider/__tests__/backupboundary.node.tsx
+++ b/packages/core/src/react-integration/provider/__tests__/backupboundary.node.tsx
@@ -1,0 +1,33 @@
+// eslint-env jest
+import React, { version } from 'react';
+import { renderToString } from 'react-dom/server';
+
+import BackupBoundary from '../BackupBoundary';
+
+describe('<BackupBoundary />', () => {
+  let warnspy: jest.SpyInstance;
+  beforeEach(() => {
+    warnspy = jest.spyOn(global.console, 'warn');
+  });
+  afterEach(() => {
+    warnspy.mockRestore();
+  });
+
+  it('should warn users about missing Suspense', () => {
+    const tree = (
+      <BackupBoundary>
+        <div>hi</div>
+      </BackupBoundary>
+    );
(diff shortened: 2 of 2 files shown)


packages/core/src/react-integration/provider/BackupBoundary.tsx
```diff
@@ -1,15 +1,26 @@
 import React, { Suspense, memo, useMemo, version } from 'react';
 
 /* istanbul ignore next  */
-const NoSuspense =
-  (version.startsWith('16') || version.startsWith('17')) &&
-  typeof window === 'undefined';
+const LegacyReact = version.startsWith('16') || version.startsWith('17');
```
ljharb, line 4: the NoSuspense check, and this one, are super brittle - it'd be much better to feature-detect rather than useragent-sniff, if at all possible.

packages/core/src/react-integration/provider/BackupBoundary.tsx
```diff
@@ -1,15 +1,26 @@
...
+/* istanbul ignore next  */
+const SSR = typeof document === 'undefined';
```
ljharb, line 6: i'd probably use `window` but this should be fine too

packages/core/src/react-integration/provider/BackupBoundary.tsx
```diff
@@ -1,15 +1,26 @@
...
+const SSR_ID = 'rest-hooks-SSR';
```
ljharb, line 7: what happens if there's more than one render tree on the page?

ljharb (approved):

packages/core/src/react-integration/provider/BackupBoundary.tsx
ntucker: I agree feature detection is much better; but I can't find any indicators on google, and using the actual feature itself (by trying to render suspense) would break everything so that wouldn't work either. I assume when you say 'useragent sniff' you are drawing and analogy to browser behavior and are not actually thinking that the 'version' export of react is a browser user-agent.

packages/core/src/react-integration/provider/BackupBoundary.tsx
ntucker: I figured since I was actually using document for the condition this checks on it would be slightly more reliable.

packages/core/src/react-integration/provider/BackupBoundary.tsx
ljharb: lol yes, exactly what i meant :-)

def not a blocker if there's no better way for now, ofc

packages/core/src/react-integration/provider/BackupBoundary.tsx
ljharb: good point!

packages/core/src/react-integration/provider/BackupBoundary.tsx
ntucker: and only one does SSR and the other does not? well the only differences is a div inserted with a specific ID so while this would 'change behavior' otherwise I don't see it actually breaking anything as a non-SSR thing couldn't care less exactly what is rendered.

In other words - we must make sure SSR and client match; in this worst case scenario we simply lose some debugging capability but there are no errors.

ntucker: The whole point of the backup boundary is for people who forget to put suspense somewhere

packages/core/src/react-integration/provider/BackupBoundary.tsx
ntucker: Given that these behaviors with major versions are pretty well defined; and multiple react versions are not possible because they break so many other things - I don't see this actually being much of an issue.

PR #2049, opened 2022-06-10
docs: Version rest/graphql seperate from main

ntucker:
This moves rest+graphql out of main version line since they are decoupled.


ljharb (approved):
just docs and website changes

PR #2055, opened 2022-06-22
fix: Incoming entities should always be processed if existing fails validation

ntucker:
### Motivation

This can happen with partial responses - requiring a merge even with 'newer' version


ntucker:
See how this interfaces with #1403 before implementation

ntucker:
Abandoning in favor of https://resthooks.io/rest/api/schema.Entity#shouldReorder

PR #2056, opened 2022-06-23
enhance: Array schema pushes through 'found'

ntucker:


ntucker:
This has no meaning as failing validation marks 'deleted', which has higher prescedence than not found. Furthermore, array infers undefined all the time therefore found will never be used here.

PR #2129, opened 2022-08-12
feat: RestEndpoint, createResource

### LINKED ISSUE #2128: RFC: Resource has an Entity (@rest-hooks/rest v6)
# Motivation

## Existing design

Resource *is* an Entity. Endpoints are defined as static members.

The motivation is for brevity: This allows one import to both define the expected type as well as access the endpoints to send as hook 'subjects'.

## Problems

1. Class static side is not well supported by TypeScript. This leads to the somewhat confusing but also limiting [generic workaround](https://resthooks.io/rest/guides/rest-types).
2. Inheritance does not work well for providing out-of-the-box endpoint definitions. Overrides are better
    - It's a struggle between general types that allow overrides or precise types that help developers.
    - Hacks like ‘SchemaDetail’ are an attempt around this but are confusing, expensive for typescript to compute and likely break in certain configurations.
3. Union Resources are awkward to define as their expected schema ends up not being the Entity.
    - In general, custom schemas are often shared by multiple endpoints, making it desirable to not require them to be just an Entity
4. Endpoints themselves don't maintain referential equality
    - This results in hacks in our hooks that violate expectations (ignoring referential changes to endpoints). There are legit reasons one might want to create a endpoint that changes and have that trigger fetches. 

# Solution: Resource Singletons

**Resource Is-a-entity → Has-a-entity**

Note: Entity being class is still fine as there is no type overrides expected.

### Simplest case

```tsx
class Todo extends Entity {
  readonly id: string = '';

  pk() { return this.id }
}
const TodoResource = new Resource(Todo).extend(
  CRUD('http\\://test.com/todos' as const)
)
// this is where the schema 'has a'
assert(TodoResource.schema === Todo);
// these are now all very specific. will error if any fields are missing  and not exactly matching
const todo = useSuspense(TodoResource.detail, { id: '5' });
const todos = useSuspense(TodoResource.list);
controller.fetch(TodoResource.create, { content: 'ntucker' })
controller.fetch(TodoResource.update, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.partialUpdate, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.delete, { id: '5' })
```

### Extra url parameter changes types

```tsx
class User extends Entity {
  readonly id: number | undefined = undefined;
  readonly username: string = '';
  readonly email: string = '';
  readonly isAdmin: boolean = false;

  pk() {
    return this.id?.toString();
  }
}

const UserResource = new Resource(User).extend(
  CRUD('http\\://test.com/groups/:group/users' as const, 'id' as const),
).extend(resource => ({
  detail: resource.detail.extend({ schema: User2 }),
  another: resource.endpoint.extend({ schema: { data: resouce.schema, next: '' } }),
  ban: resource.endpoint.extend({ urlRoot: 'http\\://test.com/groups/:group/users/:id/ban' as const, method: 'post', sideEffect: true });
}));

const user = useSuspense(UserResource.detail, { group: 'five', id: '5' });
const users = useSuspense(UserResource.list, { group: 'five' });
controller.fetch(UserResource.create, { group: 'five' }, { username: 'ntucker' })
controller.fetch(UserResource.update, { group: 'five', id: '5' }, { username: 'ntucker' })
controller.fetch(UserResource.partialUpdate, { group: 'five', id: '5' }, { username: 'ntucker' })
controller.fetch(UserResource.delete, { group: 'five', id: '5' })
```

In the extend function, all ‘protected’ members are exposed (like fetch(), endpoint, etc). However, the only public members to UserResource are the endpoints (and the schema) - making typeahead exceptionally discoverable.

### Entity Mixin

```tsx
// entity as mixin
class Todo {
  readonly id: string = '';
}
const TodoEntity = schema.Entity(Todo, todo => todo.id);
const TodoResource = new Resource(TodoEntity).extend(
  CRUD('http\\://test.com/todos' as const)
)

// need more entity lifecycles? just extend the mixin
class TodoEntity extends schema.Entity(Todo, todo => todo.id) {
  static useIncoming() {}
}
```

[[Use an existing class for a resource · Issue #416 · coinbase/rest-hooks](https://github.com/coinbase/rest-hooks/issues/416)](https://github.com/coinbase/rest-hooks/issues/416)

### Union

```tsx
class Feed extends Entity {
  readonly id: string = '';
  readonly title: string = '';
  readonly type: 'link' | 'post' = 'post';
  pk() { return this.id }
}
class FeedLink extends Feed {
  readonly url: string = '';
  readonly type: 'link' = 'link';
}
class FeedPost extends Feed {
  readonly type: 'post' = 'post';
}
const FeedUnion = schema.Union({post: FeedPost, link: FeedLink}, 'type' as const);
const FeedResource = new Resource(FeedUnion).extend(
  CRUD('http\\://test.com/feed' as const)
)
```

## Tradeoffs

This does make it a bit more verbose

```ts
class Todo extends Entity {
  readonly id: string = '';

  pk() { return this.id }
}
const TodoResource = new Resource(Todo).extend(
  CRUD('http\\://test.com/todos' as const)
)
```

vs

```ts
class TodoResource extends Resource{
  readonly id: string = '';

  pk() { return this.id }

  static urlRoot = 'http\\://test.com/todos' as const
}
```

However, in practice this seems like a small sacrifice due to the common need for endpoint overrides, and how verbose that it. This has the added bonus of using the extra semantics to make the concepts more clear - `Entity` is simply a data description, and Resource is a set of operations (usually networking) to perform.

## Alternatives considered

Rather than `Resource.extend()`, it might be more obvious to use an enhancer function (like a mixin). E.g., `CRUD('url')(new Resource(User))`. However, while this works fine for provided endpoint patterns like `CRUD` - it puts a burden on users doing custom overrides. Thus the 'extend()' method has been show to work well to automatically type, while providing override capabilities (see Endpoint).

## Open question

1. Change Resource.detail → Resource.get?
    1. Since this is an endpoint itself this makes more sense than before as calling get now actually does the ‘get’.
    6. Also Resource.list → Resource.getList?

ntucker:
Fixes #2128 .

# Motivation

## Existing design

Resource *is* an Entity. Endpoints are defined as static members.

The motivation is for brevity: This allows one import to both define the expected type as well as access the endpoints to send as hook 'subjects'.

## Problems

1. Class static side is not well supported by TypeScript. This leads to the somewhat confusing but also limiting [generic workaround](https://resthooks.io/rest/guides/rest-types).
2. Inheritance does not work well for providing out-of-the-box endpoint definitions. Overrides are better
    - It's a struggle between general types that allow overrides or precise types that help developers.
    - Hacks like ‘SchemaDetail’ are an attempt around this but are confusing, expensive for typescript to compute and likely break in certain configurations.
3. Union Resources are awkward to define as their expected schema ends up not being the Entity.
    - In general, custom schemas are often shared by multiple endpoints, making it desirable to not require them to be just an Entity
4. Endpoints themselves don't maintain referential equality
    - This results in hacks in our hooks that violate expectations (ignoring referential changes to endpoints). There are legit reasons one might want to create a endpoint that changes and have that trigger fetches. 
 
# Solution: Trinity of definitions

1. Schema: Data lifecycles
2. Endpoint: Networking lifecycles
3. Resource: Binding the two above - a collection of endpoints for a given Schema

```tsx
class Todo extends Entity {
  readonly id: string = '';

  pk() { return this.id }
}
const TodoResource = createResource('http\\://test.com/todos/:id' as const, Todo);

// these are now all very specific. will error if any fields are missing  and not exactly matching
const todo = useSuspense(TodoResource.get, { id: '5' });
const todos = useSuspense(TodoResource.getList);
controller.fetch(TodoResource.create, { content: 'ntucker' })
controller.fetch(TodoResource.update, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.partialUpdate, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.delete, { id: '5' })
```

## RestEndpoint definition

```ts
export interface RestInstance<
  F extends FetchFunction = FetchFunction,
  S extends Schema | undefined = any,
  M extends true | undefined = true | undefined,
> extends EndpointInstanceInterface<F, S, M> {
  readonly urlRoot: string;
  readonly fetchInit: RequestInit;
  readonly method: string;
  readonly signal: AbortSignal | undefined;

  /* fetch lifecycles */
  /* before-fetch */
  url(...args: Parameters<F>): string;
  getFetchInit(
    this: any,
    body?: RequestInit['body'] | Record<string, unknown>,
  ): RequestInit;
  /* after-fetch */
  fetchResponse(input: RequestInfo, init: RequestInit): Promise<Response>;
  parseResponse(response: Response): Promise<any>;
  process(value: any, ...args: Parameters<F>): any;

  extend<
    E extends RestInstance,
    O extends RestEndpointOptions<F> & Partial<Omit<E, keyof RestInstance>>,
  >(
    this: E,
    options: O,
  ): RestExtendedEndpoint<typeof options, E, F>;
}
```

### Base overrides for lifecycles

```ts
class GithubEndpoint<
  O extends RestGenerics = any,
> extends RestEndpoint<O> {
  getFetchInit(body: any): RequestInit {
    if (body) {
      return super.getFetchInit(deeplyApplyKeyTransform(body, snakeCase));
    }
    return super.getFetchInit();
  }

  async parseResponse(response: Response) {
    const results = await super.parseResponse(response);
    if (
      (response.headers && response.headers.has('link')) ||
      Array.isArray(results)
    ) {
      return {
        link: response.headers.get('link'),
        results,
      };
    }
    return results;
  }

  process(value: any, ...args: any) {
    return deeplyApplyKeyTransform(value, camelCase);
  }
}
```

### Default values

```ts
class IssueEndpoint<O extends RestGenerics=any> extends GithubEndpoint<O> {
  pollFrequency = 60000;
}
```

## Pagination

```ts
const MyResource.getNextPage = MyResource.getList.paginated(
  ({
    cursor,
    ...rest
  }: {
    cursor: string | number;
    group: string | number;
  }) => [rest],
);
```

# Tradeoffs

Providing an Endpoint class with defaults is incredibly annoying and non-obvious. This doesn't seem like an essential use-case so we're trading this to make other cases of extending endpoints easier:

```ts
class UserEndpoint<
  O extends Partial<RestGenerics> = {
    schema: DefaultUser;
    urlRoot: 'http\\://test.com/user/:id';
  },
> extends MyEndpoint<
  Defaults<
    O,
    { schema: DefaultUser; urlRoot: 'http\\://test.com/user/:id' }
  >
> {
  constructor(options: O & { name: string }) {
    super(options as any);
  }

  urlRoot = 'http\\://test.com/user/:id' as const;
}
````

What we get is simple extensions for Generics where defaults don't change:

```ts
class IssueEndpoint<O extends RestGenerics = any> extends GithubEndpoint<O> {
  pollFrequency = 60000;
}
```

### Super doesn't work with .extend()

This creates an inconsistency and requires more expert level knowledge of javascript; but is achievable:

```ts
const updateUser = new RestEndpoint({
  method: 'PUT' as const,
  urlRoot: 'http\\://test.com/user/:id' as const,
  name: 'get',
  schema: User,
  getFetchInit(body: any): RequestInit {
    if (body && isPojo(body)) {
      return RestEndpoint.prototype.getFetchInit.call(this, {
        ...body,
        email: 'always@always.com',
      });
    }
    return RestEndpoint.prototype.getFetchInit.call(this, body);
  },
});
```

This is deemed acceptable as we expect most lifecycle overrides to be in class definitions rather than endpoints themselves. However, it might be worth considering a pattern where calling super is less likely as well. For instance, this was never needed for React lifecycles.


ntucker:
test

ntucker:
asdf

ntucker:
testing comments

```ts
const a = 'abc';
```

ntucker:
testing comments

```ts
const a = 'abc';
```

ntucker:
test

ntucker:
test

ntucker:
test

notwillk (approved):

PR #2159, opened 2022-09-05
Make @rest-hooks/normalizr and @rest-hooks/endpoint disjoint

ntucker:
### Motivation

API definitions and usage should be completely disjoint - this enables distinct upgrade paths as well as eliminates problems from multiple versions of shared pieces.

#### @rest-hooks/normalizr is now for Endpoint users

This is a breaking change for normalizr as some exports are removed.

- normalize, denormalize functions
- interface compatibility definitions for those wanting to use (like Denormalize<T>, Normalize<T>)

#### @rest-hooks/endpoint is now for API definitions

In terms of exports there is no change, as endpoint used to be re-exporting these from normalizr.

- Endpoint
- All schema implementations

Shared interfaces are currently in interface.ts in each package. This could be split into another package but since it's interfaces only for now we'll just repeat the code.

This means all schema *implementations* will be moved into endpoint. All Endpoint *interfaces* will be exported from normalizr for users to use.

#### Shared interface exports

- Schema
- EntityInterface
- UnvisitFunction
- NormalizedIndex
- EndpointInterface
- Normalize
- NormalizeNullable
- Denormalize
- DenormalizeNullable
- AbstractInstanceTyp

#### TypeScript 3.4 support

Some of the legacy type files (`/ts3.4`) had been neglected, so they were updated.


notwillk (approved):
LgTM

PR #2187, opened 2022-09-23
feat: RestEndpoint, createResource, hookifyResource

ntucker:
BREAKING CHANGE: Removed BaseResource, Resource, HookableResource

# Motivation

## Existing design

Resource *is* an Entity. Endpoints are defined as static members.

The motivation is for brevity: This allows one import to both define the expected type as well as access the endpoints to send as hook 'subjects'.

## Problems

1. Class static side is not well supported by TypeScript. This leads to the somewhat confusing but also limiting [generic workaround](https://resthooks.io/rest/guides/rest-types).
2. Inheritance does not work well for providing out-of-the-box endpoint definitions. Overrides are better
    - It's a struggle between general types that allow overrides or precise types that help developers.
    - Hacks like ‘SchemaDetail’ are an attempt around this but are confusing, expensive for typescript to compute and likely break in certain configurations.
3. Union Resources are awkward to define as their expected schema ends up not being the Entity.
    - In general, custom schemas are often shared by multiple endpoints, making it desirable to not require them to be just an Entity
4. Endpoints themselves don't maintain referential equality
    - This results in hacks in our hooks that violate expectations (ignoring referential changes to endpoints). There are legit reasons one might want to create a endpoint that changes and have that trigger fetches. 
 
# Solution: Trinity of definitions

1. Schema: Data lifecycles
2. Endpoint: Networking lifecycles
3. Resource: Binding the two above - a collection of endpoints for a given Schema

```tsx
class Todo extends Entity {
  readonly id: string = '';

  pk() { return this.id }
}
const TodoResource = createResource('http\\://test.com/todos/:id' as const, Todo);

// these are now all very specific. will error if any fields are missing  and not exactly matching
const todo = useSuspense(TodoResource.get, { id: '5' });
const todos = useSuspense(TodoResource.getList);
controller.fetch(TodoResource.create, { content: 'ntucker' })
controller.fetch(TodoResource.update, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.partialUpdate, { id: '5' }, { content: 'ntucker' })
controller.fetch(TodoResource.delete, { id: '5' })
```

## RestEndpoint definition

```ts
export interface RestInstance<
  F extends FetchFunction = FetchFunction,
  S extends Schema | undefined = any,
  M extends true | undefined = true | undefined,
> extends EndpointInstanceInterface<F, S, M> {
  readonly urlRoot: string;
  readonly fetchInit: RequestInit;
  readonly method: string;
  readonly signal: AbortSignal | undefined;

  /* fetch lifecycles */
  /* before-fetch */
  url(...args: Parameters<F>): string;
  getFetchInit(
    this: any,
    body?: RequestInit['body'] | Record<string, unknown>,
  ): RequestInit;
  /* after-fetch */
  fetchResponse(input: RequestInfo, init: RequestInit): Promise<Response>;
  parseResponse(response: Response): Promise<any>;
  process(value: any, ...args: Parameters<F>): any;

  extend<
    E extends RestInstance,
    O extends RestEndpointOptions<F> & Partial<Omit<E, keyof RestInstance>>,
  >(
    this: E,
    options: O,
  ): RestExtendedEndpoint<typeof options, E, F>;
}
```

### Base overrides for lifecycles

```ts
class GithubEndpoint<
  O extends RestGenerics = any,
> extends RestEndpoint<O> {
  getFetchInit(body: any): RequestInit {
    if (body) {
      return super.getFetchInit(deeplyApplyKeyTransform(body, snakeCase));
    }
    return super.getFetchInit();
  }

  async parseResponse(response: Response) {
    const results = await super.parseResponse(response);
    if (
      (response.headers && response.headers.has('link')) ||
      Array.isArray(results)
    ) {
      return {
        link: response.headers.get('link'),
        results,
      };
    }
    return results;
  }

  process(value: any, ...args: any) {
    return deeplyApplyKeyTransform(value, camelCase);
  }
}
```

### Default values

```ts
class IssueEndpoint<O extends RestGenerics=any> extends GithubEndpoint<O> {
  pollFrequency = 60000;
}
```

## Pagination

```ts
const MyResource.getNextPage = MyResource.getList.paginated(
  ({
    cursor,
    ...rest
  }: {
    cursor: string | number;
    group: string | number;
  }) => [rest],
);
```

# Tradeoffs

Providing an Endpoint class with defaults is incredibly annoying and non-obvious. This doesn't seem like an essential use-case so we're trading this to make other cases of extending endpoints easier:

```ts
class UserEndpoint<
  O extends Partial<RestGenerics> = {
    schema: DefaultUser;
    urlRoot: 'http\\://test.com/user/:id';
  },
> extends MyEndpoint<
  Defaults<
    O,
    { schema: DefaultUser; urlRoot: 'http\\://test.com/user/:id' }
  >
> {
  constructor(options: O & { name: string }) {
    super(options as any);
  }

  urlRoot = 'http\\://test.com/user/:id' as const;
}
````

What we get is simple extensions for Generics where defaults don't change:

```ts
class IssueEndpoint<O extends RestGenerics = any> extends GithubEndpoint<O> {
  pollFrequency = 60000;
}
```

### Super doesn't work with .extend()

This creates an inconsistency and requires more expert level knowledge of javascript; but is achievable:

```ts
const updateUser = new RestEndpoint({
  method: 'PUT' as const,
  urlRoot: 'http\\://test.com/user/:id' as const,
  name: 'get',
  schema: User,
  getFetchInit(body: any): RequestInit {
    if (body && isPojo(body)) {
      return RestEndpoint.prototype.getFetchInit.call(this, {
        ...body,
        email: 'always@always.com',
      });
    }
    return RestEndpoint.prototype.getFetchInit.call(this, body);
  },
});
```

This is deemed acceptable as we expect most lifecycle overrides to be in class definitions rather than endpoints themselves. However, it might be worth considering a pattern where calling super is less likely as well. For instance, this was never needed for React lifecycles.


ntucker:
Changed createResource to take named arguments. This has the advantage of specifying the 'path' as 'readonly' making it automatically not widen the type. This eliminates the need for 'as const', which is a much less obvious syntax to use. Additionally it makes adding future arguments much more straightforward.

ntucker:
Was able to eliminate the need for any 'as const' (constructor and .extend()) by wrapping the generic with Readonly<>

notwillk (approved):
Lgtm

PR #2266, opened 2022-11-09
enhance: Add searchParams arg for RestEndpoint typing

ntucker:
### Motivation

Same functionality as `body` for the searchParams.

### Solution

Add type only argument to RestEndpoint construction and .extends() arguments.


notwillk (commented):
I like this yarn.lock way better than the old one!

notwillk (approved):

PR #2273, opened 2022-11-14
feat(react-native): Revalidate stale on react-navigation focus

ntucker:
### Motivation

For some reason react navigation keeps all components around even after they are not on screen. This means navigating back will continue showing stale data.

I recall this issue when CB was using rest hooks in their consumer RN app.

### Solution

Treat focus events like re-mounting components on web: we revalidate stale at this point.

This code only loads in react native. This code only has effect when used in react-navigation context. Not using either of these continues previous behavior.

Most of the effort was put into understanding how to add tests for react-navigation specific functionality.

#### useFocusEffect.native.ts

This handles async loading of react-navigation as we don't want to require it. Due to this, we add to sideeffects in package.json.

#### useSuspense.native.ts

Completely identical to web version, but with added focus handler:

```ts
 useFocusEffect(() => {
    // revalidating non-suspending data is low priority, so make sure it doesn't stutter animations
    const task = InteractionManager.runAfterInteractions(() => {
      if (Date.now() > expiresAt && key) {
        controller.fetch(endpoint, ...(args as readonly [...Parameters<E>]));
      }
    });

    return () => task.cancel();
  }, []);
```


notwillk (approved):
Glad to see a fox for this!

Issue #2051, opened 2022-06-10
useSuspense always displays fallback in next.js 12.1.6


rtorrente:
**React version** : 18.1.0

**Package manager version**  : yarn : 1.22.17

**Node version** : v16.13.1

**Rest Hooks version** 6.3.4

**Next JS Version** 12.1.6

**Describe the bug**

I'm trying to use next.js for the first time when creating a new app and I’ve got a strange bug when using useSuspense.
If the website is already loaded (e.g. I'm on the home page) and I navigate (with Next Router) to a page with useSuspense, it works. I have my data loaded without problem.
But if I go directly to this page (or refresh this page), I have the suspense fallback component stuck on the screen (and the page will always be showing the loading state).

I think there is a problem with SSR but I don't really need SSR for this page, so there is no problem if SSR sends the fallback since the data can be retrieved on the client side.

Has anyone successfully set up rest-hooks with next.js using suspense?

ntucker:
Am I understanding correctly that even after the page loads (showing the fallback due to lack of SSR) it still doesn't load the data?

rtorrente:
@ntucker exactly

For concrete example, I put LinearProgress material ui component as a suspense fallback and if I access the page directly, LinearProgress run forever (and browser shows circular loading state indicates that page isn't fully loaded)

ntucker:
From the betas to final 18, React changed the nature of SSR - so it's basically avoiding re-renders when the data comes in. I'm working on a solution to this, but was traveling in Europe for the past 3 weeks. Will have designs posted; but figured I should update here.

ntucker:
Confirmed fixed in 6.5.0-beta.1. Will be releasing full version within next week.

ntucker:
Released in [6.5](https://resthooks.io/blog/2022/11/09/Query-getState-SSR-partial-hydration) @rtorrente

ntucker:
The latest version of NextJS supports React suspense, thus a [guide for SSR](https://resthooks.io/docs/guides/ssr) has been added. It includes NextJS as well as React 18 usage with express.

PR #2288, opened 2022-11-19
feat: Add nextjs helpers and docs

ntucker:
Helps #2051

### Motivation

NextJS is popular

### Solution

Lots of investigation lead to Document to control SSR, and App to control client-side.

Added an example to provide a stackblitz reference as well as validate (unit tests don't really work in these cases).

### Open Questions

How does nonce work with NextJS? Their docs are so sparse


ntucker:
@rtorrente this should make nextjs integration a piece of cake!

rtorrente:
@ntucker Amazing ! It looks very promising with a very easy integration of SSR !!
I will try to take a closer look soon

notwillk (approved):

PR #2294, opened 2022-11-22
pkg(security): CVE-2022-39353

pavkcon:
### Descriptions
xmldom parses XML that is not well-formed because it contains multiple top level elements, and adds all root nodes to the `childNodes` collection of the `Document`, without reporting any error or throwing.
This breaks the assumption that there is only a single root node in the tree, which led to https://nvd.nist.gov/vuln/detail/CVE-2022-39299 and is a potential issue for dependents.

Fixes # `< 0.7.7` `@xmldom/xmldom`


ntucker (approved):
Thanks for the contribution!

PR #2302, opened 2022-11-29
fix: Network down/ CORS errors should be 'soft' by default

ntucker:
### Motivation

These are potentially recoverable so we want to retry when that logic is added, but also keep existing data if it is still valid.

### Solution

Change artificial status from 400 to 500


notwillk (approved):
500 > 400

PR #2309, opened 2022-12-03
enhance: Include Object.hasOwn polyfill

### LINKED ISSUE #2304: Object.hasOwn is not a function on safari v14.1
**React version**: 18.2.0

**Next.js version**: 13.0.1

**TypeScript version**: 4.8.4

**Rest Hooks version**: `rest-hooks` 6.4.2 and `@rest-hooks/rest` 5.2.0

**Package manager version**: yarn 1.22.19

**Node version**: 16.16.0

**Describe the bug**

Hi,

On some old safari (14.1.2), I encounter this runtime error: Object.hasOwn is not a function (from NetworkManager.js), which I think comes from this commit https://github.com/coinbase/rest-hooks/commit/bdbc6a49350cae24a9d8cda0d4e360ce20cb91cd. I'm also not sure why this problem exists since it's supposed to be handled by some transcompiler.

I know Object.hasOwn is too new for safari v14.1, but I'm looking for a solution without upgrading the browser.

hasOwn

My tsconfig has `esnext` in `compilerOptions.lib` and I've tried using `babel-plugin-transform-object-hasown` but it didn't solve the problem.

I've also tried to downgrade rest hooks to some versions earlier than the commit above (https://github.com/coinbase/rest-hooks/commit/bdbc6a49350cae24a9d8cda0d4e360ce20cb91cd), e.g. `rest-hooks 6.3.11` and `@rest-hooks/rest 5.1.1` (or even older versions), but it seems that `@rest-hooks/core 3.4.0` is always installed (in node_modules) and thus the problem still persists. (How can I install rest hooks without having the Object.hasOwn syntax inside? Maybe it could be one of the easiest solutions.)

**To Reproduce**

The problem is from a private repo so I can't share it here. I'm using next.js v13 with ts v4.8. I think as long as Object.hasOwn is somewhere in the installed dependencies the error will be thrown. 

**Expected behavior**

No this `Object.hasOwn` error on Safari v14.1.

---

P.S. Thanks for this awesome project!

ntucker:
Fixes #2304 .

### Motivation

Our `exports` target is currently 2020 and later browsers. Sadly Safari is late to the party, so we still need Object.hasOwn.

### Solution

- Remove 'babel-plugin-transform-object-hasown' even for legacy builds as this copies a function in multiple places
- Add relevant core-js import to files that use Object.hasOwn
- Use `entry` in babel config to ensure these core-js imports are only added when necessary.
- Add core-js as a dependency on a version that includes said polyfill
- Slightly move up @babel/runtime version
- Add note about polyfills to install docs

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -9,6 +9,7 @@ module.exports = function (api) {
         {
           typing: 'typescript',
           loose: true,
+          useBuiltIns: 'entry',
           resolver: {
             extensions: ['.ts.', '.tsx', '.js', '.jsx', '.es', '.es6', '.mjs'],
             resolvePath(sourcePath, currentFile, opts) {
@@ -26,10 +27,6 @@ module.exports = function (api) {
         },
       ],
     ],
-    plugins: [
-      process.env.BROWSERSLIST_ENV !== '2020' &&
-        'babel-plugin-transform-object-hasown',
-    ].filter(p => p),
     assumptions: {
       noDocumentAll: true,
       noClassCalls: true,
--- a/docs/core/getting-started/installation.md
+++ b/docs/core/getting-started/installation.md
@@ -100,7 +100,16 @@ export default function App({ Component, pageProps }: AppProps) {
 
 Alternatively [integrate state with redux](../guides/redux.md)
 
-<details><summary><b>Legacy (IE) browser support</b></summary>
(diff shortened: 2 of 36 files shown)


notwillk (approved):
Looks so nice, I'm gonna approve it twice

notwillk (approved):

PR #2414, opened 2023-02-06
feat: Add schema.Entity() a mixin

### LINKED ISSUE #416: Use an existing class for a resource
Hello,

I have a normal class, can I use it to create a resource? Something like this?

```typescript
export class Demo {
  id: number,
  value: string
}
```

```typescript
export default class DemoResource extends Resource<Demo> {
 
  pk() {
    return this.id?.toString();
  }

  static urlRoot = 'http://test.com/demo/';
}
```

ntucker:
Fixes #416 .

### Solution

```ts
class User {
  username = '';
  id = '';
  createdAt = new Date(0);
}
```

#### Extending

```ts
class UserSchema extends schema.Entity(User) {
  pk() { return this.id }

  static schema = {
    createdAt: Date,
  }
}
```

#### Options

```ts
const UserSchema = schema.Entity(User, {
  pk() {
    return this.id;
  },
  key: 'User',
  schema: { createdAt: Date },
});
```

```ts
const UserSchema = schema.Entity(User, {
  pk: 'id',
  key: 'User',
  schema: { createdAt: Date },
});
```

### TODO:
- [ ] enforce constraint on class passed in works with pk - and default to id (so it will force id to exist)
- [ ] change mixin to use 'new' to be consistent with other schemas

### DIFF
--- a/packages/endpoint/src/schema.d.ts
+++ b/packages/endpoint/src/schema.d.ts
@@ -19,6 +19,11 @@ import type {
   EntityMap,
 } from './normal.js';
 import { default as Delete } from './schemas/Delete.js';
+import {
+  EntityOptions,
+  IEntityClass,
+  IEntityInstance,
+} from './schemas/EntitySchema.js';
 
 export { Delete, EntityMap };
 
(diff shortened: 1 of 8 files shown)


ntucker:
Closing in favor of #2449

PR #2417, opened 2023-02-07
feat: Entity.createIfValid() lifecycle

ntucker:
### Motivation

Faster is more better

### Solution

After adding a launch profile and asking ChatGPT - I was able to produce a cpu profile.

#### Rearranging denormalize construction (Entity.createIfValid())

Wrapping unvisit is not only awkward, it makes v8 not optimize the function, which is called often. While profiling this was in the top 7 in self-time.

> Benchmarks consistently show about a <strike>30%</strike> 10% performance increase for denormalize step

The original intent was to reduce the surface area of the interface between entity and denormalize/normalize. However, this is deceptive as it's still exposing an interface in that weird unvisit object. Since unvisit is exposed to other schemas, not just entity this could be considered an even larger surface areas as well.

To make it more clear we introduce a new lifecycle which allows us to go back to the [original algo](https://github.com/paularmstrong/normalizr/blob/master/src/index.js#L78) by constructing the instance in unvisitEntity (denormalize) itself.

We also are moving the validation call inside this lifecycle, as we must use the processedEntity, rather than the instance for this computation (to deal with missing fields). These two things cleans up Entity.denormalize, which is probably a good thing as well.

Also, when using the new lifecycle, immutable inputs now result in the instance being created. Previously it would retain the immutable map type, which is not the type we say is returned, nor likely what anyone expects.

#### Entity.denormalize no longer overrides instance defaults with `undefined` when it is missing

We previously set entity members with a schema that were missing to undefined explicitly. However, this mismatches the expectations that default values set in an entity should be used (and thus breaking types). `undefined` can still be achieved by setting it as a default, or having no default at all.

While technically a breaking change, this is a bug since it mismatched types, so if anyone was relying on this behavior they would have already run into issues.

#### pk === 'undefined'

It's easy to simply write `${this.id}` for a pk. In which case, if the field is missing, it will serialize to 'undefined'. This obviously means the member needed for a proper pk was missing, thus will likely result in confusing bugs as entities will all map to the same value. Now it simply properly realizes this is missing a proper PK and short circuits immediately like the other cases.

#### repeated schema.key calls

`schema.key` is used numerous times in a given function, so let's just make a local variable which doesn't take many bytes, and prevents more expensive calls from repeating. This is likely optimized away to something faster as well. Furthermore, the `this.name` default implementation is actually not super-duper fast so it should have some minor wins.

### Open questions

Calling it Entity.clone() for now, since the [original line](https://github.com/paularmstrong/normalizr/blob/master/src/index.js#L78) called the return value `entityCopy`. However, I'm not sure I like the name. We can't use fromJS() because we could return undefined (to deal with validation case)

### Entity.createIfValid() compatibility plan

Since we want to keep support between disjoint consumer/producers (normalizr vs endpoint packages); we will ensure two-way compatibility with previous behaviors until future breaking changes.

#### normalizr

We retain UnvisitFunction oddity. We follow this path when we don't detect the new lifecycle (which is optional in our types).

The code here was actually cleaned up a bit to make the other path simpler, which is a nice benefit.

#### endpoint

We retain a large if-case for when the unvisit function has the added callback. In this case, we use the previous behavior.

#### testing

legacy package includes two derivative implementations with their own test suites, so these legacy cases should still be reasonably covered, as well as ensuring extension of entities still work as expected

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -6,7 +6,7 @@ on:
       - master
     paths:
       - 'packages/normalizr/**'
-      - 'packages/endpoint/schemas/**'
+      - 'packages/endpoint/src/schemas/**'
       - 'packages/core/**'
       - 'examples/benchmark/**'
   push:
@@ -14,7 +14,7 @@ on:
       - master
     paths:
       - 'packages/normalizr/**'
-      - 'packages/endpoint/schemas/**'
+      - 'packages/endpoint/src/schemas/**'
       - 'packages/core/**'
       - 'examples/benchmark/**'
 jobs:
--- a/docs/graphql/api/Entity.md
+++ b/docs/graphql/api/Entity.md
@@ -76,7 +76,7 @@ import Lifecycle from '../diagrams/\_entity_lifecycle.mdx';
 
 ### static fromJS(props): Entity {#fromJS}
 
-Factory method called during denormalization. Use this instead of `new MyEntity()`
+Factory method that copies props to a new instance. Use this instead of `new MyEntity()`
 
 ### process(input, parent, key): processedEntity {#process}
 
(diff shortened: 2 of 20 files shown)


ntucker:
## Council notes

Reference: https://jorgenringen.github.io/2018/01/factory_method_naming_conventions/

- `createIfValid()`

PR #2442, opened 2023-02-14
feat: Interceptor response 'this' can be used to track 'server state'

ntucker:
### Motivation

To provide more realistic mocking scenarios, it's critical that operations are able to reflect server state updates. This is necessary when the proper simulated response cannot be computed simply from the request sent, but also requires an understanding of previous requests.

For instance, on the docs site, the [server timing variation example](https://resthooks.io/rest/guides/optimistic-updates#server-timings) will not work without persisting counter state.

### Solution

The state should be tracked per mocking instance, thus we need to send an object that can be mutated. `this` seems like an intuitive case as the interceptor is operating on the mocking instance.

### DIFF
--- a/docs/core/api/MockResolver.md
+++ b/docs/core/api/MockResolver.md
@@ -3,12 +3,10 @@ title: '<MockResolver />'
 ---
 
 ```typescript
-function MockResolver({
-  children,
-  fixtures,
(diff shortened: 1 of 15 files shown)


0xcaff:
tests

PR #2504, opened 2023-03-19
internal: Skip website, docs dirs for checkout

ntucker:
### Motivation

Checkout is a bottleneck in circleci and currently takes 30s

### Solution

website and docs dir are not needed. They also have lots of files and some assets (larger size) in them.


ntucker:
circleci doesn't actually have a way of configuring this

PR #2593, opened 2023-04-18
feat: Add schema.Collection

ntucker:
Fixes #2566 .

## Motivation

[Atomic mutations](https://resthooks.io/docs/concepts/atomic-mutations) are one of the core values of Rest Hooks. However, while update was in the first pre-release, and delete quickly added for version 1, create has always been a [bit awkward](https://resthooks.io/rest/api/RestEndpoint#update).

**Often creating a new element you are on a view where you want to see it added to a list somewhere.**

One of the challenges here is that there can be many distinct lists you view. Furthermore, where something is added to a list isn't always the same.

## Previously

https://resthooks.io/docs/concepts/atomic-mutations#create

```ts
const createUser = new Endpoint(postToUserFunction, {
  schema: User,
  update: (newUserId: string) => ({
    [userList.key()]: (users = []) => [newUserId, ...users],
  }),
});
```

Problems:

- Very awkward code we force users to write
- Users only had access to the pk, rather than full entity
- Only works on top level results (doesn't work with nesting)
- Forces users to enumerate all lists they might want to add - even if they aren't in store
  - Which means they may create a list that hasn't been populated

### TODO

- [x] Test Values
- [x] Test addWith
- [x] Test pagination
- [x] Add simpler cursor name pagination
- [x] Test nested collections
- [x] Unions
- [ ] .insert

### Collections: Entities but for Arrays and Values instead of classes.

They literally are entities, so they get normalized as such. They just have different lifecycle implementations.

Inspiration: [Backbone collections](https://backbonejs.org/#Collection-unshift)

### Usage

```ts
class Todo extends IDEntity {
  userId = 0;
  title = '';
  completed = false;

  static key = 'Todo';
}

class User extends IDEntity {
  name = '';
  username = '';
  email = '';
  todos: Todo[] = [];

  static key = 'User';
  static schema = {
    todos: new schema.Collection([Todo], {
      nestKey: (parent, key) => ({
        userId: parent.id,
      }),
    }),
  };
}

const UserResource = createResource({
  path: '/users/:id',
  schema: User,
});
const TodoResource = createResource({
  path: '/todos/:id',
  searchParams: {} as { userId?: string } | undefined,
  schema: Todo,
});
```

```ts
const handleCreate = data => ctrl.fetch(TodoResource.getList.push, { userId: currentuser.id }, data);
```

This will add to these if they are already in store (but only if they already exist):

- User({ id: currentUser.id }).todos
- getTodo({ userId: currentUser.id })
- getTodo()

#### construction

Takes Array or Values as first arg. Second arg is options object.

```ts
const userTodos = new schema.Collection(new schema.Array(Todo), {
  argsKey: ({ userId }: { userId: string }) => ({
    userId,
  }),
});
```

Options:

Must provide `argsKey` *or* `nestKey` to compute a serializable object to be used in PK table. argsKey takes ...args and nestKey takes (parent,key) as inputs.

`createCollectionFilter` is optional and can be used to override the default filter algorithm. This is used by the "adders" as they inherit from the collection.

```ts
const defaultFilter =
  (urlParams: Record<string, any>, body?: Record<string, any>) =>
  (collectionKey: Record<string, any>) =>
    Object.entries(collectionKey).every(
      ([key, value]) =>
        key.startsWith('order') ||
        urlParams[key] === value ||
        body?.[key] === value,
    );
```

#### addWith(merge, createCollectionFilter)

This is the general `collection extender` method. All specializations simply call this method. This is also used in the paginator as it has a custom algorithm.

`merge` overrides the `Entity.merge()` algorithm. This is used to determine where new values are placed (end, beginning, somewhere in the middle?)

`createCollectionFilter` is optional and will override the inherited function from the Collection itself.

#### push/unshift/assign

These all call `addWith()` with variations on merge to place newly created objects at the beginning or end respectively for Arrays. `assign` is for Values() and simply merges (there is no concept of order in Values)

#### RestEndpoint adder specializations

For a `RestEndpoint` containing a collection schema, the extenders `push/unshift/assign` can be used as convenience to create an appropriate endpoint. They use the parent schema, but us their respective `addWith()` variations. Additionally this sets the `method` to "POST".

### Pagination

`.paginated()` now detects whether there is a collection or array. With collections it will use a custom extender. Arrays will work via update.

**new** `.paginated('cursor')` - 99% of paginations simply extract one field from the args. Let the user specify that field name rather than writing a whole function.

```ts
const handleLoadNextPage = () => ctrl.fetch(TodoResource.getList.paginated('cursor'), {cursor});
```

### Normalizr changes

`args` are now part of both normalization and denormalization. Of course these will default to an empty array in case users don't need args. However, they will need to be passed on my implementations such as schema.normalize(). This shouldn't be a problem as schema implementations without it wont be expecting it anyway.

`pk()` now takes a fourth argument - args. 

### next/createResource

- Uses collections instead of arrays
  - create uses getList.push
- Added new options
  - searchParams
    - `searchParams` only applies to getList and create.
  - body
    - `body` only applies to update,create,partialUpdate
  - urlPrefix
  - requestInit
  - getHeaders
  - getRequestInit
  - fetchResponse
  - parseResponse

### DIFF
--- a/.changeset/big-horses-begin.md
+++ b/.changeset/big-horses-begin.md
@@ -1,0 +1,6 @@
+---
+'@rest-hooks/rest': minor
+---
+
+Added RestEndpoint.paginated('cursor').
+Place the name of the pagination field instead of writing a function
--- a/.changeset/cuddly-horses-peel.md
+++ b/.changeset/cuddly-horses-peel.md
@@ -1,0 +1,5 @@
+---
+'@rest-hooks/normalizr': major
+---
+
+normalize/denormalize both take args array
--- a/.changeset/fluffy-students-raise.md
+++ b/.changeset/fluffy-students-raise.md
@@ -1,0 +1,7 @@
+---
+'@rest-hooks/normalizr': minor
+'@rest-hooks/react': minor
+'@rest-hooks/core': minor
+---
+
+Support Collections
--- a/.changeset/giant-teachers-buy.md
+++ b/.changeset/giant-teachers-buy.md
@@ -1,0 +1,5 @@
+---
+'@rest-hooks/rest': minor
+---
+
+Add RestEndpoint collections extenders: push/unshift/assign/insert
--- a/.changeset/grumpy-walls-compare.md
+++ b/.changeset/grumpy-walls-compare.md
@@ -1,0 +1,10 @@
+---
+'@rest-hooks/normalizr': minor
+'@rest-hooks/core': minor
+'@rest-hooks/react': minor
+'@rest-hooks/endpoint': minor
+'@rest-hooks/rest': minor
+'rest-hooks/legacy': minor
+---
+
+Support using args[] during normalization/denormalization
--- a/.changeset/khaki-keys-wait.md
+++ b/.changeset/khaki-keys-wait.md
@@ -1,0 +1,7 @@
+---
+'@rest-hooks/normalizr': minor
+'@rest-hooks/endpoint': minor
+'@rest-hooks/rest': minor
+---
+
+Entity.pk() has a fourth argument: args[]
--- a/.changeset/khaki-toys-worry.md
+++ b/.changeset/khaki-toys-worry.md
@@ -1,0 +1,5 @@
+---
+'@rest-hooks/rest': minor
+---
+
+Add next/createResource: Uses Collections and has more options from RestEndpoint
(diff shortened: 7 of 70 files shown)


ntucker:
Will update with [insertions ](https://github.com/data-client/rest-hooks/discussions/2566#discussioncomment-5735908) in a follow up PR

PR #2642, opened 2023-05-15
feat: Add per-endpoint customization to createResource()

ntucker:
Fixes #2446 .

### Motivation

Simpler resource construction

### Solution

```ts
const UserResource = createResource({
  path: 'http\\://test.com/groups/:group/users/:id',
  schema: User,
  getList: {
    searchParams: {} as { isAdmin?: boolean },
  },
  partialUpdate: {
    getOptimisticResponse(snap, params, body) {
      return {
        id: params.id,
        ...body,
      };
    },
  },
  delete: {
    getOptimisticResponse(snap, params) {
      return params;
    },
  },
  endpoints: {
    current: { path: '/user' },
  },
});
```

```ts
UserResource.getList({isAdmin: true});
UserResource.current();
```

To modify existing endpoint options - use its name (like 'get' for 'get' endpoint).

To add additional endpoints - put in the 'endpoints' option. This will extend from the 'get' endpoint, allowing it to inherit the defaults of the resource.

### DIFF
--- a/.changeset/cuddly-eels-sniff.md
+++ b/.changeset/cuddly-eels-sniff.md
@@ -1,0 +1,7 @@
+---
+'@rest-hooks/rest': minor
+---
+
+Add resource endpoint customization options:
+  - each default endpoint can be customized using its name ('update' for 'update')
+  - additional endpoints can use the 'endpoints' option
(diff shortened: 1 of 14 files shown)


ntucker:
Closing in favor of #2715

PR #2999, opened 2024-04-08
feat: Support expo-router focus events

ntucker:
Fixes # .

### Solution

https://docs.expo.dev/router/reference/testing/

We may only need to add tests - it looks like they may unmount - be sure to investigate


ntucker:
validated expo router requires no additional work

PR #3037, opened 2024-04-28
internal: Cache react matrix dependencies to speedup test speed

ntucker:
### Motivation

Faster running tests.

### Solution

&gt; 30s is spent (half the time) on react17 installation. Let's add a cache for this resolution.


ntucker:
This method doesn't work at least.

PR #3218, opened 2024-09-13
pkg: Update path-to-regexp to v8

ntucker:
### Release Notes

<details>
<summary>pillarjs/path-to-regexp (path-to-regexp)</summary>

### [`v8.0.0`](https://redirect.github.com/pillarjs/path-to-regexp/releases/tag/v8.0.0): Simpler API

[Compare Source](https://redirect.github.com/pillarjs/path-to-regexp/compare/v7.2.0...v8.0.0)

Heads up! This is a fairly large change (again) and I need to apologize in advance. If I foresaw what this version would have ended up being I would not have released version 7. A longer blog post and explanation will be incoming this week, but the pivot has been due to work on Express.js v5 and this will the finalized syntax used in Express moving forward.

**Added**

-   Adds key names to wildcards using `*name` syntax, aligns with `:` behavior but using an asterisk instead

**Changed**

-   Removes group suffixes of `?`, `+`, and `*` - only optional exists moving forward (use wildcards for `+`, `{*foo}` for `*`)
-   Parameter names follow JS identifier rules and allow unicode characters

**Added**

-   Parameter names can now be quoted, e.g. `:"foo-bar"`

**Removed**

-   Removes `loose` mode
-   Removes regular expression overrides of parameters

### [`v7.2.0`](https://redirect.github.com/pillarjs/path-to-regexp/releases/tag/v7.2.0): Support array inputs (again)

[Compare Source](https://redirect.github.com/pillarjs/path-to-regexp/compare/v7.1.0...v7.2.0)

**Added**

-   Support array inputs for `match` and `pathToRegexp`  [`3fdd88f`](https://redirect.github.com/pillarjs/path-to-regexp/commit/3fdd88f)

### [`v7.1.0`](https://redirect.github.com/pillarjs/path-to-regexp/releases/tag/v7.1.0): Strict mode

[Compare Source](https://redirect.github.com/pillarjs/path-to-regexp/compare/v7.0.0...v7.1.0)

**Added**

-   Adds a `strict` option to detect potential ReDOS issues

**Fixed**

-   Fixes separator to default to `suffix + prefix` when not specified
-   Allows separator to be undefined in `TokenData`
    -   This is only relevant if you are building `TokenData` manually, previously `parse` filled it in automatically

**Comments**

-   I highly recommend enabling `strict: true` and I'm *probably* releasing a V8 with it enabled by default ASAP as a necessary security mitigation

### [`v7.0.0`](https://redirect.github.com/pillarjs/path-to-regexp/releases/tag/v7.0.0): Wildcard, unicode, and modifier changes

[Compare Source](https://redirect.github.com/pillarjs/path-to-regexp/compare/v6.2.2...v7.0.0)

Hi all! There's a few major breaking changes in this release so read carefully.

**Breaking changes:**

-   The function returned by `compile` only accepts strings as values (i.e. no numbers, use `String(value)` before compiling a path)
    -   For repeated values, when `encode !== false`, it must be an array of strings
-   Parameter names can contain all unicode identifier characters (defined as regex `\p{XID_Continue}`).
-   Modifiers (`?`, `*`, `+`) must be used after a param explicitly wrapped in `{}`
    -   No more implied prefix of `/` or `.`
-   No support for arrays or regexes as inputs
-   The wildcard (standalone `*`) has been added back and matches Express.js expected behavior
-   Removed `endsWith` option
-   Renamed `strict: true` to `trailing: false`
-   Reserved `;`, `,`, `!`, and `@` for future use-cases
-   Removed `tokensToRegexp`, `tokensToFunction` and `regexpToFunction` in favor of simplifying exports
-   Enable a "loose" mode by default, so `/` can be repeated multiple times in a matched path (i.e. `/foo` works like `//foo`, etc)
-   `encode` and `decode` no longer receive the token as the second parameter
-   Removed the ESM + CommonJS dual package in favor of only one CommonJS supported export
-   Minimum JS support for ES2020 (previous ES2015)
-   Encode defaults to `encodeURIComponent` and decode defaults to `decodeURIComponent`

**Added:**

-   Adds `encodePath` to fix an issue around `encode` being used for both path and parameters (the path and parameter should be encoded slightly differently)
-   Adds `loose` as an option to support arbitrarily matching the delimiter in paths, e.g. `foo/bar` and `foo///bar` should work the same
-   Allow `encode` and `decode` to be set to `false` which skips all processing of the parameters input/output
-   All remaining methods support `TokenData` (exported, returned by `parse`) as input
    -   This should be useful if you are programmatically building paths to match or want to avoid parsing multiple times

**Requests for feedback:**

-   Requiring `{}` is an obvious drawback but I'm seeking feedback on whether it helps make path behavior clearer
    -   Related: Removing `/` and `.` as implicit prefixes
-   Removing array and regex support is to reduce the overall package size for things many users don't need
-   Unicode IDs are added to align more closely with browser URLPattern behavior, which uses JS identifiers

</details>


ntucker:
Closing in favor of https://github.com/reactive/data-client/pull/3239 (moved these changes there)

PR #3480, opened 2025-04-29
fix: export types needed for EntityMixin

Tomaszal:
Fixes the following TypeScript error when using `EntityMixin`:

`The inferred type of ... cannot be named without a reference to '../../node_modules/@data-client/endpoint/lib/schemas/EntityTypes'. This is likely not portable. A type annotation is necessary.ts(2742)`

### DIFF
--- a/.changeset/shiny-olives-behave.md
+++ b/.changeset/shiny-olives-behave.md
@@ -1,0 +1,5 @@
+---
+'@data-client/endpoint': patch
+---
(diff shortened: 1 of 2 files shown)


ntucker (approved):

ntucker:
Thanks @Tomaszal for this PR; just released a new version of @data-client/endpoint with the fix.

PR #4034, opened 2026-07-19
demo: Add cache GC measurement harnesses (Node, browser, Android)

ntucker:
## Summary
- Add opt-in cache GC baselines for Node (`examples/benchmark` `start:gc`), Chromium (`examples/benchmark-react` `bench:gc`), and release Hermes Android (`examples/benchmark-native`)
- Wire BuildManifest / provenance verification so measurements bind to verified build identity rather than live HEAD
- Document scenario axes, timing boundaries, and validation gates in example READMEs and the GC plan

## Test plan
- [ ] `yarn install` succeeds with the new `examples/benchmark-native` workspace
- [ ] Node: `yarn workspace example-benchmark build && yarn workspace example-benchmark start:gc:verify` then a short `start:gc` filter run
- [ ] Browser: `yarn workspace example-benchmark-react build && yarn workspace example-benchmark-react test:gc-metrics && yarn workspace example-benchmark-react test:gc-provenance`
- [ ] Browser: `yarn workspace example-benchmark-react bench:gc --scenario unique-1000 --samples 1`
- [ ] Native: `yarn workspace example-benchmark-native typecheck && yarn workspace example-benchmark-native test && yarn workspace example-benchmark-native lint`
- [ ] Confirm CircleCI setup still strips example workspaces (including the new index)

Made with [Cursor](https://cursor.com)


ntucker:
## Staff review — CHANGE_THIS_PR

Demo/measurement harnesses look directionally right (timerless `BenchmarkGCPolicy`, prepare vs timed boundaries, `gc`/`no-gc` controls, BuildManifest provenance). One merge blocker before Android baselines are trustworthy:

### Release collect cannot retrieve reports on typical physical devices

Authoritative path is release-Hermes on a physical device (README + plan). `BenchNativeModule.writeReport` writes `gc-report.json` under private `filesDir`, and `scripts/collect-report.sh` only pulls via `run-as … cat`. The `release` buildType is not `debuggable` (AGP default; no override), so `run-as` fails on normal user builds. Wait can even pass via logcat `REPORT_READY`, then fail on pull. Emulator/userdebug exceptions don’t count for the mid-range physical gate.

Please fix collect before relying on this harness: write the post-run report to an adb-pullable path (e.g. `externalFilesDir`) and update `collect-report.sh` + README. Prefer that over `release { debuggable true }` so we don’t soften release-Hermes measurement semantics; `debuggable true` is only a last resort with an explicit caveat (this APK is already debug-signed).

### Follow-ups (not blocking)

- Align `examples/benchmark-native` `react-native` / `@react-native/*` from `0.86.0` → `0.86.2` to match root/`packages/react` and drop dual lockfile trees (declared scaffold vs root gradle already tends toward 0.86.2).
- No shared harness package yet — plan already says v1 is vocabulary-only; extract after real cross-harness drift.
- CircleCI `npm pkg delete workspaces[N]` correctly adds `[7]` for this insert, but remains index-fragile; later delete/set by name.
- Soften plan wording that implies Android already measures navigation/gestures; v1 is a synthetic interaction/frame probe.
- BuildManifest complexity is fine for baseline integrity; consolidate later if the three copies drift.

Bugbot: the older `node -e` / `argv` finding is a false positive (already verified in-thread). The tip finding on release `run-as` is real and matches the change request above.

Adversarial pass: SURVIVE on collect (prefer externalFilesDir); RN pin downgraded to FOLLOW_UP.

ntucker:
## Staff review — LGTM (prior CHANGE_THIS_PR cleared)

Tip `66dcaa7` fixes the release-collect blocker: report goes to `externalFilesDir`, Q+ mirrors via MediaStore Downloads, `collect-report.sh` `adb pull`s (no `run-as`), README documents why not `release { debuggable true }`, and the script contract test locks it in. Preferable path over softening release debuggability — good.

No new CHANGE_THIS_PR. Prior follow-ups still stand after merge (RN `0.86.0`→`0.86.2` align, shared harness later, CircleCI workspace index-by-name, plan wording on gestures, BuildManifest consolidate-if-drift). Optional later: if MediaStore insert fails on Q+, still emit `REPORT_READY` for the externalFilesDir write so wait/pull degrade gracefully — not a merge blocker.

Hold squash until Ready + Bugbot green on this tip + any required E2E you care about for the harness. I will not merge from here.

PR #4090, opened 2026-09-11
feat(react): incrementally hydrate SSR state from streamed deltas

ntucker:
### Motivation

A streaming HTML response is a sequence of committed revisions. A store serialized **once** — after a quiet window, or after `useReadyCacheState()` — cannot describe that sequence. Late islands arrive as markup whose endpoint/meta never landed in the browser, so [`useSuspense()`](https://dataclient.io/docs/api/useSuspense) takes the fetch branch on an empty seed.

PR 4090 already replaced the Next 10 ms `createPersistedStoreServer` snapshot **on the wire**: an inert baseline plus per-flush `StateDelta`s. This tip also moves streamed hydration **out of the public SPA store**: public `DataProvider` / `@data-client/react` use the master reducer and `StateContext` reads. `./nextjs` and an unstable `./ssr` `__INTERNAL__` seam compose `createStreamingReducer` and the adapter-owned snapshot.

That is not streaming success. Head still folds into the live store from `StreamedStateReceiver`’s layout effect (`flushSync` plus a live write), has no per-key waiter, and has no render-pure revision-visibility protocol. Generic `renderToPipeableStream` / Anansi still emit one document snapshot by default.

**HOLD. Do not merge without Nathaniel.** Green tests mean the SPA path was restored and Next behavior was relocated, not that duplicate fetches, concurrent streaming, or live merge integrity are solved.

### Solution

**Adapter-owned streaming (this tip):**

- Core master `createReducer` has no `HYDRATE` case. `HYDRATE` / `hydrateReducer` / `createHydrate` live under `state/stream` and root `__INTERNAL__` only. `HydrateAction` is not a public `ActionTypes` member.
- Public `DataProvider` is a thin wrapper over private `DataProviderBase` with core `createReducer`. `DataStore` uses that same reducer for committed updates and optimistic replay.
- Browser `useCacheState` is `use(StateContext)` again. The `SSRDataProvider` server external-store branch remains. `ServerSnapshotContext` / `createServerSnapshot` are deleted.
- Shared `packages/react/src/server/stream/` owns `createStreamingReducer`, minimal `SnapshotStore` / `DeltaQueue`, and `StreamedStateReceiver`. Next keeps transport (global queue, baseline element, scripts, nonce, managers/pending extensions).
- Next `createPersistedStoreClient` composes `DataProviderBase + createStreamingReducer + StreamedStateReceiver`. `flushSync` and per-delta live publication are **preserved HOLD behavior**.
[...]

PR #4094, opened 2026-09-18
test(react): Lock concurrent rendering GOALS on public APIs

ntucker:
Fixes # .

### Motivation

Lock **Fully concurrent rendering on the latest React** for public `@data-client/react` (`DataProvider`, `useSuspense`, `useCache`, `useLive`, `useController`). No `packages/*` source. No changeset. No #4090 adapter APIs.

`GOALS.md` on this PR carries that bullet (and the 60 FPS “not a tradeoff” clause) so the written goal and the tests can land together. Master still lacks the bullet. HOLD #4090 is not required.

This is not a universal “all live Context publishes preserve pending siblings” lock. Shape 1 is the green gate-before-consumer path. Shapes 2 and 3 record topology; they do not declare the recorded outcome good.

### What is locked (green)

A green test fails if any of these regress on latest React:

- `useTransition` over a cache-missing `useSuspense`: old UI stays, no fallback, `isPending`, urgent input still commits, then B.
- Isolated store publishes do not commit at Immediate/SyncLane.
- Public cache reads do not call `useSyncExternalStore`. React 19 uses `use(StateContext)`; React 18’s `useContext` fallback is labeled.
- No call through the exported `react-dom.flushSync` in the public-provider transition/hydration scenarios.
- Shape 1 streaming hydrate: ready A keeps identity and interactivity, zero fetches, empty recoverable errors.
- Shape 1 publish does **not** client-render the gated pending sibling (B stays `<!--$?-->`). React 19: both transition and default-lane. React 18: transition path must keep B pending; **default-lane hydration client-render of B is the explicit worse fallback**.
- Streamed reveal adopts B by node identity (not a client-rendered replacement).

### Recorded only (not product success)

- Shape 2 (read-then-suspend) and shape 3 (two pending islands): per-minor table of pending / render / adoption / fetch. Drift or an unrecorded React minor fails. A row could record client-render and still pass.
- Probe controls: uSES notify is Immediate even inside `startTransition`; discrete `setResponse` is Immediate; recreating Suspense client-renders pending B. Those paths are expected; they do not turn equivalent Data Client behavior red.

On public master, shapes 2 and 3 currently record: both stay pending after publish, B not rendered, streamed node adopted, `fetchIds: []` (18.3.1 / 19.2.3 / 19.3.0). #4090’s adapter live `HYDRATE` client-rendered B; that difference is reported, not rewritten.

### Not caught

- #4090 `StreamedStateReceiver` / adapter `flushSync`. That needs #4090 code.
- An app that simply omits `useTransition`. The suite proves the library works *inside a caller-owned transition*.
- A sync mechanism that bypasses the `react-dom.flushSync` export, unless it also makes an isolated publish Immediate.

### Solution

Tests only, plus `GOALS.md`. CircleCI adds a verified `unit_tests-19.3` row (`yarn up react@19.3.0`); the `latest` row stays on the 19.2.3 pin. Suites skip on React 16/17.

Harness: DevTools commit probe (hook in ReactDOM `scripts/testSetup.js`; fail-loud calibration in probe-controls), two-phase `renderToPipeableStream` / `hydrateRoot` replay, pending `template[id]` discovered from the installed shell.

### Open questions

Retire `unit_tests-19.3` when `latest` pins ≥ 19.3.

<div><a href="https://cursor.com/agents/bc-02a1c82c-1fee-578d-822b-8e5c85d5bbf3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-02a1c82c-1fee-578d-822b-8e5c85d5bbf3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ntucker:
**FOLLOW_UP** (Staff Architect) — does not block merge after Ready / `/simplify`.

Solid GOALS lock: public `@data-client/react` only, no `packages/*` source, honest conditional cases (shape 1 green vs shapes 2–3 recorded). Harnesses + self-calibrated DevTools probe are the right architecture; 19.3 CI row is correctly focused.

After land, please consider:

1. **Drop duplicate calibration** — `concurrent-hydration.web.tsx` and `concurrent-transition.web.tsx` still copy the uSES-Immediate / plain-setState-Normal blocks that `concurrent-probe-controls.web.tsx` already owns post-`0acb4b2`. Keep one suite as the probe control; let the others rely on it (or a shared `beforeAll` helper).

2. **Soften `B:0` / `B:1` coupling** — Shape 1’s green path pins React streaming Suspense template ids. If a future React minor renumbers boundaries, prefer asserting dehydrated pending via `<!--$?-->` + “B not in `renders` / no fetch” where that still proves the contract.

3. **Retire `unit_tests-19.3` when `latest` pins ≥19.3** — Avoid a permanent extra matrix job once the repo pin catches up.

4. **DevTools priority maintenance** — Treat Immediate/Normal remaps as probe recalibration, not a data-client product break (already self-calibrated; call that out in the spike/docs follow-ups).

No CHANGE_THIS_PR. Draft is fine to review; merge gate remains Ready after `/simplify` + green checks.

ntucker:
**FOLLOW_UP** (no change required on this PR)

Direction looks right for a public-API concurrent-rendering goal lock: shape 1 as the green regression guard, shapes 2/3 as per-minor recorded controls (not #4090 adapter contracts), shared probe + streaming harness after simplify, and a focused `unit_tests-19.3` row that deletes resolutions rather than fighting the 19.2.3 pin.

When you next touch this area (not blocking Ready/merge):

1. **Soften `B:0` / `B:1`** — prefer discovering pending `template[id]` from the installed shell (or fallback-text + `<!--$?-->`) so a React HTML-protocol id tweak doesn’t brick the goal suite.
2. **DevTools hook** — consider installing `__REACT_DEVTOOLS_GLOBAL_HOOK__` from ReactDOM `scripts/testSetup.js` so suites can drop the `import/order` ritual and stay safe if a `jest.mock('react-dom')` ever `requireActual`s client early. Keep per-suite calibration fail-loud either way.
3. **Fixture drift** — shared Todo/Endpoint/`mockInitialState` helper; keep Immediate/Normal calibration primarily in `concurrent-probe-controls` unless a file’s mocks need a local smoke.
4. **19.3 → pin** — when 19.3 becomes `latest`, widen the CI row past the concurrent path glob and add shape 2/3 table rows for each new React minor.

